PR #10, opened 2015-04-05
update charter with newer template

wayneca:
changes have been made to the charter template from feedback from bluetooth cg and AC discussions.  For instance, it makes it clearer that CGs usually do things by whatever informal process they want.  The charter provides fallbacks for when there is disagreement about whether the result was fair.

### DIFF
--- a/charter/index.html
+++ b/charter/index.html
@@ -291,92 +291,150 @@
   </dl>
 </section>
 
-<section><h1>Community and Business Group Process</h1>
-  <p>
-    Terms of in this charter that conflict with those of the Community and
-    Business Group Process are void.
-  </p>
-  <section><h2>Work Limited to Charter Scope</h2>
-    <p>
-      The group will not publish Community Group Reports that are Specifications
-      on topics other than those listed under "Community Group Reports that are
-      Specifications" above. See below for how to modify the charter. The CLA
-      applies to these Community Group Reports.
-    </p>
+<section>
+    <h2>Community and Business Group Process</h2>
+    <p>Terms of in this charter that conflict with those of the Community and
+    Business Group Process are void.</p>
   </section>
-  <section><h2>Contribution Mechanics</h2>
-    <p>
-      For these Reports, Community Group participants agree to send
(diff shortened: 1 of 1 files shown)


charter/index.html
```diff
@@ -291,92 +291,150 @@ <h1 id="title">
...
-  </p>
-  <section><h2>Work Limited to Charter Scope</h2>
-    <p>
-      The group will not publish Community Group Reports that are Specifications
-      on topics other than those listed under "Community Group Reports that are
-      Specifications" above. See below for how to modify the charter. The CLA
-      applies to these Community Group Reports.
-    </p>
+<section>
+    <h2>Community and Business Group Process</h2>
```
jyasskin, line 295: FWIW, reformatting the HTML at the same time as you're changing the content, makes it really hard to see what's changed. Sometimes you can't avoid that, when you're reordering sections, but even then, it can help to reorder in one commit, and then reword in a separate one.

jyasskin:
Should this have updated the Last Modified date?

charter/index.html
```diff
@@ -291,92 +291,150 @@ <h1 id="title">
...
-      The group will conduct all of its technical work on its public mailing
-      list. Any decisions reached at any meeting are tentative. Any group
-      participant may object to a decision reached at an online meeting within
-      7 days of publication of the decision on the mail list. That decision must
-      then be confirmed on the mail list by the Decision Process above.
-    </p>
+  
+  <section>
+    <h2>Transparency</h2>
+    <p>The group will conduct all of its technical work on its The group will
```
jyasskin, line 381: 2 sentences collided here.

charter/index.html
wayneca, line 295: I agree.  I took this from the charter template after making the changes in the template, so it was already reformatted there.

charter/index.html
wayneca, line 381: I did a PR to fix it.  The original sentence was from the template where all technical work on the spec was on the public mail list, not GitHub.

PR #11, opened 2015-04-06
charter - change date and fix errorl

wayneca:
updated the last modified date and corrected the first sentence in transparency.  the problem with the older version of that sentence is is said all technical work would take place on the public mail list -- that was from a version of the template where GitHub wasn't being used.


jyasskin:
There's also an "acceeptable" farther down. These look like improvements to me.

PR #13, opened 2015-04-08
Added alternative implementation details to Web NFC

zolkis:
There is much rearrangement of the text. In essence, added an encoded JSON format for Web NFC data and metadata, both for abstracting away NFC Forum types, and for supporting implementations where an NDEF message contains only one NDEF record.

### DIFF
--- a/security-privacy.html
+++ b/security-privacy.html
@@ -224,13 +224,13 @@
     </p>
     <dl title="dictionary NfcAccess" class="idl">
       <dt>DOMString scope</dt>
-      <dt>boolean write</dt>
+      <dt>boolean canWrite</dt>
     </dl>
     <p>
       The <dfn id="widl-NfcAccess-scope">scope</dfn> property is a string
       pattern for matching the URL's that are explicitly allowed to access
       the data, for instance <br>
-      <code>{ "scope": "https://mydomain.com", "write": "true"}</code>. <br>
+      <code>{ "scope": "https://mydomain.com", "canWrite": "true"}</code>. <br>
       Scopes which don't
       <a href="http://w3c.github.io/web-nfc/#the-within-scope-algorithm">match</a> this are not permitted to access the data.
       If the whitelist is empty, then the platform default policy applies,
@@ -239,7 +239,7 @@
       <a href="http://w3c.github.io/web-nfc/#the-within-scope-algorithm">match</a> the origin of the page which wrote the existing data.
     </p>
     <p>
-      When the <dfn id="widl-NfcAccess-write">write</dfn> property is
+      When the <dfn id="widl-NfcAccess-canWrite">canWrite</dfn> property is
       <code>true</code>, then the URL's matching <code><var>scope</var></code>
       are permitted to write (in fact, update) the data, otherwise they are only permitted to read it.
     </p>
@@ -342,8 +342,8 @@
                       "id": "http://w3.org/web-nfc/access/0/scope",
                       "type": "string"
                     },
-                    "write": {
-                      "id": "http://w3.org/web-nfc/access/0/write",
+                    "canWrite": {
+                      "id": "http://w3.org/web-nfc/access/0/canWrite",
                       "type": "boolean"
                     }
                   }
--- a/security-privacy.html
+++ b/security-privacy.html
@@ -313,6 +313,11 @@
         </li>
       </ul>
     </p>
+    <p class="note">
+      The encoded JSON format for Web NFC data and metadata is useful for
+      abstracting away NFC Forum types, and for supporting implementations where
+      an NDEF message contains only one NDEF record.
+    </p>
   </section>
 </section>
 
--- a/security-privacy.html
+++ b/security-privacy.html
@@ -194,11 +194,11 @@
     In order to support security policies concerning the Web NFC API, certain
     mechanisms need to be supported in all the implementations, such as:
     <ul>
-      <li>a mechanism for telling apart “Web NFC tags” from generic NFC tags
-        </li>
-      <li>a mechanism for white-listing URL's that can read, or write a certain
+      <li>an identification mechanism for telling apart “Web NFC tags” from
+        generic NFC tags</li>
+      <li>a mechanism for whitelisting URL's that can read, or write a certain
         Web NFC tag</li>
-      <li>a mechanism for black-listing URI's that can read, or write a certain
+      <li>a mechanism for blacklisting URI's that can read, or write a certain
         Web NFC tag.</li>
     </ul>
   </p>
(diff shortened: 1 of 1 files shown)


security-privacy.html
```diff
@@ -194,11 +194,11 @@
...
     <ul>
-      <li>a mechanism for telling apart “Web NFC tags” from generic NFC tags
-        </li>
-      <li>a mechanism for white-listing URL's that can read, or write a certain
+      <li>an identification mechanism for telling apart “Web NFC tags” from
+        generic NFC tags</li>
+      <li>a mechanism for whitelisting URL's that can read, or write a certain
         Web NFC tag</li>
-      <li>a mechanism for black-listing URI's that can read, or write a certain
+      <li>a mechanism for blacklisting URI's that can read, or write a certain
```
kenchris: This sounds weird... read (past) write (present), "that can be read or written using..." ?

security-privacy.html
```diff
@@ -207,90 +207,117 @@
 <section class="informative"> <h2>Implementation details</h2>
   <p>
     This section contains discussion of implementation possibilities of security
-    related mechanisms.
+    related mechanisms. The following implementation alternatives are possible:
+    <ul>
+      <li>
+        Use additional NDEF records in an NDEF message containing
```
kenchris: , before containing

security-privacy.html
```diff
@@ -207,90 +207,117 @@
...
+        Web NFC specific metadata, such as identification, whitelists and
```
kenchris: metadata specific to Web NFC, such as

security-privacy.html
```diff
@@ -207,90 +207,117 @@
...
+        blacklists.
+      </li>
+        Use a specific Web NFC record format and encode the Web NFC specific
```
kenchris: , before and - as that is clearer

security-privacy.html
```diff
@@ -207,90 +207,117 @@
...
+        metadata into that.
+      </li>
+    </ul>
   </p>
-  <section> <h3>Web NFC tags</h3>
+  <section> <h3>Web NFC using additional NDEF records for metadata</h3>
```
kenchris: this isnt really meta data in the strictest sense.

security-privacy.html
```diff
@@ -207,90 +207,117 @@
...
+    <p>
+      A typical Web NFC NDEF message will contain the following
+      NDEF records:
+      <ul>
+        <li>Normal NDEF records carrying data.</li>
+        <li>
+          An additional NDEF record for identification.
```
kenchris: I would say "An additional NDEF Web NFC identification record" to make it clear what we are identifying

security-privacy.html
```diff
@@ -207,90 +207,117 @@
...
+          The preferred type of this NDEF record is NFC Forum External Type
+          (TNF=0x04), with the External Type field set to e.g.
```
kenchris: why e.g. ... it sounds like it is optional what to use

security-privacy.html
```diff
@@ -207,90 +207,117 @@
...
+          <code>urn:nfc:ext:w3.org:web-nfc-id</code>.
+        </li>
+        <li>
+          An optional additional NDEF record for whitelisting for
```
kenchris: Optionally, a NDEF record representing a read whitelist

security-privacy.html
```diff
@@ -207,90 +207,117 @@
...
+        </li>
+      </ul>
+    </p>
+  </section>
+  <section> <h3>Web NFC using encoded metadata</h3>
     <p>
-      For making tags that can be read and written by the Web NFC API, and for
-      making sure that web pages cannot access generic NFC tags, the following
-      are possible:
+      Web NFC may also define an own format for accepted NDEF records,
```
kenchris: an own format? I dont ge tit

security-privacy.html
```diff
@@ -207,90 +207,117 @@
...
+      where the preferred type of this NDEF record is NFC Forum External Type
+      (TNF=0x04), with the External Type field set to
+      <code>urn:nfc:ext:w3.org:webnfc</code>. The payload type would be
```
kenchris: would? SHOULD ?

security-privacy.html
```diff
@@ -207,90 +207,117 @@
...
+      <code>application/webnfc+json</code> (by default UTF-8 encoded). All data
```
kenchris: encoded as UTF-8 by default)

security-privacy.html
```diff
@@ -207,90 +207,117 @@
...
+      would be JSON, for instance with the following schema:
```
kenchris: would should must?

security-privacy.html
```diff
@@ -207,90 +207,117 @@
...
+          <code>id</code> field.
+        </li>
+        <li>
+          <code>messageType</code>: <code>"string"</code>; contains the type of
+          the data, either <code>"text"</code>, <code>"url"</code>,
+          <code>"json"</code>, or <code>"blob"</code>.
+        </li>
+        <li>
+          <code>message</code>: <code>"string"</code>; contains the payload data
+          encoded as text (text, URL and JSON), and for blob encoded as Base64.
```
kenchris: Base64 needs a link I think...

security-privacy.html
```diff
@@ -207,90 +207,117 @@
...
+        </li>
+        <li>
+          <code>whitelistRead</code>: <code>"array"</code>; contains the URL
```
kenchris: readWhitelist makes more sense to me

kenchris: but I am not sure how commonly understood whitelist/blacklists are... maybe we can find a better name.

security-privacy.html
```diff
@@ -207,90 +207,117 @@
...
+          scopes that are explicitly allowed to read the data. No other scope
+          is permitted to read the data.
+        </li>
+        <li>
+          <code>whitelistWrite</code>: <code>"array"</code>; contains the URL
+          scopes that are explicitly allowed to rewrite the data. No other scope
+          is permitted to rewrite (update) the data.
+        </li>
+        <li>
+          <code>blacklistRead</code>: <code>"array"</code>; contains the URL
```
kenchris: I find a blacklist useless... it is so easy to work around... put an iframe to a different url, read and send info back. Let's not fool people into believe it is safe

kenchris: Maybe instaed we could have an "access" field and people could supply some array like

{ "http://pol.dk", "rw" } 

We could encode it as rw|http://pol.dk etc on the tag

security-privacy.html
```diff
@@ -207,90 +211,161 @@
...
+        </li>
+        <li>
+          <code>messageType</code>: <code>"string"</code>; contains the type of
+          the data, either <code>"text"</code>, <code>"url"</code>,
+          <code>"json"</code>, or <code>"blob"</code>.
+        </li>
+        <li>
+          <code>message</code>: <code>"string"</code>; contains the payload data
+          encoded as text (text, URL and JSON), and for blob encoded as
+          [[Base64]].
```
jyasskin: Usually we'd just use `[[rfc4648]]` since it's already in the database.

security-privacy.html
```diff
@@ -207,90 +211,161 @@
...
+          The preferred type of this NDEF record is NFC Forum External Type
+          (TNF=0x04), with the External Type field set to
+          <code>urn:nfc:ext:w3.org:webnfc-id</code>.
+        </li>
+        <li>
+          Optionally, a NDEF record representing a read whitelist.
+          The preferred type of this NDEF record is NFC Forum External Type
+          (TNF=0x04), with the External Type field set to
+          <code>urn:nfc:ext:w3.org:webnfc-rwl</code>.
+          The payload is a string list of URL's, delimited by white space.
```
jyasskin: Should these be URLs or [ascii-serialized Origins](https://html.spec.whatwg.org/multipage/browsers.html#ascii-serialisation-of-an-origin)?

security-privacy.html
```diff
@@ -207,90 +211,161 @@
...
+          <code>message</code>: <code>"string"</code>; contains the payload data
+          encoded as text (text, URL and JSON), and for blob encoded as
+          [[Base64]].
+        </li>
+        <li>
+          <code>access</code>: <code>"array"</code>; contains tuples
+          for the desired access type (read-write, or read-only),
+          and the URL scopes that are explicitly allowed to access the data,
+          for instance
+          <code>{ "scope": "https://mydomain.com", "write": "true"}</code>.
```
jyasskin: Do you want to allow `*.mydomain.com`? And plain `*`? We'll need to find somewhere to define these origin filters; it doesn't make much sense for that to live in NFC long-term. Maybe ping webappsec.

security-privacy.html
```diff
@@ -207,90 +211,161 @@
...
+      <!--
+      The full JSON schema is the following:
+      <pre title = "Web NFC JSON schema" class="example">
+        {
+          "$schema": "http://w3.org/web-nfc#",
+          "id": "http://w3.org/web-nfc",
+          "type": "object",
+          "properties": {
+            "id": {
+              "id": "http://w3.org/web-nfc/id",
```
jyasskin, line 318: The W3C usually puts dates in their URIs, so `http://w3.org/2015/web-nfc/*`. I don't know if that makes sense here though.

security-privacy.html
```diff
@@ -207,90 +211,161 @@
...
+            "messageType",
+            "message",
+          ]
+        }
+      </pre>
+    </p>
+    -->
+    <p class="note">
+      The encoded JSON format for Web NFC data and metadata is useful for
+      abstracting away NFC Forum types, and for supporting implementations where
```
jyasskin, line 365: I don't really understand this paragraph. Do you mean that you've chosen the JSON format because it allows you to avoid exposing raw NFC Forum types, and because it can store multiple values?

security-privacy.html
```diff
@@ -207,90 +211,161 @@
...
+        <li>
+          <code>access</code>: <code>"array"</code>; contains tuples
+          for the desired access type (read-write, or read-only),
+          and the URL scopes that are explicitly allowed to access the data,
+          for instance
+          <code>{ "scope": "https://mydomain.com", "write": "true"}</code>.
+          Scopes not matching this are not permitted to access the data.
+          If the list is empty, then the platform default policy applies, which
+          SHOULD permit reading for everyone, but SHOULD prevent writing, except
+          to sub-scopes of the URL scope of the page which has created the data.
```
jyasskin: What's the "page which has created the data"?

security-privacy.html
```diff
@@ -207,90 +211,161 @@
...
+    <p>
+      Web NFC may also define an own NDEF extended record format,
+      where the preferred type of this NDEF record is NFC Forum External Type
+      (TNF=0x04), with the External Type field set to
+      <code>urn:nfc:ext:w3.org:webnfc</code>. The payload type is
+      <code>application/webnfc+json</code> (encoded as UTF-8 by default). All
+      data is JSON, with the following schema:
+      <ul>
+        <li>
+          <code>id</code>: <code>"string"</code>; contains the Web NFC
```
jyasskin, line 284: What's the need for this?

security-privacy.html
```diff
@@ -207,90 +211,161 @@
...
+          identifier, e.g. the SHA256 signature of <code>"w3.org/web-nfc”</code>
+        </li>
+        <li>
+          <code>scope</code>: <code>"string"</code>; contains the origin of the
```
jyasskin: I'm a little nervous about letting things claim an origin without any evidence the owners of that origin actually own the data. I'm not totally sure what to do instead though. Can you add this as an open question somewhere here, and then ping webappsec?

security-privacy.html
zolkis: Yes, corrected.

security-privacy.html
```diff
@@ -207,90 +211,161 @@
...
-        (TNF=0x04), with the External Type field set to e.g.
-        <code>urn:nfc:ext:w3.org:web-nfc-bl-write</code>.
-        The payload is a string list of URL's, delimited by white space.
-      </p>
-    </section>
+  </section>
+  <section> <h3>Web NFC using encoded metadata</h3>
+    <p>
+      Web NFC may also define an own NDEF extended record format,
+      where the preferred type of this NDEF record is NFC Forum External Type
```
jyasskin: Can you cite the spec that defines these terms?

security-privacy.html
zolkis, line 284: There is a slight chance there are NFC tags out there with a given content that coincides with that of a Web NFC tag. This minimizes that risk. Otherwise IMO it would be enough to have the web-nfc specific urn namespace in the external type. The answer is: not really needed, but I am waiting on feedback from @sicking on this.

security-privacy.html
zolkis: The use case is preventing random web pages from overwriting tags. The origin which has originally written (I understand it may not have created the data itself, but at least it created the tag) the data is recorded, so that implementations could enforce default policies like "only scopes matching the writer (creator) scope can update the data". If this is not good enough, create an issue for tracking it (I can also create it), since this is pretty important detail.

security-privacy.html
zolkis: Thanks, done.

security-privacy.html
zolkis: If we want to include pattern matching, yes. If it's only about scopes and sub-scopes, it's matter of defining the algorithm for matching scopes. Worth creating an issue for tracking.

security-privacy.html
zolkis: The page that has written the data; the origin has been recorded by the implementation under the writing page.
Maybe the wording 'created' is misleading indeed - I will use 'written' instead.

security-privacy.html
zolkis, line 318: This whole schema is commented out, I just included for the editors' reference.

security-privacy.html
kenchris: We really need one definition of scopes as it is also used by Service Workers @PaulKinlan @jakearchibald

security-privacy.html
```diff
@@ -194,12 +200,10 @@
     In order to support security policies concerning the Web NFC API, certain
     mechanisms need to be supported in all the implementations, such as:
     <ul>
-      <li>a mechanism for telling apart “Web NFC tags” from generic NFC tags
-        </li>
-      <li>a mechanism for white-listing URL's that can read, or write a certain
+      <li>an identification mechanism for telling apart “Web NFC tags” from
+        generic NFC tags</li>
+      <li>a mechanism for whitelisting URL's that can read, or write a certain
```
kenchris, line 199: white-listing?

security-privacy.html
zolkis, line 365: Yes for the first half. No for the second: when the platform supports only one record per message, we need to encode all data there, since we cannot use the first alternative. But then it's not any more a generic tag, i.e. we cannot use records of type "text", "URL", etc, as defined by NFC Forum, instead we need to encode all data into a Web NFC specific record, used as a container.

security-privacy.html
zolkis: In the Terminology section it states that "For NFC terminology, please consult the Web NFC specification."
Now I could copy that from the Web NFC spec, making it duplicate and awkward to maintain, or move this doc as a section into the Web NFC spec.

security-privacy.html
zolkis, line 199: It's fine as 'whitelisting'. http://en.wikipedia.org/wiki/Whitelist

security-privacy.html
zolkis: Suggestion?

kenchris: Additional NDEF records for Web NFC

We both use it for data, identify and configuration

security-privacy.html
zolkis: I included an "issue" note about this, and will also create a github issue.

security-privacy.html
zolkis: I included an "issue" note about this and will also create a github issue once the PR is in.

PR #18, opened 2015-05-04
Reworked the specification, security doc and use cases doc, 

zolkis:
reflecting security reviews and implementation feedback.


anssiko:
Thanks, LGTM. We should publish this as a CG Draft Report to get wider feedback and review.

kenchris:
lgtm

PR #41, opened 2015-08-24
Web NFC snapshot release 24 August 2015

anssiko:
Implementors need to be aware that this version of the specification is
considered unstable. This snapshot is taken to make it easier to refer to
this specific version in future discussions.


anssiko:
@kenchris @zolkis Please take a look.

zolkis:
lgtm, merging

PR #43, opened 2015-09-02
draft.idl: Added compatibility mode flag NFCWatchMode to NFCWatchOptions

zolkis:


zolkis:
This is a change done to the upcoming version of the IDL, and the text is being updated.

Issue #40, opened 2015-08-18
Simplify process for obtaining permissions


alexshalamov:
At the moment spec defines that UA should obtain user permissions when:
- NFCAdapter.send() is called for 'tag' target
- NFCAdapter.send() is called for 'peer' target
- When event listener is added to NFCAdapter object (for listening to messages)

I couldn't find any other specs that require to obtain permission from user when event listener is added.

I would propose to have permission checks when:
- navigator.nfc.requestAdapter() is called
- NFCAdapter.send() is called (no matter if target was 'peer', 'tag' or 'any')

kenchris:
What does WebBluetooth do?

I think it is bad to check permission when calling requestAdapter because that will presumable be done in the initialization of the document, which in practise means up front permissions which we are trying to avoid.

I kind of fear the same will happen with the listeners, so maybe it is possible to prompt before executing them instead?

annevk:
Adding event listeners must not have side effects. This is definitely the wrong design.

kenchris:
@annevk right, I also don't like prompting when adding listeners, but would it be possible to prompt before executing them? Basically prompt when receiving data (ie not share the data before) and potentially only prompt if there are listeners present?

annevk:
You mean on dispatch? That would be okay. Only prompting if there are listeners present would be less okay... Surely there must be some better design for this?

kenchris:
Yes on dispatch. The problem with requestAdapter here is that, no one expects to actually use it immediately after unlike the requestDevice (of WebBluetooth fame), because you never know when you might receive data - so developers will presumable call this as part of initialization. And we really want to avoid upfront prompts especially because it is hard for the user to associate that with an action

annevk:
Since the processing model for the event is not defined it's hard to come up with alternative models that might meet such a need. E.g., if it's always a one-off, using a promise-based pulling API might be better.

kenchris:
It's not. The browsing context will basically never know when someone bumps a tag or peer against the device.

annevk:
The adapter could have a `requestNFCData()` method or some such. That way there's also less of a chance of missing a message.

kenchris:
I guess then it might as well be a configuration on the requestAdapter (request read)

alexshalamov:
@kenchris WebBluetooth requires UA to inform user when requestDevice is called. https://webbluetoothcg.github.io/web-bluetooth/#security-and-privacy

alexshalamov:
[...]

Issue #39, opened 2015-08-18
send() method algorithm talks about arrays


annevk:
Pretty sure that IDL does not pass an array but more of an abstract sequence concept.

zolkis:
Right. This is also fixed in the upcoming version of the spec.

Issue #38, opened 2015-08-18
How do you construct an NFCMessage to pass to send()?


annevk:
You cannot create your own basically?

zolkis:
Yes, it should be a dictionary. In the new version of the spec (under work) it already is.

Issue #36, opened 2015-08-18
website is not a thing


annevk:
Please make sure everything you say is established in first principles.

kenchris:
What is the right term? web page?

annevk:
I think often you mean to say browsing context (or address some part of that concept).

annevk:
Note that the same goes for "web app". E.g., "web app's document URL" is pure fiction. Where is it defined that a "web app" has a document URL? And what does that mean?

zolkis:
Yeah, these are more colloquial than standardized/canonical names, even though are defined terms. Would be good to use one terminology - suggestions?

annevk:
Well, you basically want to look at what these map to in an implementation and where the concepts are defined that this implementation already has. Typically that means studying the HTML Standard.

kenchris:
Seems pretty fitting:

http://www.w3.org/TR/html5/browsers.html#windows

A browsing context is an environment in which Document objects are presented to the user. Note: A tab or window in a Web browser typically contains a browsing context, as does an iframe or frames in a frameset.

jyasskin:
Here's a first cut at replacing uses of websites in the NFC spec: 
1. The uses in http://w3c.github.io/web-nfc/index.html#reading-nfc-tags are fine: they're non-normative and can use colloquial language.
2. For "a website must be visible and in focus", copy wording from http://www.w3.org/TR/page-visibility/, or just refer to its terms.
3. For 'only websites with a schema starting with "https://"', write it into the algorithms instead. https://webbluetoothcg.github.io/web-bluetooth/#requestDevice-secure-context has an example.

kenchris:
Thanks a lot for these pointers @jyasskin !

PR #44, opened 2015-09-05
New version of the Web NFC specification and the Security and Privacy document.

### LINKED ISSUE #4: Use wording from [powerful-features] to implement HTTPS restriction
_From @jyasskin on February 7, 2015 2:35_

https://w3c.github.io/webappsec/specs/powerfulfeatures/#new suggests putting wording in each algorithm that needs to be constrained to HTTPS.

_Copied from original issue: w3c/nfc#62_

### LINKED ISSUE #29: Say "in parallel" rather than "asynchronously"
The former is a defined term in HTML.

### LINKED ISSUE #34: Add "Sending data to NFC peers" example
The spec is clear on the [use cases](https://w3c.github.io/web-nfc/#use-cases) it aims to address. Currently, examples exists for reading and writing NFC tags, but an example for sending data to NFC peers is missing from [examples](https://w3c.github.io/web-nfc/#examples). It would be good to explain the key use cases in terms of example code. 

[Maintaining example code (as well as the use cases) as the API evolves allows people who don't read WebIDL (e.g. most web developers) to provide feedback. Also, example code helps find any developer ergonomics issues that may be masked by WebIDL.]

### LINKED ISSUE #45: Need ability to cancel a message push
When a pushMessage (formerly send) Promise is unresolved, do subsequent calls queue, replace, or are they rejected?

In any case, there needs to be a mechanism for cancelling a message push.

Also, what is the meaning of

> 10. If after making the request the connection is lost, reject promise with "TimeoutError"

in the [draft API](http://w3c.github.io/web-nfc/index.html#the-send-method)? Is this when data transfer is incomplete?

zolkis:
Algorithms extended and reworked.
Changed data representation. 
New security policies. 
Use watches instead of events. 
Fixed terminology related issues. 
Renamed send() to pushMessage().
Added timeout to push options. 
Issues addressed: #2, #3, #22, #26, #28, #30, #31, #32, #33, #35, #36, #38, #39, #40.
Also,
Fixed #4: use wording from [powerful features].
Fixed #29: use "in parallel" instead on "asynchronously".
Fixed #34: add example for sending data to NFC peer.
Fixed #45: add cancelPush() method.
Also, added steps on what happens with watches and push messages when focus is switched to another browsing context.


mrj:
Should NFCDeviceType be changed to NFCPushTarget?

zolkis:
We went back and forth between these two names. Perhaps NFCPushTarget is more correct.

zolkis:
Will make a new commit in this PR to fix it.

zolkis:
It seems that I have already done that and have already forgotten it :). Perhaps the comment referred to removed code (draft.idl).

mrj:
Yes, sorry, I thought draft.idl was the latest because it has the latest modification date.

anssiko:
See also the [HTML diff](http://services.w3.org/htmldiff?doc1=http%3A%2F%2Fw3c.github.io%2Fweb-nfc%2Freleases%2F20150824%2F&doc2=https%3A%2F%2Frawgit.com%2Fzolkis%2Fweb-nfc%2Fgh-pages%2Findex.html) to the [latest published version](http://w3c.github.io/web-nfc/releases/20150824/).

kenchris:
ok lgtm

anssiko:
Thank you @zolkis and @kenchris. This is a significant update that addresses the open issues. I will merge this now. Couple of comments that can be addressed later:
- In "The pushMessage algorithm", "NFCPushOptions dictionary", and "The watch() method" sections, _options_ is undefined. You want to "Let _options_ be ..."
- The pushMessage algorithm is rather long (by design). If parts of it could be factored out, that might help with readability. This is just an editorial comment, however.

zolkis:
> options is undefined. You want to "Let options be ..."

I thought formal parameters from the IDL don't need steps to explain their name binding.

> The pushMessage algorithm is rather long (by design)

Since often the algorithm is terminated in the sub-steps, it would be cumbersome to break it out to sub-algorithms, since then it's not clear what is meta-parameter and return values and what really belongs to the algorithm. I tried that slicing e.g. with the security steps, but it became a mess.

PR #46, opened 2015-09-07
Review fixes for #44.

zolkis:
Fixing comments for #44.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1667,39 +1667,11 @@
               </li>
 
               <li id="steps-map-url-to-ndef">
-                If <var>record</var>.kind is <code>"url"</code>, then
-                execute the <dfn>map URL to NDEF</dfn> sub-steps, or ensure
-                that the underlying platform provides equivalent values to
-                <var>ndef</var>:
-                <ul>
-                  <li>
-                    If <var>record</var>.data is not of type <code>string</code>,
-                    reject <var>promise</var> with
-                    <code>"SyntaxError"</code>, and abort these steps.
-                  </li>
-                  <li>
-                    Otherwise,
-                    <ul>
-                      <li>
-                        Set the <var>ndef.TNF</var> field to 1 (Well Known Type).
-                      </li>
-                      <li>
-                        Set the <var>ndef.TYPE</var> field to "U" (0x55
-                        following NFC binary encoding).
-                      </li>
-                      <li>
-                        Set the <var>ndef.PAYLOAD</var> field to
-                        <var>record</var>.data encoded according to the
-                        [[NFC-STANDARDS]], NFC Forum URI Record Type Definition
-                        specification: set the first octet according to the
-                        applicable scheme abbreviation, and set the rest of the
-                        payload bytes to <var>record</var>.data in UTF-8 encoding.
-                      </li>
-                    </ul>
-                  </li>
-                </ul>
+                If <var>record</var>.kind is <code>"url"</code>, then let <var>ndef</var>
+                the be result of passing <var>record</var> to <a>map a URL to NDEF</a>.
+                If this threw an exception, reject <var>promise</var> with that exception
+                and abort these steps.
               </li>
-
               <li id="steps-map-json-to-ndef">
                 If <var>record</var>.kind is <code>"json"</code>, then
                 execute the <dfn>map JSON to NDEF</dfn> sub-steps, or ensure
@@ -1843,7 +1815,7 @@
               </li>
             </ul>
           </li> <!-- write Web NFC record -->
-
+            
           <li>
             Make a request to the underlying platform to send <var>output</var>
             to the next device which comes in proximity range, and start a timer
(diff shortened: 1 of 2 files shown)


zolkis:
Thanks, see commit 3 in this PR.

index.html
```diff
@@ -1790,19 +1793,19 @@
           When an <a>NFC device</a> comes in communication range,
           <ul>
             <li>
-              If <var>options.target</var> has the value <code>"tag"</code>,
+              If <var>target</var> has the value <code>"tag"</code>,
```
anssiko: If the <var>options</var> object's <var>target</var> member has the value <code>"tag"</code>,

index.html
```diff
@@ -1790,19 +1793,19 @@
...
               then the sending should only happen if an <a>NFC tag</a> is
               tapped, otherwise reject <var>promise</var>
               <code>"InvalidAccessError"</code> and terminate this algorithm.
             </li>
             <li>
-              If <var>options.target</var> has the value <code>"peer"</code>,
+              If <var>target</var> has the value <code>"peer"</code>,
```
anssiko: Ditto.

index.html
```diff
@@ -1790,19 +1793,19 @@
...
               then the sending should only happen if an <a>NFC peer</a> is
               tapped, otherwise reject <var>promise</var> with
               <code>"InvalidAccessError"</code> and terminate this algorithm.
             </li>
             <li>
-              If <var>options.target</var> has the value <code>"any"</code>
+              If <var>target</var> has the value <code>"any"</code>
```
anssiko: Ditto.

index.html
```diff
@@ -2101,11 +2104,16 @@
           </p>
         </li>
         <li>
-          If the argument <var>options</var> contains invalid values,
+          If the first argument <var>options</var> contains invalid values,
           reject <var>promise</var> with a <code>"<a>SyntaxError</a>"</code>,
           and terminate this algorithm.
         </li>
         <li>
+          If the second argument <var>callback</var> is not an instance of
```
anssiko: You could say: "is not a Function" and link Function to the Terminology (Function is defined in http://heycam.github.io/webidl/#common-Function).

anssiko:
Thanks for the PR, I added some inline comments.

zolkis:
Thanks Anssi, I will fix them in later commits, don't merge until fixed. By now I will have to include them with other changes.

index.html
```diff
@@ -1418,6 +1461,9 @@
           </ul>
         </li>
         <li>
+          Let <var>options</var> be the second argument.
```
jyasskin: You don't need this, since you've introduced it with `pushMessage(message, options)` above.

index.html
```diff
@@ -1418,6 +1461,9 @@
...
+        </li>
+        <li>
           Let <var>timeout</var> be the value of <var>options.timeout</var> if
```
jyasskin: You can and should specify default values in the definition of [NFCPushOptions](https://rawgit.com/zolkis/web-nfc/gh-pages/index.html#idl-def-NFCPushOptions) (the argument type), instead of introducing local variables in the algorithm. See https://heycam.github.io/webidl/#dfn-dictionary-member-default-value.

For `timeout`, we might rather let UAs pick their default timeout, but for `target` it definitely makes sense to put it in the dictionary.

index.html
jyasskin: @zolkis is right for the current version of the document, since [step 9](https://rawgit.com/zolkis/web-nfc/2c04fe90c72064cb912a5f82329db1fa2033bf39/index.html#dfn-sending-nfc-content) introduces a local variable named _target_.

If it weren't a local variable, I've been using the convention `<code><var>options</var>.target</code>` to pull out members of dictionaries.

index.html
```diff
@@ -1317,6 +1361,34 @@
...
+  <section> <h3>The <strong>NFCPushTarget</strong> enum</h3>
+    <dl title="enum NFCPushTarget" class="idl">
+      <dt>tag</dt>
+      <dt>peer</dt>
+      <dt>any</dt>
+    </dl>
+  </section>
+
+  <section> <h3>The <strong>NFCPushOptions</strong> dictionary</h3>
+    <dl title="dictionary NFCPushOptions" class="idl">
```
jyasskin: Consider switching over to [contiguous IDL](http://www.w3.org/respec/examples/webidl-contiguous.html#introduction). It supports more of IDL and makes it easier to link things (no more `widl-long-string`).

Eventually, you might consider switching to [Bikeshed](https://github.com/tabatkins/bikeshed), which does a better job of cross-linking between specifications, but that's a fairly large investment.

index.html
anssiko: Good catch. I stand corrected. The algo is rather long to parse on a small screen.

index.html
anssiko: The more verbose style is consistent with e.g. the HTML spec. I suggest we prefer clarity over brevity. But I don't feel strongly, so leaving it to the editors.

index.html
anssiko: +1 on contiguous IDL, but I suggest we do not rush it in this release. IIRC it was @jyasskin's feature :-)

zolkis: I used to use contiguous IDL, until at some point in SysApps we started using the dl style. 
Anyway already in this PR I'd like to also include the full IDL as an Appendix (as contiguous IDL).

I will look into Bikeshed after this PR is landed, I hope (and likely) it's better than Anolis. Otherwise I used to prefer ReSpec - I like writing the page in an editor configured with autocomplete etc, and I like that I don't have to generate index.html every time I want to see the changes.

anssiko:
@jyasskin The editors believe [your feedback](https://lists.w3.org/Archives/Public/public-web-nfc/2015Sep/0033.html) has now been addressed. Could you please review this PR, and let @zolkis and @kenchris know if you'd like the editors to do further changes before merging. This version of the spec does not need to be perfect, but it should be a good base for the next iteration of the experimental implementation.

For convenience, here's the [RawGit HTML preview of the head of this PR](https://rawgit.com/zolkis/web-nfc/gh-pages/index.html).

Thank you again for your extensive and detailed feedback.

anssiko:
Since GH has issues showing the changes, you can use the [HTML diff](http://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fweb-nfc%2F&doc2=https%3A%2F%2Frawgit.com%2Fzolkis%2Fweb-nfc%2Fgh-pages%2Findex.html) to ease the review.

anssiko:
Will merge this PR now and suggest the editors address the remaining feedback in another PR.

PR #50, opened 2015-09-18
Improvements on visibility and focus; default values; type checks; match patterns; the usage of origin.

zolkis:
Following up on:
https://lists.w3.org/Archives/Public/public-web-nfc/2015Sep/0063.html

### DIFF
--- a/index.html
+++ b/index.html
@@ -180,8 +180,8 @@
   </p>
   <p>
     The term of executing algorithms
-    <a href="https://html.spec.whatwg.org/#in-parallel"><dfn>in parallel</dfn>
-    </a> is defined in [[!HTML5]].
+    <a href="https://html.spec.whatwg.org/#in-parallel"><dfn>in parallel</dfn></a>
+    is defined in [[!HTML5]].
   </p>
   <p>
     <a href="http://www.w3.org/TR/url-1/"><dfn>URL</dfn></a>
--- a/index.html
+++ b/index.html
@@ -1057,6 +1057,27 @@
     and <a>Writing or pushing content</a> sections.
     </p>
   </section> <!-- NFCRecord dictionary -->
+  <section> <h3>The <strong>NFCMessage</strong> dictionary</h3>
+    <p>
+      The content of any <a>Web NFC message</a> is exposed by the
+      <dfn>NFCMessage</dfn> dictionary:
+    </p>
+    <pre class="idl">
+      dictionary NFCMessage {
+        sequence&lt;NFCRecord&gt; data;
+        USVString url;
+      };
+    </pre>
+    <p>
+      The <dfn>NFCMessage.url</dfn>
+      property represents the <a>Web NFC Id</a> of the <a>Web NFC message</a>.
+    </p>
+    <p>
+      The <dfn>NFCMessage.data</dfn>
+      property represents the <a>NDEF message</a> defining the
+      <a>Web NFC message</a>.
+    </p>
+  </section>
 
   <section><h3>Data mapping</h3>
   <p>
@@ -1379,8 +1400,6 @@
   to handle incoming <a>Web NFC message</a>s either from an <a>NFC tag</a> or an
   <a>NFC peer</a>.
   <pre class="idl">
-    typedef sequence&lt;NFCRecord&gt; NFCMessage;
-
     interface NFCAdapter {
       Promise&lt;void&gt; pushMessage(NFCMessage message, NFCPushOptions options);
       void cancelPush(NFCPushTarget target);
@@ -1388,7 +1407,7 @@
       Promise&lt;void&gt; unwatch(long id);
     };
 
-    callback MessageCallback = void (NFCMessage message, USVString url);
+    callback MessageCallback = void (NFCMessage message);
   </pre>
 
   <section> <h3>The <strong>NFCPushOptions</strong> dictionary</h3>
@@ -1711,7 +1730,7 @@
           </li>
           <li>
             For each <a>NFCRecord</a> <var>record</var> in the sequence
-            <var>message</var>, execute the following steps, or make sure
+            <var>message</var>.data, execute the following steps, or make sure
             that the underlying platform provides equivalent values to
             <var>ndef</var>:
             <ol id="steps-map-nfcrecord-ndef">
@@ -1850,7 +1869,8 @@
 
             In addition, UAs MAY check that
             <var>record</var>.type is an IANA registered media type
-            for <a href="http://www.iana.org/assignments/media-types/media-types.xhtml#text">text</a>.
+            for <a href="http://www.iana.org/assignments/media-types/media-types.xhtml#text">
+            text</a>.
             If not, then the UA MAY throw a <code>"SyntaxError"</code>
             exception, and abort these steps.
           </li>
@@ -1863,7 +1883,8 @@
             the language code extracted from <var>record</var>.type.
 
             If <var>language</var> is not one of the language codes listed
-            in the <a href="http://www.iana.org/assignments/language-subtag-registry/language-subtag-registry">IANA language registry</a>
+            in the <a href="http://www.iana.org/assignments/language-subtag-registry/language-subtag-registry">
+            IANA language registry</a>
             (or [[ISO-639.2]]), then UAs MAY throw a <code>"SyntaxError"</code>
             exception, and abort these steps.
 
@@ -2374,11 +2395,10 @@
         </p>
       </li>
       <li>
-        Let <var>message</var> be a new empty sequence of <code>NFCRecord</code>
-        objects.
+        Let <var>message</var> be a new <code>NFCMessage</code> object.
       </li>
       <li>
-        Let <var>messageId</var> be <code>null</code>, used for storing the
+        Let <var>message</var>.url be <code>null</code>, used for storing the
         <a>Web NFC Id</a> of <var>message</var>.
       </li>
       <li>
@@ -2500,7 +2520,7 @@
           <li>
             If <var>ndef.TNF</var> is 4 (NFC External Type record), and
             <var>ndef.TYPE</var> is <code>urn:nfc:ext:w3.org:webnfc</code>,
-            then set <var>messageId</var> to the value decoded from
+            then set <var>message</var>.url to the value decoded from
             <var>ndef.PAYLOAD</var> in UTF-16.
           </li> <!-- parsing Web NFC record -->
           <li>
@@ -2525,10 +2545,12 @@
             Otherwise, skip to the next <a>NDEF record</a> in <var>input</var>.
           </li>
           <li>
-            Add <var>record</var> to <var>message</var>.
-            Set <var>NFCAdapter@[[\receivedMessage]]</var> to <var>message</var>.
+            Add <var>record</var> to <var>message</var>.data.
           </li>
         </ol>
+      </li>
+      <li>
+        Set <var>NFCAdapter@[[\receivedMessage]]</var> to <var>message</var>.
       </li>
       <li>
         If <var>NFCAdapter@[[\suspended]]</var> is <code>false</code>,
@@ -2544,8 +2566,7 @@
     </p>
     <ol id="steps-dispatch">
       <li>
-        Set <var>message</var> to be the
-        <var>NFCAdapter@[[\receivedMessage]]</var>.
+        Let <var>message</var> be <var>NFCAdapter@[[\receivedMessage]]</var>.
       </li>
       <li>
         For each <a>NFC watch</a> <var>watch</var> that has
@@ -2559,17 +2580,17 @@
           </li>
           <li>
             If <var>options</var>.mode is <code>"web-nfc-only"</code>, and
-            <var>messageId</var> is <var>null</var>, then skip to the next
+            <var>message</var>.url is <var>null</var>, then skip to the next
             <a>NFC watch</a>.
           </li>
           <li>
             Otherwise, if <var>options</var>.mode is
-            <code>"web-nfc-only"</code>, and <var>messageId</var> is not
+            <code>"web-nfc-only"</code>, and <var>message</var>.url is not
             <code>null</code>, or if <var>options</var>.mode is
             <code>"any"</code>, then execute the following sub-steps:
             <ol>
               <li>
-                If <var>messageId</var> does not match the <a>URL pattern</a>
+                If <var>message</var>.url does not match the <a>URL pattern</a>
                 stored in <var>options</var>.url, then skip to the next
                 <a>NFC watch</a>.
               </li>
@@ -2586,9 +2607,11 @@
               <li>
                 <a>Queue a task</a> using the <dfn>Web NFC task source</dfn>
                 to invoke the <code>MessageCallback</code>var>callback</var>
-                with <var>message</var> as its first argument, and
-                <var>messageId</var> as its second. Then set
-                <var>NFCAdapter@[[\receivedMessage]]</var> to <code>null</code>.
+                with <var>message</var> as its argument.
+              </li>
+              <li>
+                Set <var>NFCAdapter@[[\receivedMessage]]</var> to
+                <code>null</code>.
               </li>
             </ol>
           </li>
--- a/index.html
+++ b/index.html
@@ -1486,7 +1486,7 @@
             <var>message</var>, execute the following steps, or make sure
                 that the underlying platform provides equivalent values to
                 <var>ndef</var>:
-            <ul id="steps-map-nfcrecord-ndef">
+            <ol id="steps-map-nfcrecord-ndef">
               <li>
                 If <var>record</var>.kind is <code>"empty"</code>, then
                 <ol>
@@ -1586,53 +1586,58 @@
             <var>timer</var> MUST be cleared.
           </li>
           <li>
-            If <var>timer</var> exists and expires, reject <var>promise</var>
-            with <code>"TimeoutError"</code> and abort these steps.
-          </li>
-          <li>
-            When an <a>NFC device</a> comes within communication range,
-            <ol>
+            Wait until one of the following events happens:
+            <ul>
               <li>
-                If <var>target</var> has the value <code>"tag"</code>,
-                then the sending MAY only happen if an <a>NFC tag</a> is
-                tapped, otherwise reject <var>promise</var> with
-                <code>"InvalidAccessError"</code> and abort these steps.
+                The <var>timer</var> exists and expires, reject <var>promise</var>
+                with <code>"TimeoutError"</code> and abort these steps.
               </li>
               <li>
-                If <var>target</var> has the value <code>"peer"</code>,
-                then the sending MAY only happen if an <a>NFC peer</a> is
-                tapped, otherwise reject <var>promise</var> with
-                <code>"InvalidAccessError"</code> and abort these steps.
+                When an <a>NFC device</a> comes within communication range,
+                <ol>
+                  <li>
+                    If <var>target</var> has the value <code>"tag"</code>,
+                    then the sending MAY only happen if an <a>NFC tag</a> is
+                    tapped, otherwise reject <var>promise</var> with
+                    <code>"InvalidAccessError"</code> and abort these steps.
+                  </li>
+                  <li>
+                    If <var>target</var> has the value <code>"peer"</code>,
+                    then the sending MAY only happen if an <a>NFC peer</a> is
+                    tapped, otherwise reject <var>promise</var> with
+                    <code>"InvalidAccessError"</code> and abort these steps.
+                  </li>
+                  <li>
+                    If <var>target</var> has the value <code>"any"</code>, then the
+                    sending should happen if either an <a>NFC peer</a> or an
+                    <a>NFC tag</a> is tapped within the specified timeout.
+                  </li>
+                  <li>
+                    If the connection is lost during the transfer, or an error is
+                    signaled by the underlying platform, reject
+                    <var>promise</var> with <code>"<a>NetworkError</a>"</code> and
+                    terminate this algorithm.
+                  </li>
+                  <li>
+                    If the transfer is successful, cancel <var>timer</var> and
+                    resolve <var>promise</var>.
+                  </li>
+                </ol>
               </li>
               <li>
-                If <var>target</var> has the value <code>"any"</code>, then the 
-                sending should happen if either an <a>NFC peer</a> or an
-                <a>NFC tag</a> is tapped within the specified timeout.
+                If <var>timer</var> is still active, and the
+                <a href="cancelPush"><code>cancelPush()</code></a>
+                method is called, reject <var>promise</var> with
+                <code>"<a>AbortError</a>"</code>, as described in the
+                <a href="steps-cancelPush"><code>cancelPush()</code> algorithm</a>.
               </li>
               <li>
-                If the connection is lost during the transfer, or an error is
-                signaled by the underlying platform, reject
-                <var>promise</var> with <code>"<a>NetworkError</a>"</code> and
-                terminate this algorithm.
+                If the <a>top-level browsing context</a> loses focus (e.g. the user
+                navigated to another page), then pending push operations MUST be
+                canceled by executing the
+                <a href="steps-cancelPush"><code>cancelPush()</code></a> steps.
               </li>
-              <li>
-                If the transfer is successful, cancel <var>timer</var> and
-                resolve <var>promise</var>.
-              </li>
-            </ol>
-          </li>
-          <li>
-            If <var>timer</var> is still active, and the
-            <a href="cancelPush"><code>cancelPush()</code></a>
-            method is called, reject <var>promise</var> with
-            <code>"<a>AbortError</a>"</code>, as described in the
-            <a href="steps-cancelPush"><code>cancelPush()</code> algorithm</a>.
-          </li>
-          <li>
-            If the <a>top-level browsing context</a> loses focus (e.g. the user
-            navigated to another page), then pending push operations MUST be
-            canceled by executing the
-            <a href="steps-cancelPush"><code>cancelPush()</code></a> steps.
+            </ul>
           </li>
         </ol>
       </p>
--- a/index.html
+++ b/index.html
@@ -1447,9 +1447,13 @@
           <li>
             Let <var>output</var> be the notation for the <a>NDEF message</a>
             to be created by an API call to the underlying platform, based on
-            data provided by these steps. On any termination of this algorithm,
-            <var>output</var> SHOULD be cleared, so that the platform resources
-            are available for the next invocation of the algorithm.
+            data provided by these steps.
+            <p class="note">
+              On any termination of this algorithm, all platform resources
+              bound to <var>output</var> should be cleared, so that the
+              platform resources are available for the next invocation of the
+              algorithm.
+            </p>
           </li>
           <li>
             For each <a>NFCRecord</a> <var>record</var> in the sequence
@@ -1549,7 +1553,7 @@
             Wait until one of the following events happens:
             <ul>
               <li>
-                The <var>timer</var> exists and expires, reject <var>promise</var>
+                The <var>timer</var> expires, reject <var>promise</var>
                 with <code>"TimeoutError"</code> and abort these steps.
               </li>
               <li>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -97,6 +98,7 @@
     table.simple td {
         border-right: 1px solid #000;
     }
+    .issue[id^='issue-'] > *:not([role='heading']) { /* display: none; */ }
```
jyasskin: Looks like this isn't doing anything at the moment. Remove it?

kenchris:
Lgtm

index.html
```diff
@@ -926,26 +920,26 @@
     </p>
     <ul>
       <li>
-        Only <a>browsing context</a>s with a defined <a>origin</a>, whose schema
-        is starting with <code>"https"</code> must be able to access
+        Only <code>
+        <a href="https://html.spec.whatwg.org/#document">Document</a></code>s
```
jyasskin, line 931: It's not really the Document that's accessing anything. It's script in an execution environment (https://html.spec.whatwg.org/#script-execution-environment).

index.html
```diff
@@ -1100,6 +1070,12 @@
   </p>
   </section> <!-- NFCRecordType interface -->
 
+  <section><h3>The <strong>NFCRecordData</strong> type</h3>
+    <pre class="idl">
+      typedef (DOMString or unrestricted double or object or ArrayBuffer) NFCRecordData;
```
jyasskin: Totally optional, but you can just embed this typedef in the same IDL block as `NFCRecord`. It doesn't really need its own section.

index.html
```diff
@@ -1055,25 +1044,6 @@
...
-    <p>
-      A <dfn>valid NFCRecord</dfn> <var>record</var> is an instance of
-      <code><a>NFCRecord</a></code> which matches the following criteria:
-      <ul>
-        <li>
-          <var>record</var>.kind has values defined by
-          <code>NFCRecordType</code>, or <code>undefined</code>.
-        </li>
-        <li>
-          <var>record</var>.type is a <a>valid MIME type</a>, or
```
jyasskin, line 1067: The "valid MIME type" restriction was a real difference from what you can express in IDL. Did you mean to drop it?

index.html
```diff
@@ -1287,13 +1263,27 @@
         reject <var>promise</var> with
         <code>"<a>NotSupportedError</a>"</code>, and abort these steps.
       </li>
-          <li>Let <var>adapter</var> be <code>null</code>.</li>
+      <li>
+        Let <var>adapter</var> be <code>null</code>.
+        Associate <var>adapter</var> with the current <code>Window</code>
```
jyasskin: "Associate _adapter_ with" generally means the object _adapter_ references. In this case, that's always `null`. Do you mean that every `Window` object has an associated NFC adapter, initially `null`?

index.html
```diff
@@ -2494,34 +2551,90 @@
...
+      and if they expire, the event should be recorded and handled
+      next time the execution is resumed, i.e. when the <code>focus</code>
+      event is fired on the <code>Window</code> object.
+    </p>
+    <p>
+      When the <code>Window</code> object associated with the
+      <code><a>Document</a></code> using the Web NFC API
+      <a href="http://www.w3.org/TR/page-visibility/#now-visible-algorithm">
+      becomes visible</a> and
+      <a href="https://html.spec.whatwg.org/#handler-onfocus">in focus</a>,
```
jyasskin: This is the wrong link for "in focus". I'm looking for the right one. Current leads are the `:focus` pseudo-class and https://html.spec.whatwg.org/#currently-focused-area-of-a-top-level-browsing-context

jyasskin: It looks like https://html.spec.whatwg.org/#focus may be the best link. There's also https://html.spec.whatwg.org/#focused, but that only applies to elements, while you want this to apply to documents, windows, or browsing contexts.

index.html
zolkis: It strips off pasting in the text taken from the issue description, but allows to provide own text description. With `display: none` it only links in the issue, without any text. At least I'd prefer keeping it this way, or of course I could remove the commented part.

index.html
zolkis, line 931: Yes, we should be pedantic. We need a Window because the visibility state and focus, need a Document for origin, and a script execution environment. This ought to be possible to express in a more concise way, but anyway you are right.

index.html
zolkis, line 1067: Yes. Not all branches in algorithms require it being a valid MIME type, and the checks are done in the specific steps.

index.html
zolkis: Right, I wanted to express that implementations associate the variable to the Window object, initially null. The reason was the suspended state of the adapter is affected by the calling script's Document's associated Window object's visibility and focus state. Perhaps I could even remove the sentence, without inducing uncertainty, since later steps will anyway need this association. So, we should not go into implementation details when not strictly needed.

index.html
zolkis: Thank you!

index.html
zolkis: Should we do the same with NFCRecordType enum as well?

anssiko: Yes. Please put `typedef NFCRecordData` above `dictionary NFCRecord` within the same IDL block as suggested. It will improve readability and is aligned with the platform conventions.

The same with `enum NFCRecordType`, move it above the dictionary within the same IDL block where it is used.

For consistency and to conform with the conventions, also move `callback MessageCallback` _below_ `interface NFCAdapter` within the same IDL block. I'd also move `typedef NFCMessage` above that interface too.

Also merge `enum NFCPushTarget` and `dictionary NFCPushOptions`.

Consider merging `enum NFCWatchMode` and `dictionary NFCWatchOptions` if you think that'd improve readability (i.e. consider how the IDL and prose work together vs. split separate).

(Merging definitions was hard or impossible with ReSpec before Contiguous IDL support, thus many specs keep these separate. That is due to a tool limitation.)

kenchris: OK I can look at this, @zolkis

index.html
```diff
@@ -97,6 +98,7 @@
     table.simple td {
         border-right: 1px solid #000;
     }
+    .issue[id^='issue-'] > *:not([role='heading']) {}
```
jyasskin: This CSS rule has no properties. How does it "strip off ... the text taken from the issue description"? `document.querySelector(".issue[id^='issue-']")` also returns nothing, so the rule doesn't match anything anyway.

index.html
```diff
@@ -1287,13 +1243,25 @@
         reject <var>promise</var> with
         <code>"<a>NotSupportedError</a>"</code>, and abort these steps.
       </li>
-          <li>Let <var>adapter</var> be <code>null</code>.</li>
+      <li>
+        Let <var>adapter</var> be <code>null</code>.
+        <p class="note">
+          There is a different <code>NFCAdapter</code> instance per
+          <code>Window</code> object, whose
```
jyasskin: This "whose" refers to the `NFCAdapter`, but I think you mean that the Window's visibility and focus state determine things.

index.html
```diff
@@ -1332,6 +1300,33 @@
           </li>
         </ol>
       </li>
+      <li id="adapter-inbox">
+        Let <var>receivedMessage</var> be associated with <var>adapter</var>,
```
jyasskin: We use ["internal slots"](http://ecma-international.org/ecma-262/6.0/#sec-object-internal-methods-and-internal-slots) to refer to implementation-accessible attributes of Javascript objects. Streams and Web Bluetooth are [using the `adapter@[[slot]]` convention](https://webbluetoothcg.github.io/web-bluetooth/#terminology) to refer to these.

[Streams uses tables to define the internal slots of particular objects](https://streams.spec.whatwg.org/#rs-internal-slots), and Web Bluetooth will probably pick that up.

index.html
```diff
@@ -2188,9 +2188,12 @@
             <code>"NotSupportedError"</code> , and abort these steps.
           </li>
           <li>
-            Otherwise, let <var>watchId</var> be a locally unique number
-            assigned to <var>options</var> and <var>callback</var> and store
-            them together as an <a>NFC watch</a>.
+            Let <var id="watch-suspended">suspended</var> be <code>false</code>.
```
jyasskin: This is probably wrong, because it'll override the assignment from "Handling visibility and focus". I think you don't want a `suspended` variable _at all_. Instead, each place it's used, just check whether the window ([responsible browsing context](https://html.spec.whatwg.org/multipage/webappapis.html#responsible-browsing-context)?) is visible and focused.

index.html
```diff
@@ -1975,42 +2069,46 @@
 
     <section id="match_pattern"> <h3>Match patterns</h3>
       <p>
-        A match pattern is a string which can contain the
+        A <a>match pattern</a> is a string which can contain the
         <code>'*'</code> wildcard character.
```
jyasskin: Nit: You can probably just say "A match pattern is defined by the following ABNF."

index.html
```diff
@@ -1975,42 +2069,46 @@
...
-        Match patterns are used primarily for matching <a>IANA media type</a>s,
-        for instance the match pattern <code>'application/*+json'</code> matches
+        It is defined by the following ABNF:
+        <pre>
+          <dfn data-lt="match pattern">match-pattern</dfn> = <a>top-level type name</a> <code>"/"</code> [ tree <code>"."</code> ] <a>subtype name</a> [ <code>"+"</code> suffix ] [ <code>";"</code> parameters ]
+          <dfn>top-level type name</dfn> = <code>"*"</code> | &lt;any char except <code>"/"</code> and <code>"*"</code>&gt;
+          <dfn>subtype name</dfn>        = <code>"*"</code> | &lt;any char except <code>"+"</code>&gt;
+        </pre>
+        A <a>match pattern</a> is used primarily for matching <a>IANA media type</a>s,
```
jyasskin: It's used _only_ for matching media types.

There's also no algorithm for checking whether it matches... Say it's a glob?

index.html
```diff
@@ -1975,42 +2069,46 @@
...
+      <p class="note">
+        The <code>'*'</code> character has slightly different meaning depending
+        on whether it is in the <i>scheme</i>, <i>host</i> or <i>path</i> part.
         In a <i>scheme</i> <code>'*'</code> will match only the allowed
         <code>'https'</code> scheme.
         If the <i>host</i> is <code>'*'</code>, then it matches any host. If the
         <i>host</i> is e.g. <code>'*.myhost'</code>, then it will match the
-        specified host or any of its subdomains.
+        specified host or any of its subdomains. In the <i>host</i>
+        part, <code>'*'</code> MUST be the first character, so it can only be
```
jyasskin: You can't use MUST to constrain user input. Instead, say how it matches or fails to match if the \* is somewhere other than the first character.

index.html
```diff
@@ -1565,83 +1612,129 @@
...
           </li>
           <li>
-            If <var>timer</var> exists and expires, reject <var>promise</var>
-            with <code>"TimeoutError"</code> and abort these steps.
+            If <var>target</var> has the value <code>"peer"</code> or
+            <code>"any"</code>, then set the <var>peerMessage</var> variable
+            <a href="#adapter-peermessage">associated</a> with the current
+            <a>NFC adapter</a> to <var>output</var>.
+          </li>
+          <li id="adapter-timer">
```
jyasskin: This is the second element with this `id`.

index.html
```diff
@@ -926,26 +925,26 @@
...
-        that UAs should make sure the display is on, and the device is unlocked.
+        In order to use NFC, the <code>Window</code> object associated with the
+        <code><a>Document</a></code> contained by the
+        <a href="http://www.w3.org/TR/html5/browsers.html#top-level-browsing-context">
+        top level browsing context</a> using
+        the Web NFC API must be
+        <a href="http://www.w3.org/TR/page-visibility/#now-visible-algorithm">
+        visible</a> and in
+        <a href="https://html.spec.whatwg.org/#focus">focus</a>.
+        This also means that UAs should make sure the display is on, and the
```
jyasskin: "make sure the display is on and the device is unlocked" could mean that the UA should turn on the display and unlock the device. Instead, perhaps "This also means that UAs should block access to the NFC radio if the display is off or the device is locked."

index.html
```diff
@@ -1389,6 +1381,82 @@
...
+        <dfn>NFCWatchOptions</dfn> dictionary is used:
+      </p>
+      <pre class="idl">
+        enum NFCWatchMode {
+          "web-nfc-only",
+          "any"
+        };
+
+        dictionary NFCWatchOptions {
+          USVString url = null;
```
jyasskin: I think the convention for dictionary fields is to leave them `undefined` when they're not present, if the only other default would be `null`. I'm not sure of that, so no need to change it, but this may come up later.

I am pretty sure that if `null` is an allowed value, the type needs to be [`USVString?`](https://heycam.github.io/webidl/#idl-nullable-type).

index.html
```diff
@@ -1389,6 +1381,82 @@
...
+          USVString type = null;
+          NFCWatchMode mode = "web-nfc-only";
+        };
+      </pre>
+      <p>
+        The <dfn>NFCWatchOptions.url</dfn> property
+        denotes the <a>URL pattern</a> which is used for matching the
+        URL <a>origin</a> and <a>URL path</a> of the <a>Web NFC message</a>
+        which is stored in the <a>Web NFC record</a>.
+        The values <code>null</code>, <code>undefined</code>, and
```
jyasskin: On the other hand, `""` as a default could make sense. Then you'd avoid having 3 representations of the same thing. https://heycam.github.io/webidl/#es-to-dictionary would convert both absent fields and explicit `undefined`s to `""`.

index.html
```diff
@@ -1389,6 +1381,82 @@
...
+      </p>
+      <p>
+        The <code>"any"</code> value means that all <a>NDEF message</a>s
+        are exposed.
+      </p>
+      <pre
+        title="Filter accepting only JSON content from https://www.w3.org"
+        class="example highlight">
+        var watchOptions = {
+          url: "www.w3.org/*",  // any path from the domain is accepted
```
jyasskin: Are you missing a scheme here? The definition of `url` says it's an origin and path, and origins have schemes.

index.html
```diff
@@ -1565,83 +1612,129 @@
...
-                If the transfer is successful, cancel <var>timer</var> and
-                resolve <var>promise</var>.
+                When an <a>NFC device</a> <var>device</var> comes within
+                communication range,
+                <ol>
+                  <li>
+                    If <var>suspended</var> variable
+                    <a href="#adapter-suspended">associated</a> with the current
+                    <a>NFC adapter</a> is <code>true</code>, then
+                     reject <var>promise</var> with
```
jyasskin: This looks like an information leak: while the page is in the background, it learns the instant an NFC radio comes nearby. Instead, maybe just ignore any nearby devices until the page is focused again?

index.html
```diff
@@ -1565,83 +1612,129 @@
...
+                    <var>promise</var> with <code>"InvalidAccessError"</code>
+                    and abort these steps.
+                  </li>
+                  <li>
+                    If <var>device</var> is an <a>NFC peer</a>, and
+                    <var>peerMessage</var> is <code>null</code>, then  reject
+                    <var>promise</var> with <code>"InvalidAccessError"</code>
+                    and abort these steps.
+                  </li>
+                  <li>
```
jyasskin: I'd still appreciate a "transfer the data" step in here.

index.html
```diff
@@ -1939,29 +2030,32 @@
...
             following sub-steps:
-            <ul>
+            <ol>
               <li>
-                Remove the message that has been set for pushing.
+                Reset <var>timer</var>.
               </li>
               <li>
-                Cancel the timer associated with <code>messagePush()</code>.
+                Set the <var>tagMessage</var> variable
```
jyasskin: Why doesn't `pushMessage()` reset these variables?

index.html
```diff
@@ -1565,83 +1612,129 @@
...
-            If <var>timer</var> exists and expires, reject <var>promise</var>
-            with <code>"TimeoutError"</code> and abort these steps.
+            If <var>target</var> has the value <code>"peer"</code> or
+            <code>"any"</code>, then set the <var>peerMessage</var> variable
+            <a href="#adapter-peermessage">associated</a> with the current
+            <a>NFC adapter</a> to <var>output</var>.
+          </li>
+          <li id="adapter-timer">
+            If <var>options</var>.timeout value is not <code>Infinity</code>,
+            start the timer <var>timer</var>
```
jyasskin: Since 2 copies of this algorithm could be running at any given time (for "peer" and "tag" targets), you may need 2 timers. Or you could just treat the timer as a local variable here, and fix up references to it from elsewhere in the spec to talk about active `pushMessage()` executions.

index.html
```diff
@@ -1939,29 +2030,32 @@
...
-            <a href="#steps-pushMessage"><code>pushMessage()</code> algorithm</a>
-            with <code>"<a>AbortError</a>"</code>.
+              <li>
+                Set the <var>peerMessage</var> variable
+                <a href="#adapter-peermessage">associated</a> with the current
+                <a>NFC adapter</a> to <code>null</code>.
+              </li>
+              <li>
+                Reject the pending <var>promise</var> in the
+                <a href="#steps-pushMessage"><code>pushMessage()</code></a>
```
jyasskin: This is a little awkward, but not bad enough that it absolutely needs to be fixed. The timer isn't intrinsically connected to any particular executing `pushMessage()` execution, so strictly it doesn't make sense to refer to "the" pending promise from the pushMessage() algorithm: there could be 2. But that algorithm also rejects its promise when `cancelPush()` is called, and it's clear enough what you mean here.

index.html
```diff
@@ -1975,42 +2069,46 @@
...
+        <a data-lt="star path">path</a>, defined by the following ABNF:
         <pre>
-          <em>&lt;url-pattern&gt;</em> := <em>&lt;scheme&gt;</em>://<em>&lt;host&gt;</em><em>&lt;path&gt;</em>
-          <em>&lt;scheme&gt;</em> := '*' | 'https'
-          <em>&lt;host&gt;</em> := '*' | '*.' <em>&lt;any char except '/' and '*'&gt;</em>+
-          <em>&lt;path&gt;</em> := '/' <em>&lt;any chars&gt;</em>
+          <dfn data-lt="url pattern">url-pattern</dfn> = <a data-lt="star scheme">scheme</a> <code>"://"</code> <a data-lt="star host">host</a> <a data-lt="star path">path</a>
+          <dfn data-lt="star scheme">scheme</dfn>      = <code>"*"</code> | <code>"https"</code>
+          <dfn data-lt="star host">host</dfn>        = <code>"/"</code> &lt;any char&gt;
+          <dfn data-lt="star path">path</dfn>        = <code>"*"</code> | <code>"*."</code> &lt;any char except <code>"/"</code> and <code>"*"</code>&gt;
```
jyasskin: `"*."` applies to hosts, not paths, right?

index.html
```diff
@@ -1975,42 +2069,46 @@
...
         </pre>
       </p>
-      <p>
-        Note that <code>'*'</code> has slightly different meaning depending on
-        whether it is in the <i>scheme</i>, <i>host</i> or <i>path</i> part.
+      <p class="note">
```
jyasskin, line 2184: This can't be a note unless it's normatively specified somewhere else, and I don't think it is. It could use a statement of how most characters match themselves, but '*' matches a sequence of characters.

index.html
```diff
@@ -2188,9 +2188,12 @@
             <code>"NotSupportedError"</code> , and abort these steps.
           </li>
           <li>
-            Otherwise, let <var>watchId</var> be a locally unique number
-            assigned to <var>options</var> and <var>callback</var> and store
-            them together as an <a>NFC watch</a>.
+            Let <var id="watch-suspended">suspended</var> be <code>false</code>.
+          </li>
+          <li>
+            Let <var>watchId</var> be a locally unique number
```
jyasskin: Following https://html.spec.whatwg.org/multipage/webappapis.html#timer-initialisation-steps, "a number that will identify this NFC watch"?

index.html
```diff
@@ -2292,31 +2295,29 @@
...
+    <ol id ="parse-ndef">
       <li>
         Let <var>input</var> be the notation for the <a>NDEF message</a>
         which has been received by the underlying platform.
         <p class="note">
           This is a placeholder for a possible future feature. In this step,
-          implementations MAY check if the <a>origin</a> of the
-          <a>browsing context</a> matches the <a>Web NFC Id</a> of
-          <var>input</var>.If yes, and also <a>prearranged trust relationship</a>
+          implementations MAY check if the <a>origin</a> matches the
```
jyasskin: Which origin? Maybe the origin of the incumbent settings object like below?

index.html
```diff
@@ -2494,34 +2522,90 @@
             <code>"web-nfc-only"</code>, and <var>messageId</var> is not
```
jyasskin: Where is `messageId` initialized?

index.html
```diff
@@ -2466,23 +2467,50 @@
...
+            <code><a>NFCAdapter</a></code> object be <var>message</var>.
           </li>
-        </ul>
+        </ol>
+      </li>
+      <li>
+        If <var>suspended</var> variable
+        <a href="#adapter-suspended">associated</a> with the current
+        <a>NFC adapter</a> is <code>false</code>, then for each
+        <a>NFC watch</a> <var>watch</var> registered using the
```
jyasskin: `dispatch NFC content` also does its own iteration over NFC watches. You should only do it in one place.

index.html
```diff
@@ -2494,34 +2522,90 @@
...
+        <a href="#adapter-suspended">associated</a> with the current
+        <a>NFC adapter</a> to <code>false</code>.
+      </li>
+      <li>
+        If the <var>receivedMessage</var> variable
+        <a href="#adapter-inbox">associated</a> with the current
+        <a>NFC adapter</a> is not <code>null</code>, then for each
+        <a>NFC watch</a> <var>watch</var> registered using the
+        <code>watch()</code> method for the current <a>NFC adapter</a>,
+        execute the <a href="#steps-dispatch">dispatch NFC content</a>
```
jyasskin: This looks like it'll tell NFC watches about NFC devices that were nearby when the browsing context wasn't focused. It'll just delay telling them until the context gets focused again. I don't think just a delay mitigates the privacy risk.

jyasskin:
This is looking way better. Thanks!

index.html
kenchris: Right this is wrong

index.html
anssiko: I am the one to blame here :-) I introduced @zolkis to the ReSpec-GitHub issue integration (an undocumented feature AFAIK), and proposed the use of CSS to hide the issue description. By default, the issue that gets rendered contains also the first comment from the respective GH issue -- often too elaborate description to be included in the spec.

Now, the issue that made use of this feature seem to have been resolved and removed from the spec thus this CSS is currently not in use.

Anyway, given e.g. the following HTML:

`<div class="issue" data-number="51"></div>`

The rendered end result is something like:

```
<div class="issue" id="issue-51">
<div class="issue-title">
  <a href="https://github.com/w3c/web-nfc/issues/51"><span>Issue 51</span></a>:
  Create per-origin NFC adapter properties section?
</div>
</div>
```

Saves some editorial effort since no manual synching between the spec and GH issues is needed. Also, this way the GH issues can be used as the canonical source for issues (if so preferred). That said, I leave it to the editors to decide how to best manage the issues, i.e. inline or in GH issues or a combination of both.

anssiko:
@jyasskin Thanks for the thorough and detailed feedback, again.

@zolkis @kenchris When you feel you have satisfied all of the feedback, please feel free to merge this PR.

index.html
zolkis: In order to set up a watch, the page needs to get permission, and they do get a Promise. Then we need to respect the functionality: if the page is put in the background, there needs to be a feedback why the expected functionality (data transfer) does not happen. We can create an issue (i.e. needs more discussion) if you think this is not a good policy.

index.html
zolkis: The data transfer happens out of the control of the user agent. It is controlled by the underlying platform and the tapping gesture. The UA just sets it up and eventually gets feedback about the status. We could say "initiate data transfer".

index.html
zolkis: Moreover, the pushMessage() algorithm also rejects "the" pending promise of previous invocations with a conflicting target. So there is max one promise on pushMessage. Perhaps the language could still be improved, though.

index.html
zolkis: Thanks, this is a zombie from an earlier attempt. It was not the same 'suspended' as in the adapter, but it doesn't matter any more :).

index.html
zolkis: Good catch. Removed from here.

index.html
zolkis: It is initialized in the beginning of the steps, and it is updated when the Web NFC record is read. It is checked in the other steps.

index.html
zolkis: To my understanding, there is no privacy risk at this point. The user already gave permission to those pages when they have set up a watch. However, it may be incorrect from functional point of view to dispatch delayed messages.

zolkis: After a discussion with @alexshalamov it seems that for now we should implement silence during suspended state, i.e. no rejection of promises for push, and silent not-dispatch of received messages.

index.html
kenchris: Ok, changed

index.html
kenchris: Reworded

index.html
jyasskin: In most web specs, the "UA" is specified as if it included the underlying platform, and the spec describes the visible result instead of the interaction with the OS. E.g. http://tools.ietf.org/html/rfc6455#section-4.1 talks about the client opening a TCP connection, even though the underlying platform is actually the entity opening the connection.

However, you can use "initiate" and "wait for it to finish" here if you'd rather keep the separation.

index.html
jyasskin: Ah, I see. Then the `display:none` should come back, uncommented?

index.html
```diff
@@ -1269,6 +1227,36 @@
...
+     <tr>
+      <td>[[\tagMessage]]</td>
+      <td>An <a>NFCMessage</a> to hold the message to be pushed to a <a>NFC tag</a>, initially set to <code>null</code>.</td>
+     </tr>
+     <tr>
+      <td>[[\peerMessage]]</td>
+      <td>An <a>NFCMessage</a> to hold the message to be pushed to a <a>NFC peer</a>, initially set to <code>null</code>.</td>
+     </tr>
+
+   </td></tr></tbody></table>
```
jyasskin: This doesn't need the `</td></tr>`.

index.html
zolkis: On Tue, Sep 22, 2015 at 6:03 PM, Jeffrey Yasskin notifications@github.com
wrote:

> In index.html
> https://github.com/w3c/web-nfc/pull/50#discussion_r40097227:
> 
> > @@ -97,6 +98,7 @@
> >      table.simple td {
> >          border-right: 1px solid #000;
> >      }
> > -    .issue[id^='issue-'] > *:not([role='heading']) {}
> 
> Ah, I see. Then the display:none should come back, uncommented?

I have removed the whole line, since it's not used any more. Besides the
display:none would remove any text, but I wanted to be able to add my own
text for describing the issue. So it was good as it was when it had its use
- but again, now it's not needed any more.

index.html
```diff
@@ -1973,46 +2055,45 @@
...
-        <i>host</i> is e.g. <code>'*.myhost'</code>, then it will match the
-        specified host or any of its subdomains.
-        In the <i>path</i> part, each <code>'*'</code> matches 0 or more
-        characters.
+        Most characters match themselves, however <code>'*'</code> (wildcard)
+        matches a sequence of characters.
+      </p>
+      <p class="note">
+        Though <code>'*'</code> matches a sequence of characters, only URLs
+        which scheme <code>"https"</code> are accepted in the algorithms
```
jyasskin: s/which/with/

index.html
```diff
@@ -1973,46 +2055,45 @@
...
-        Note that <code>'*'</code> has slightly different meaning depending on
-        whether it is in the <i>scheme</i>, <i>host</i> or <i>path</i> part.
-        In a <i>scheme</i> <code>'*'</code> will match only the allowed
-        <code>'https'</code> scheme.
-        If the <i>host</i> is <code>'*'</code>, then it matches any host. If the
-        <i>host</i> is e.g. <code>'*.myhost'</code>, then it will match the
-        specified host or any of its subdomains.
-        In the <i>path</i> part, each <code>'*'</code> matches 0 or more
-        characters.
+        Most characters match themselves, however <code>'*'</code> (wildcard)
```
jyasskin, line 2180: "A URL pattern that doesn't obey the above ABNF matches nothing."?

index.html
jyasskin: Sure: https://github.com/w3c/web-nfc/issues/53

index.html
```diff
@@ -1565,83 +1613,122 @@
...
+          <li>
+            If <var>target</var> has the value <code>"peer"</code> or
+            <code>"any"</code>, then set <var>this@[[\peerMessage]]</var>
+            to <var>output</var>.
           </li>
           <li>
-            If <var>timer</var> exists and expires, reject <var>promise</var>
-            with <code>"TimeoutError"</code> and abort these steps.
+            If <var>options</var>.timeout value is not <code>Infinity</code>,
+            start the timer <var>this@[[\timer]]</var> with the timeout value
```
jyasskin: Since 2 copies of this algorithm could be running at any given time (for "peer" and "tag" targets), you may need 2 timers. Or you could just treat the timer as a local variable here, and fix up references to it from elsewhere in the spec to talk about active pushMessage() executions.

index.html
jyasskin: It's initialized at the beginning of the steps of a different algorithm. We don't use dynamic scoping in web algorithms. You need to pass in whatever variables you intend to use.

index.html
kenchris, line 1067: Should we remove the Valid MIME type in the term section then?

zolkis, line 1067: It is meant to be either a valid MIME type, or undefined. However, not all steps are required to check the validity. I would not remove it.

index.html
zolkis: Argh... this is the result of splitting the algorithms...

index.html
zolkis: We can indeed set up 2 push messages, but currently the second call will restart the timer with the new value. As an alternative, we could indeed keep 2 separate outboxes each with a separate timer, and remove the 'any' (default) value from NFCPushOptions.target. Actually I like that.

index.html
zolkis: I'll add messageId to NFCMessage as `id`.

index.html
zolkis: @alexshalamov and @anssiko also think the latter would be better. Proceeding.

zolkis:
Another proofreading needed, then it can be merged.

PR #59, opened 2015-10-01
Added section on NFCAdapter releasing.

zolkis:
Also, corrected examples, and fixed pushMessage() signature.


index.html
```diff
@@ -610,7 +610,7 @@
     });
 
     function processMessage(message) {
-      message.forEach((record) => {
+      message.data.forEach((record) => {
         if (record.kind == "string") {
           console.log(“Data is string: “ + data);
```
kenchris: what about data here?

PR #61, opened 2015-10-06
Improvements on internal slots and algorithms.

zolkis:


zolkis:
One more commit will follow a bit later. I want to check if we really need internal slots for timers and message buffers. Looks like internal slots caused a bit of confusion on implementation side, and want to minimize their usage.

zolkis:
Thanks, I am merging. At least this is simpler. If we can refer to running instances of algorithms, this should be fine. If there are problems with the formulation, we can fix them later.

PR #89, opened 2015-11-06
Add more examples and improve existing ones

kenchris:
Had to indent the pre contents because the stripping code isn't working when using pre inside aside. But this was the only way that I could get a nice non-pre description in the examples like in the Web Bluetooth spec.

I avoid using comments in the examples and add description above instead - this makes them much more readable.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1462,7 +1462,7 @@
 
         dictionary NFCWatchOptions {
           USVString url = "";
-          USVString recordType = "";
+          NFCRecordType? recordType;
           USVString mediaType = "";
           NFCWatchMode mode = "web-nfc-only";
         };
--- a/index.html
+++ b/index.html
@@ -609,106 +609,222 @@
     This section shows how developers can make use of the various features of
     this specification.
   </p>
-  <pre title="Push a text string to a peer device" class="example highlight">
-    // Specifying an option for pushing to peers only,
-    // and thus, do nothing when the user taps a tag.
-    navigator.nfc.push("Text meant for peers only", { target: "peer" })
-      .then(() => {
-        console.log("Message pushed.");
-      }).catch((error) => {
-        console.log("Push failed :-( try again.");
-      });
-  </pre>
-  <pre title="Push a text string to either a tag or peer"
-       class="example highlight">
-    // Push a text string (here: serialized JSON), with no options specified
-    // (target defaults to any device).
-    navigator.nfc.push('{ prop1: "value1"; prop2: "value2" }')
-      .then(() => {
-        console.log("Message pushed.");
-      }).catch((error) => {
-        console.log("Push failed :-( try again.");
-      });
-  </pre>
-  <pre title="Push a URL to either a tag or peer" class="example highlight">
-    // To push an NDEF record of URL type, use NFCMessage
+
+  <aside title="Push a text string to either a tag or peer"
+       class="example">
+    <p>
+      Pushing a text string (like the serialized JSON below)
+      to any kind of device is straight forward. Options can
+      be left out, as they default to pushing to both tags and
+      peers.
+    </p>
+    <pre class="highlight">
+navigator.nfc.push(
+    '{ "prop1": "value1", "prop2": "value2" }'
+  ).then(() => {
+    console.log("Message pushed.");
+  }).catch((error) => {
+    console.log("Push failed :-( try again.");
+  });
+    </pre>
+  </aside>
+
+  <aside title="Push a text string to a peer device" class="example">
+    <p>
+      It is possible to restrict to which devices (tags or peers) data
+      should be pushed. Below we specify to only push to peers,
+      and thus, no data is pushed when the user taps a tag.
+    </p>
+    <pre class="highlight">
+navigator.nfc.push(
+    "Text meant for peers only", { target: "peer" }
+  ).then(() => {
+    console.log("Message pushed.");
+  }).catch((error) => {
+    console.log("Push failed :-( try again.");
+  });
+    </pre>
+  </aside>
+
+  <aside title="Push a URL to either a tag or peer" class="example">
+    <p>
+      In order to push an NDEF record of URL type, simply use NFCMessage.
+    </p>
+    <pre class="highlight">
+navigator.nfc.push({
+    data: [{ recordType: "url", data: "https://w3c.github.io/web-nfc/" }]
+  }).then(() => {
+    console.log("Message pushed.");
+  }).catch((error) => {
+    console.log("Push failed :-( try again.");
+  });
+    </pre>
+  </aside>
+
+  <aside title="Read data from tag, and write to empty ones" class="example">
+    <p>
+      Below we read various different kinds of data which can be stored on a tag.
+      In the case the tag is empty, we write a text message with the value
+      "Hello World".
+    </p>
+    <pre class="highlight">
+navigator.nfc.watch((message) => {
+  if (message.data[0].recordType == 'empty') {
     navigator.nfc.push({
-        data: [{ recordType: "url", data: "/path/resource/" }];
-      }).then(() => {
-        console.log("Message pushed.");
-      }).catch((error) => {
-        console.log("Push failed :-( try again.");
-      });
-  </pre>
-  <pre title="Read and update data stored on tag" class="example highlight">
-    // Use default options for watch.
-    navigator.nfc.watch((message) => {
-      console.log('NDEF message received from URL ' + message.url);
-      if (message.data[0].recordType == 'empty') {
-        navigator.nfc.push({
-          url: "/mypath/myapp",  // path set by the application
-          data: [
-            {
-              recordType: "url",
-              mediaType: "",  // this line could be omitted
-              data: "Initial data"
-            }
-          ]});
-      }
-      processMessage(message);
-    }).then(() => { console.log("Watch added.");})
-    .catch((error) => { console.log("Adding watch failed: " + error.name )});
+      url: "/custom/path",
+      data: [{ recordType: "text", data: 'Hello World' }]
+    });
+  } else {
+    console.log('Read message written by ' + message.url);
+    processMessage(message);
+  }
+}).then(() => {
+  console.log("Added a watch.");
+}).catch((error) => {
+  console.log("Adding watch failed: " + error.name);
+});
 
-    function processMessage(message) {
-      message.data.forEach((record) => {
-        if (record.recordType == "string") {
-          console.log('Data is string: ' + record.data);
-        } else if (record.recordType == 'json') {
-          processJSON(record.data);
-        } else if (record.recordType == 'url') {
+function processMessage(message) {
+  for (let record of message.data) {
+    switch (record.recordType) {
+      case "string":
+        console.log('Data is string: ' + record.data);
+        break;
+      case "url":
         console.log('Data is URL: ' + record.data);
-        } else if (record.recordType == 'opaque') {
-          processBinary(record.data);
-      });
-    };
+        break;
+      case "json":
+        let value = JSON.parse(record.data);
+        console.log('JSON data: ' + value.myProperty.toString());
+        break;
+      case "opaque":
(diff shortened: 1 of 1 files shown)


zolkis:
In line 2509, could you please correct NfcWatchMode to NFCWatchMode?

index.html
```diff
@@ -609,106 +609,211 @@
...
+  });
+    </pre>
+  </aside>
 
-    function reader(message) {
-      console.log("Game state received from: " + message.url);
-      console.log("Stored state: " + message.data);
+  <aside title="Push a URL to either a tag or peer" class="example">
+    <p>
+      In order to push an NDEF record of URL type, simply use NFCMessage.
```
zolkis, line 651: nag: it's not simply, it is the more complex way of using push(), for the purpose to have control on the low level NDEF layout of the tag. That information got lost.

index.html
```diff
@@ -609,106 +609,211 @@
...
+  }).then(() => {
+    console.log("Message pushed.");
+  }).catch((error) => {
+    console.log("Push failed :-( try again.");
+  });
+    </pre>
+  </aside>
 
-      // Now do some calculations and then update the state.
+  <aside title="Read and update data stored on tag" class="example">
```
zolkis: This Example 4 still has the inline comments. Also Example 5. They should be cleansed, too.

index.html
```diff
@@ -609,106 +609,211 @@
...
+    <pre class="highlight">
+navigator.nfc.watch((message) => {
+  for (let record of message.data) {
+    let article =/[aeio]/.test(record.data.title) ? "an" : "a";
+    console.log(`$(record.data.name) is $(article) $(record.data.title)`);
+  }
+}, { url: document.baseURI });
+
+navigator.nfc.push({
+  data: [
```
zolkis, line 768: This won't work. push() takes an NFCMessage whose data is a sequence of NFCRecord objects, which have recordType, mediaType and data properties, but none of the below. So what happens is all the properties here get lost and the records will contain nothing.

index.html
```diff
@@ -609,106 +609,211 @@
...
+    console.log("MIME type:    " + record.mediaType);
+    console.log("=== data ===\n" + record.data);
+  }
+});
+
+navigator.nfc.push("Pushing data is fun!", { target: "tag", ignoreRead: false });
+    </pre>
+  </aside>
+
+  <aside title="Write Chinese text as UTF-8" class="example">
```
zolkis, line 808: nice example!

index.html
```diff
@@ -609,106 +609,211 @@
...
+  });
+};
+    </pre>
+  </aside>
+
+  <aside title="Push and read JSON (auto serialized and deserialized)"
+       class="example">
+    <p>
+      Storing and receiving JSON data is easy as serialization and deserialization
+      happens automatically.
```
zolkis: what do you mean here? It should not happen automatically.

index.html
```diff
@@ -609,106 +609,211 @@
...
-    function reader(message) {
-      console.log("Game state received from: " + message.url);
-      console.log("Stored state: " + message.data);
+  <aside title="Push a URL to either a tag or peer" class="example">
+    <p>
+      In order to push an NDEF record of URL type, simply use NFCMessage.
+    </p>
+    <pre class="highlight">
+navigator.nfc.push({
+    data: [{ recordType: "url", data: "/path/resource/" }];
```
alexshalamov: data: is invalid. Should be valid URL / IRI, RFC 3987. Step 5 in "11.3.1.5 Mapping URL to NDEF" will fail.

index.html
```diff
@@ -609,106 +609,211 @@
...
-          mediaType: "application/json",
-          data: { level: 3, points: 4500, lives: 3 }
-        }]
-      };
+  if (message.data[0].recordType == 'empty') {
+    navigator.nfc.push({
+      url: "/mypath/myapp",  // path set by the application
+      data: [{
+        recordType: "url",
+        data: "Initial data"
```
alexshalamov: data: is invalid, should be valid IRI / URI

index.html
```diff
@@ -609,106 +609,211 @@
...
+    } else if (record.recordType == 'url') {
+    console.log('Data is URL: ' + record.data);
+    } else if (record.recordType == 'opaque') {
+      processBinary(record.data);
+  });
+};
+
+function processBinary(data) {
+  console.log('Known (string) binary data: ');
+  console.log(String.fromCharCode.apply(null, new Uint16Array(data)));
```
alexshalamov: processBinary should check type and create proper view accordingly. Just creating Uint16 is misleading.
For example:

``` javascript
if (record.type == "image/png") {
  somePlaceholderImg.src = URL.createObjectURL(new Blob(record.data, record.type));
  somePlaceholderImg.onload = function() {
    URL.revokeObjectURL(this.src);
  }
}
```

index.html
```diff
@@ -609,106 +609,211 @@
...
+    </p>
+    <pre class="highlight">
+navigator.nfc.watch(reader, { url: "*/mypath/mygame/*" });
+
+function reader(message) {
+  console.log("Source:     " + message.url);
+  console.log("Game state: " + message.data);
+
+  var message = {
+    url: "https://company.com/path/game/update",  // app specific path
```
alexshalamov: https://company.com can be omitted

index.html
```diff
@@ -609,106 +609,211 @@
...
+    }
+  ]
+});
+    </pre>
+  </aside>
+
+  <aside title="Write data to tag and print out existing data" class="example">
+    <p>
+      Pushing data to a tag requires tapping it. If existing data should be
+      read during the same tab, we need to set the <code>ignoreRead</code>
```
alexshalamov: tab => tap

index.html
alexshalamov: Then we need to remove "object" type from NFCRecordData. At the moment spec implies that we have automatic serialization / deserialization for "objects".

We can also change "object" to "Dictionary" to reduce scope of object. Then it will be allowed to pass associative arrays. However, (de)serialization will be needed in any case.

zolkis: Currently serialization is automatic, deserialization manual (according to the push and read steps), so that should be fine. But I fully agree we should change object to Dictionary.

index.html
kenchris: but it is an example and it is already omitted elsewhere, so fine to add there, no?

index.html
kenchris: fixed, will push soon

index.html
alexshalamov: 10.7.1.9 Creating a Web NFC Id
origin + NFCMessage.url , should be valid url.

If you don't like it, we can add more steps, like:
1. id = origin;
2. If NFCMessage.url has "origin" and origin of id and NFCMessage.url are same
   - Then id = NFCMessage.url
3. If NFCMessage.url doesn't have "origin", then id = id + NFCMessage.url
   - If id is valid url, WebNFC Id = id
   - else, SyntaxError

index.html
kenchris, line 651: Suggesting to better wording please :)

index.html
```diff
@@ -609,106 +609,222 @@
...
+    </pre>
+  </aside>
+
+  <aside title="Push a URL to either a tag or peer" class="example">
+    <p>
+      In order to push an NDEF record of URL type, simply use NFCMessage.
+    </p>
+    <pre class="highlight">
+navigator.nfc.push({
+    data: [{ recordType: "url", data: "https://w3c.github.io/web-nfc/" }];
```
anssiko: Remove semicolon.

index.html
```diff
@@ -609,106 +609,222 @@
...
+      data: [{ recordType: "text", data: 'Hello World' }]
+    });
+  } else {
+    console.log('Read message written by ' + message.url);
+    processMessage(message);
+  }
+}).then(() => {
+  console.log("Added a watch.");
+}).catch((error) => {
+  console.log("Adding watch failed: " + error.name)
```
anssiko: Add semicolon.

index.html
```diff
@@ -609,106 +609,222 @@
...
+          var img = document.createElement("img");
+          img.src = URL.createObjectURL(new Blob(record.data, record.mediaType));
+          img.onload = function() {
+            window.URL.revokeObjectURL(this.src);
+          };
+        }
+        break;
+    }
+  });
+};
```
anssiko: Unnecessary semicolon.

index.html
```diff
@@ -609,106 +609,222 @@
...
+        if (record.mediaType == 'image/png') {
+          var img = document.createElement("img");
+          img.src = URL.createObjectURL(new Blob(record.data, record.mediaType));
+          img.onload = function() {
+            window.URL.revokeObjectURL(this.src);
+          };
+        }
+        break;
+    }
+  });
```
anssiko: Remove bracket and semicolon.

index.html
```diff
@@ -609,106 +609,222 @@
...
+      data: { level: 3, points: 4500, lives: 3 }
+    }]
+  };
+
+  navigator.nfc.push(message).then(() => {
+    console.log('Progress stored!');
+  }).catch((error) => {
+    console.log('Failure, please try again.');
+  });
+};
```
anssiko: Unnecessary semicolon.

index.html
```diff
@@ -609,106 +609,222 @@
...
+    <p>
+      In order to write a non-UTF-16 string, the data mush be written
+      as an <a>ArrayBuffer</a> and the charset must be set as part
+      of the <a>IANA media type</a>.
+    </p>
+    <pre class="highlight">
+let encoder = new TextEncoder([utfLabel = "utf-8"]);
+let utf8Text = encoder.encode("电脑坏了。");
+
+navigator.nfc.push(data: [
```
anssiko: navigator.nfc.push({ data: [

index.html
```diff
@@ -609,106 +609,222 @@
...
+  {
+    recordType: "text",
+    mediaType: "text/plain; lang=zh; charset=UTF-8;",
+    data: utf8Text.buffer;
```
anssiko: Remove semicolon.

index.html
```diff
@@ -609,106 +609,222 @@
...
+  }
+]);
```
anssiko: ] });

anssiko:
In general, looks like a great improvement to the examples section. I like specs (such as this one) that have a well thought out examples section that document common usages web developers are likely to face.

I added some comments inline.

index.html
```diff
@@ -609,106 +609,222 @@
...
+       class="example">
+    <p>
+      Pushing a text string (like the serialized JSON below)
+      to any kind of device is straight forward. Options can
+      be left out, as they default to pushing to both tags and
+      peers.
+    </p>
+    <pre class="highlight">
+navigator.nfc.push(
+    '{ "prop1": "value1"; "prop2": "value2" }'
```
alexshalamov: should be comma
'{ "prop1": "value1", "prop2": "value2" }'

x = JSON.parse('{ "prop1": "value1"; "prop2": "value2" }'); => exception

index.html
```diff
@@ -609,106 +609,222 @@
...
+      Storing and receiving JSON data is easy with serialization and deserialization.
+    </p>
+    <pre class="highlight">
+navigator.nfc.watch((message) => {
+  for (let record of message.data) {
+    let value = JSON.parse(record.data);
+    let article =/[aeio]/.test(value.title) ? "an" : "a";
+    console.log(`$(value.name) is $(article) $(value.title)`);
+  }
+}, { url: document.baseURI });
```
alexshalamov: url: document.baseURI, recordType: = "json"

index.html
```diff
@@ -609,106 +609,222 @@
...
+    </p>
+    <pre class="highlight">
+navigator.nfc.watch(reader, { url: "*/mypath/mygame/*" });
+
+function reader(message) {
+  console.log("Source:     " + message.url);
+  console.log("Game state: " + message.data);
+
+  var message = {
+    url: "mypath/mygame/update",
```
alexshalamov: :smile: first / is missing. According to RFC.
url: "/mypath/mygame/update",

zolkis:
lgtm, merging

PR #90, opened 2015-11-09
Change object to Dictionary. Modify receive steps to parse JSON content.

zolkis:
A small change for improving usability.


zolkis:
> I know the "throws an error" more than the past tense "threw an exception"

Should we also reformulate the other 8 occurrences of "threw an exception" that have been earlier used in the spec?

PR #95, opened 2015-12-22
Fix a wrong use of NFCPushMessage

zqzhang:
As a typedef, here parameter in type of NFCPushMessage shall be
an instance of NFCMessage.


zolkis:
Thank you for these fixes!

PR #111, opened 2017-04-24
Throw TypeError when invalid type passed to push function

### LINKED ISSUE #110: Throw TypeError when invalid type is provided to nfc.push function
SyntaxError should be used when there is a parsing error, invalid syntax, tokens.

https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-syntaxerror
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SyntaxError

In cases when invlid type is provided to the function, TypeError should be thrown

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypeError
https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror

alexshalamov:
This PR changes exception code that is thrown by push function when invalid type is provided.

Fixes #110

### DIFF
--- a/index.html
+++ b/index.html
@@ -1685,16 +1685,15 @@ navigator.nfc.push({ data: [
             Let <var>timeout</var> be <var>options</var>.timeout.
           </li>
           <li>
-            If the type of the <var>message</var> parameter is not
-            <code>DOMString</code> or <code>ArrayBuffer</code>, and it is not an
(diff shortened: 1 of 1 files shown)


zolkis:
LGTM

PR #115, opened 2017-05-10
Update exception codes for 'Creating Web NFC message' operation

### LINKED ISSUE #116: 'Creating Web NFC message' algorithm should throw correct exceptions
The 'Creating Web NFC message' algorithm should throw TypeError exception when invalid type is provided to the API. In cases when invalid syntax or tokens are provided, SyntaxError exception should be thrown.

alexshalamov:
This PR updates 'Creating Web NFC message' algorithms, so that,
when invalid type is provided to the API, TypeError exception
will be thrown. In cases when invalid syntax or tokens are
provided, SyntaxError exception will be thrown.

Fixes #116

### DIFF
--- a/index.html
+++ b/index.html
@@ -1917,28 +1917,29 @@ navigator.nfc.push({ data: [
                 If <var>record</var>.recordType is <code>undefined</code>, then:
                 <ol> <!-- guess type and mediaType from data -->
                   <li>
-                    If <var>record</var>.data is instance of
-                    <code>ArrayBuffer</code>, then
-                    set <var>record</var>.mediaType to
-                    <code>"application/octet-stream"</code> and
-                    <var>record</var>.recordType to <code>"opaque"</code>.
+                    If the type of <var>record</var>.data is an
+                    <code>ArrayBuffer</code>, set <var>record</var>.recordType
+                    to <code>"opaque"</code> and if <var>record</var>.mediaType
+                    is an empty string, set <var>record</var>.mediaType to
+                    <code>"application/octet-stream"</code>.
+                  </li>
+                  <li>
+                    Otherwise, if the type of <var>record</var>.data is an
+                    <code>Object</code>, set <var>record</var>.recordType
+                    to <code>"json"</code> and if <var>record</var>.mediaType
+                    is an empty string, set <var>record</var>.mediaType to
+                    <code>"application/json"</code>.
                   </li>
                   <li>
                     Otherwise, if the type of <var>record</var>.data is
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1917,28 +1917,29 @@
                 If <var>record</var>.recordType is <code>undefined</code>, then:
                 <ol> <!-- guess type and mediaType from data -->
                   <li>
-                    If <var>record</var>.data is instance of
-                    <code>ArrayBuffer</code>, then
-                    set <var>record</var>.mediaType to
-                    <code>"application/octet-stream"</code> and
-                    <var>record</var>.recordType to <code>"opaque"</code>.
+                    If the type of <var>record</var>.data is an
```
kenchris, line 1920: Maybe we should use Type(value) like in the manifest spec: https://www.w3.org/TR/appmanifest/#dfn-type

index.html
```diff
@@ -1917,28 +1917,29 @@
...
+                    <code>"application/json"</code>.
                   </li>
                   <li>
                     Otherwise, if the type of <var>record</var>.data is
-                    <code>"number"</code> or <code>"string"</code>, then
-                    set <var>record</var>.mediaType to <code>"plain/text"</code>
-                    and <var>record</var>.recordType to <code>"text"</code>.
+                    <code>UnrestrictedDouble</code> or <code>String</code>, then
+                    set <var>record</var>.recordType to <code>"text"</code>
+                    and if <var>record</var>.mediaType is an empty string,
```
kenchris, line 1937: Why not make this a sub item. Will probably be easier to follow

index.html
alexshalamov: The spec is using IDL types for interface definition, so, it validates those against IDL types. ToType(x) or Type(x) are ecma bindings.

index.html
alexshalamov: I tried that, was looking funny, every step had just 1 substep. Do you want me to upload version with substeps?

alexshalamov: Looks like this :D
```
2. For each NFCRecord record in the sequence message.data, run the following steps, or make sure that the underlying platform provides equivalent values to ndef:
  1. If record.recordType is undefined, then:
    1. If the type of record.data is an ArrayBuffer, set record.recordType to "opaque".
      1. If record.mediaType is an empty string, set record.mediaType to "application/octet-stream".
```

kenchris: :) OK lets drop it then

kenchris:
lgtm

PR #119, opened 2017-05-19
Allow to use '*' wildcard for path matching

### LINKED ISSUE #118: URL pattern ABNF notation doesn't match examples
Example uses '*://*.mydomain.com/*' pattern, while ABNF notation only supports "/" <VCHAR> as a path component.

alexshalamov:
Fixes #118


alexshalamov:
@zolkis @kenchris please take a look.

zolkis:
+1

zolkis (approved):

alexshalamov:
Just figured out that path part should be 
`path = "/*" or "/" VCHAR`
Will fix that on Monday.

Also, need to add note, there is ambiguous behavior for * wildcard in path. The asterisk is valid symbol for URL path, thus 'https://site.com/*' pattern will match 'https://site.com/*' and 'https://site.com/path' URLs.

PR #121, opened 2017-05-22
Make url path component optional for watch patterns

### LINKED ISSUE #120: Make path component for watch url pattern optional
Specification states that path component is optional, yet, ABNF notation specifies it as mandatory.

alexshalamov:
Fixes #120


alexshalamov:
@kenchris @zolkis please take a look.

zolkis:
+1

PR #123, opened 2017-05-24
Validate URL pattern that is passed to watch method

### LINKED ISSUE #122: If invalid URL pattern is provided to watch() method, reject promise with SyntaxError
At the moment, if invalid URL pattern is passed to watch() method, API provides no indication about this error condition. It would be good to inform developer immediately, otherwise 'dispatching NFC content' would not work without valid reason.

alexshalamov:
Fixes #122

### DIFF
--- a/index.html
+++ b/index.html
@@ -2495,6 +2495,12 @@ navigator.nfc.push({ data: [
             <code>"NotSupportedError"</code> and abort these steps.
           </li>
           <li>
+            If the <var>options</var>.url is not an empty string, validate
+            <var>options</var>.url against <a>URL pattern</a> syntax.
+            If validation fails, reject <var>promise</var> with
+            <code>"SyntaxError"</code> and abort these steps.
+          </li>
+          <li>
             Let <var>watchId</var> be a number that will identify this
             <a>NFC watch</a>.
           </li>


alexshalamov:
@zolkis @kenchris could you please take a look?

zolkis:
When we say "validate", I think we need to link to an algorithm.

index.html
```diff
@@ -2495,6 +2495,12 @@
             <code>"NotSupportedError"</code> and abort these steps.
           </li>
           <li>
+            If the <var>options</var>.url is not an empty string, validate
```
kenchris, line 2498: Shouldnt you link to how to validate?

alexshalamov: Sure, would add validation algorithm.

alexshalamov:
@zolkis @kenchris how strict (complex) validation should be? Since URL pattern is not a valid URL, do we want to re-define URL parser in the spec?

kenchris:
It probably has to be quite strict. Maybe we could monkey patch? @anssiko

alexshalamov:
@kenchris then we need to write ABNF for 'NFC URL pattern' that should be as complex as the one for URL's

At the moment, valid 'NFC URL pattern' could be https://site.com:test/path , while this is not a valid URL.

kenchris:
Yeah, for interoperability this is probably what we need to do. Not sure if there are any shortcuts or it something similar is done elsewhere in the platform

alexshalamov:
Created #124, same functionality could be achieved without rewriting URL parser or defining custom ABNF notation.

anssiko:
You might be able to use https://url.spec.whatwg.org/#url-parsing

alexshalamov:
@anssiko yes, if we simplify nfc url patterns (#124). With current ABNF notation, url parsers would fail to validate / parse the pattern.

zolkis:
Let's go with simplifying patterns. This is a good example why discipline/harmonization of web specs is needed, now we could reuse an existing algorithm.

PR #125, opened 2017-05-31
Simplify URL pattern matching concept

### LINKED ISSUE #122: If invalid URL pattern is provided to watch() method, reject promise with SyntaxError

### LINKED ISSUE #124: Simplify 'URL pattern' concept for watch operation
We can simplify URL pattern concept, while maintaining same functionality.

- Wildcard would be only allowed for path.
- Protocol is fixed to https or any other allowed scheme.
- Domain matches domain and all subdomains, specifying subdomain will reduce scope to subdomain level.

Therefore, URL pattern would be a valid URL that can be tokenized / split to components with existing URL parsers.

alexshalamov:
This PR does following:

- Uses valid URL for URL pattern matching
- Validates NFCWatchOptions.url using standard URL validation rules
- Defines URL pattern matching algorithm that was missing from spec
- Uses new algo from 'Dispatching NFC content' operation

Fixes #122
Fixes #124
Closes #123

### DIFF
--- a/index.html
+++ b/index.html
@@ -151,9 +151,17 @@
 <!-- - - - - - - - - - - - - - -  Terminology - - - - - - - - - - - - - - - -->
 <section> <h2>Terminology and conventions</h2>
   <p>
-    The terms <a href="http://www.w3.org/TR/url-1/"><dfn>URL</dfn></a> and
+    The terms <a href="http://www.w3.org/TR/url-1/"><dfn>URL</dfn></a>,
+    <a href="https://url.spec.whatwg.org/#concept-url-scheme">
+    <dfn>URL scheme</dfn></a>,
+    <a href="https://url.spec.whatwg.org/#concept-url-host">
+    <dfn>URL host</dfn></a>,
     <a href="https://url.spec.whatwg.org/#concept-url-path">
-    <dfn>URL path</dfn></a> are defined in [[!URL]].
+    <dfn>URL path</dfn></a>,
+    <a href="https://url.spec.whatwg.org/#concept-url">
+    <dfn>URL record</dfn></a> and
+    <a href="https://url.spec.whatwg.org/#concept-basic-url-parser">
+    <dfn>basic URL parser</dfn></a> are defined in [[!URL]].
   </p>
   <p>
     The following terms are defined in [[!HTML5]]:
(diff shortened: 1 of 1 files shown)


alexshalamov:
@kenchris @zolkis please take a look.

index.html
```diff
@@ -2495,6 +2571,15 @@
             <code>"NotSupportedError"</code> and abort these steps.
           </li>
           <li>
+            If the <var>options</var>.url is not an empty string and it is
+            not a valid <a>URL</a>, throw <code>"SyntaxError"</code> exception
+            and abort these steps. In addition, UA may check if
```
zolkis: IIRC `https://` is required for Web NFC. If a scheme was not provided, `https://` is assumed. If it has been provided, it then MUST be `https://`. So this paragraph needs reformulation.

index.html
```diff
@@ -2495,6 +2571,15 @@
...
+            <a href="#dfn-url-scheme">scheme</a> component of a
+            <var>options</var>.url is equal to <code>"https"</code>. If not, then
+            UA may throw <code>"SyntaxError"</code> exception and abort these
```
zolkis: Here as well, `may` to be replaced with reformulated text.

zolkis:
Thank you for improving the spec. I had one minor comment that needs to be addressed.

index.html
alexshalamov: @zolkis if scheme is not provided, it would be invalid url, right? I can change MAY to MUST if you want. Do we want to restrict, e.g., web apps running locally from filesystem?

index.html
```diff
@@ -2395,39 +2403,107 @@
...
       </p>
-      <p>
-        For example, <code>'*://*.mydomain.com/*'</code> will match
-        <code>'https://services.mydomain.com/myservice1/myapp2/'</code> and
-        <code>'https://info.mydomain.com/general/'</code>.
-      </p>
     </section>
+
+    <section> <h3>URL pattern match algorithm</h3>
+        To <dfn>match Web NFC Id with URL pattern</dfn> for a given <code><a>Web NFC Id</a></code> and <code><a>URL pattern</a></code>, run these
```
alexshalamov: will add line break here

index.html
zolkis: Strictly speaking, the spec just says a secure context is required, which _implies_ `https`. It's up to the implementation whether for a missing scheme in the URL would _assume_ (autocomplete) `https://` or err out. Maybe being strict is the better option (but not requiring a scheme in the URL, and assuming `https://` would save some bytes in the URL. Anyway it would be correct to just say MUST instead of MAY.

index.html
```diff
@@ -2395,39 +2403,107 @@
       </p>
     </section>
     <section> <h3>URL patterns</h3>
+        A <dfn>URL pattern</dfn> is a valid <a>URL</a> that can be used to match
+        <a>Web NFC Id</a> of <a>Web NFC message</a>.
```
kenchris: </a>s of a <a>

index.html
```diff
@@ -2395,39 +2403,107 @@
...
-          <dfn data-lt="star path">path</dfn>        = [ <code>"/*"</code> / <code>"/"</code> &lt;VCHAR&gt; ]
-        </pre>
-      </p>
-      <p>
-        Most characters match themselves, however <code>'*'</code> (wildcard)
-        matches a sequence of visible characters. A <a>URL pattern</a> that
-        doesn't conform to the above ABNF is a non-match.
+        <a href="#dfn-url-scheme">Scheme</a>, <a href="#dfn-url-host">host</a> and
+        <a href="#dfn-url-path">path</a> components of a <a>URL pattern</a> that are
+        used by <a href="#dfn-match-web-nfc-id-with-url-pattern">URL pattern match
```
kenchris: by the

index.html
```diff
@@ -2395,39 +2403,107 @@
...
+        algorithm</a> have following matching rules:
```
kenchris: have the following

index.html
```diff
@@ -2395,39 +2403,107 @@
...
       </p>
+
       <p class="note">
-        Though <code>'*'</code> matches a sequence of characters, only URLs
-        with scheme <code>"https"</code> are accepted in the algorithms
-        taking a <a>URL pattern</a>. The <code>'*'</code> is a valid
-        character for the URL path component, therefore,
+        For example, <code>'https://mydomain.com/*'</code> will match
+        <code>'https://service.mydomain.com/myapp/'</code> and
+        <code>'https://info.mydomain.com/general/'</code>. While
```
kenchris: , while (or use another word)

index.html
```diff
@@ -2395,39 +2403,107 @@
...
+        <code>'https://app.mydomain.com/contacts'</code> will match
```
kenchris: match both

index.html
```diff
@@ -2395,39 +2403,107 @@
...
+        steps:
+        <ol>
+          <li>
+            Let <var>id</var> be a <code><a>Web NFC Id</a></code> passed to this algorithm.
+          </li>
+          <li>
+            Let <var>pattern</var> be a <a>URL pattern</a> passed to this algorithm.
+          </li>
+          <li>
+            If <var>id</var> and <var>pattern</var> are empty <code>String</code>s
```
kenchris: , at end?

index.html
```diff
@@ -2395,39 +2403,107 @@
...
+            return <code>true</code> and abort these steps.
+          </li>
+          <li>
+            Let <code>id_url</code> be a <a>URL record</a> returned by invoking
```
kenchris: invoking the

index.html
```diff
@@ -2395,39 +2403,107 @@
...
+            <a>basic URL parser</a> algorithm passing <var>id</var> as an
+            argument. If <a>basic URL parser</a> algorithm returns
```
kenchris: if the

index.html
```diff
@@ -2395,39 +2403,107 @@
...
+            <code>failure</code>, return <code>false</code> and abort these steps.
+          </li>
+          <li>
+            Let <code>pattern_url</code> be a <a>URL record</a> returned by invoking
```
kenchris: returned by invoking sounds weird

kenchris: returned when ? after?

index.html
```diff
@@ -2395,39 +2403,107 @@
...
+            <a>basic URL parser</a> algorithm passing <var>pattern</var> as an
+            argument. If <a>basic URL parser</a> algorithm returns
```
kenchris: if the

index.html
```diff
@@ -2395,39 +2403,107 @@
...
+            <code>failure</code>, return <code>false</code> and abort these steps.
```
kenchris: "failure" ?

kenchris: maybe return undefined ?

index.html
alexshalamov: Yes, sounds bad. I will rephrase it. Thanks.

index.html
alexshalamov: Return from matching algorithm? If for some reason we receive message with invalid webnfcid, matching should fail, so, I think, returning false is correct behavior. What do you think?

Basic url parser algorithm returns 'failure' when url string cannot be parsed.

index.html
alexshalamov: Ok, I think I understand what you mean. Few comments:

Scheme or 'protocol' are mandatory for valid URL, with modifications in this PR we can utilize normal URL validation algorithms.

Before this PR, we would need to check that URL pattern would have '*' or 'https' as protocol (scheme).

So, passing 'www.site.com' is invalid URL and we can return immediately, same behavior is applicable to old URL pattern.

I will change MAY to MUST, thanks.

alexshalamov:
@kenchris @zolkis  Added commit with fixes, so it is easier to review what chagned. Will squash to single commit once all review comments are addressed.

index.html
```diff
@@ -2403,14 +2401,17 @@
       </p>
     </section>
     <section> <h3>URL patterns</h3>
-        A <dfn>URL pattern</dfn> is a valid <a>URL</a> that can be used to match
-        <a>Web NFC Id</a> of <a>Web NFC message</a>.
+        A <dfn>URL pattern</dfn> is a <a>URL</a> that can be used to match
```
kenchris: to match THE

index.html
```diff
@@ -2403,14 +2401,17 @@
...
+        <a>Web NFC Id</a> of a <a>Web NFC message</a>.
+        A <dfn>valid URL pattern</dfn> is a valid <a>URL</a> whose
+        <a href="#dfn-url-scheme">scheme</a> component is equal to
+        <code>"https"</code>.
```
kenchris: what about upper case etc?

index.html
alexshalamov: Thanks, fixing.

index.html
alexshalamov: I will change to URL => 'URL record' to avoid ambiguity. All URL parts are lowercased by default.
https://url.spec.whatwg.org/#scheme-start-state.

kenchris (approved):

alexshalamov:
Aligned chromium impl with spec https://chromium-review.googlesource.com/c/512622/
@kenchris @zolkis Could you merge if you have no more comments? Thanks.

alexshalamov:
@kenchris kiitos 😄

index.html
```diff
@@ -2395,39 +2403,106 @@
...
+            Let <code>pattern_url</code> be the result of invoking the
+            <a>basic URL parser</a> algorithm with <var>pattern</var> as an input.
+            If the <a>basic URL parser</a> algorithm returns
+            <code>failure</code>, return <code>false</code> and abort these steps.
+          </li>
+          <li>
+            If <code>id_url</code>'s <a href="#dfn-url-scheme">scheme</a> <code>String</code>
+            does not match <code>pattern_url</code>'s <a href="#dfn-url-scheme">scheme</a>
+            <code>String</code>, return <code>false</code> and abort these steps.
+          </li>
```
zolkis, line 2485: I think this is good general formulation: elsewhere the spec restricts applicable schemes by requiring secure context, but perhaps we don't need to explicitly mention to check that in this algorithm.

PR #130, opened 2017-06-07
Make stricter URL pattern's host matching rule

alexshalamov:
This PR makes URL pattern host matching rule stricter, so that, pattern.host is matched against web_nfc_id.host using following rule:

`web_nfc_id.host == pattern.host || web_nfc_id.host.endsWith("." + pattern.host)`

### DIFF
--- a/index.html
+++ b/index.html
@@ -2505,15 +2505,16 @@ navigator.nfc.push({ records: [
         <code>"https"</code>.
 
       <p>
-        <a href="#dfn-url-scheme">Scheme</a>, <a href="#dfn-url-host">host</a> and
-        <a href="#dfn-url-path">path</a> components of a <a>URL pattern</a> that are
-        used by the <a href="#dfn-match-web-nfc-id-with-url-pattern">URL pattern match
-        algorithm</a> have the following matching rules:
+        A <a>URL pattern</a>'s <a href="#dfn-url-scheme">scheme</a>,
+        <a href="#dfn-url-host">host</a> and <a href="#dfn-url-path">path</a>
+        components that are used by the
+        <a href="#dfn-match-web-nfc-id-with-url-pattern">URL pattern match algorithm</a>
+        have the following matching rules:
 
         <table class="simple">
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2505,27 +2505,36 @@
...
+        <a href="#dfn-url-host">host</a> and <a href="#dfn-url-path">path</a>
+        components that are used by the
+        <a href="#dfn-match-web-nfc-id-with-url-pattern">URL pattern match algorithm</a>
+        have the following matching rules:
 
         <table class="simple">
           <tr>
             <th><strong>URL pattern component</strong></th>
-            <th><strong>Match rule</strong></th>
+            <th><strong>Match rule for Web NFC Id component</strong></th>
```
zolkis: remove "component", it's enough to say "Matching rule for Web NFC Id"

alexshalamov: done. Thanks

index.html
```diff
@@ -2579,9 +2588,16 @@
             <code>String</code>, return <code>false</code> and abort these steps.
           </li>
           <li>
+            Let <code>subdomain_pattern</code> be the result of prepending <code>"."</code>
```
kenchris, line 2591: I think that in spec sense you don't need to use _ - you can do normal language <code>sub domain pattern</code>

We do that in manifest etc

kenchris:
lgtm

kenchris (approved):

zolkis:
LGTM

zolkis:
Alex, you can merge :).

index.html
alexshalamov: Would it compile :smile:? Do you want me to remove '_' separators from the algorithm?

zolkis: IMHO it's fine to leave the '_', and you could also use dashes or spaces between, once all of it is wrapped in `<code></code>`

alexshalamov: @zolkis ok, I will leave it then. Thanks

Issue #131, opened 2017-06-17
Clarify payment scenarios use case


cyberphone:
If you read this document
http://w3c.github.io/web-nfc/index.html#payment-scenarios
you can get the impression that a future revision could support payments.

This is incorrect, NFC based payments are currently based on low-level card protocols that were _not designed for the Web_.  In addition, Mobile "Wallets" are exclusively based on native applications.  It will stay like that since there is no "AndroidKeystore" for the Web.  Efforts for creating such features have been going on since 2011 but have to date always failed, most recently in https://www.w3.org/community/hb-secure-services/

The only way Web NFC could ever support payments would be through entirely new schemes but no such activity seems to be in the workings.

zolkis:
That use case is not for payment itself, but recording the fact of payment with a website in order to have secondary actions. However, we do need to revisit that section.

cyberphone:
Well, if you "read between the lines", this use case as well as most of the other listed use cases are really about providing Web based alternatives to native "Apps".

Although I absolutely love the Web, there are plenty of applications out there that are better served by "Apps".  For security related applications like payments you need signed and vetted applications which doesn't have a Web counterpart.  However, a viable (and extremely popular), way forward is _combining_ the Web and App worlds since these worlds offer unique capabilities of their own.

The recent W3C Web Payment API is the first Web standard of this kind, offering an application specific "bridge" between the two worlds, currently running on Android (for Android Pay and similar).  This API will presumably be extended to cope with whatever additional features are needed including loyalty cards.

Outside of the W3C standards' scope, Mozilla and Microsoft are now shipping emulations of Chrome's "Native Messaging" which is a variant on the same theme.

It had been pretty cool making a Web NFC bridge but that would be an entirely different project since not even the required hardware is available: https://github.com/w3c/web-nfc/issues/128#issuecomment-308838885.

zolkis:
> this use case as well as most of the other listed use cases are really about providing Web based alternatives to native "Apps".

Actually quite the contrary. This group intended to work on Web-specific use cases of NFC (as a transport/interaction mechanism) rather than providing a generic NFC API for making Web based alternatives to native apps. This is also the reason for departure from the lower level APIs of the former NFC WG. And again, the payment scenarios did no mean to _do_ payments with Web NFC, but rather, do *something else* when an NFC payment is done.

cyberphone:
>> this use case as well as most of the other listed use cases are really about providing Web based alternatives to native "Apps".

[...]

PR #132, opened 2017-06-20
Address #131. Clarify payment use case.

zolkis:

### DIFF
--- a/index.html
+++ b/index.html
@@ -918,7 +918,11 @@ navigator.nfc.push({ records: [
     </section>
     <section> <h3>Payment scenarios</h3>
       <p>
-        The user buys goods in a store, and payments options include NFC.
+        Payment scenarios with Web NFC generally do not refer to supporting
+        the payment process itself, but associating the payment status with
+        a web page in order to have secondary actions. For instance,
+        the user buys goods in a store, and payments options include contactless
+        payment using NFC technology.
         In general, touching the device to the point of sales terminal receiver
         area will result in a transaction between the secure element from the
         device and the point of sales terminal. With the Web NFC API, if the
@@ -926,6 +930,8 @@ navigator.nfc.push({ records: [
         with that site regarding the payment, e.g. the user could get points and
         discounts, or get delivered application or service specific data (e.g.
         tickets, keys, etc) to the device.
+      </p>
+      <p class="note">
         This use case is not supported in this version of the specification.
       </p>
     </section>


index.html
```diff
@@ -918,15 +918,21 @@
     </section>
     <section> <h3>Payment scenarios</h3>
       <p>
-        The user buys goods in a store, and payments options include NFC.
+        Payment scenarios with Web NFC generally do not refer to supporting
+        the payment process itself, but recording the fact of payment with a web
```
kenchris: recording the fact of payment - is a bit hard to understand

zolkis: suggest

index.html
```diff
@@ -918,15 +918,21 @@
...
         device and the point of sales terminal. With the Web NFC API, if the
         user navigates to a web site before paying, there may be interaction
         with that site regarding the payment, e.g. the user could get points and
         discounts, or get delivered application or service specific data (e.g.
         tickets, keys, etc) to the device.
-        This use case is not supported in this version of the specification.
+      </p>
+      <p class="note">
+        This use case is not supported in this version of the specification (see
+        also <a href="https://github.com/w3c/web-nfc/issues/131">issue 136</a>).
```
kenchris: remove also

index.html
zolkis: the example should clarify

index.html

zolkis: although issue 131 is not a full coverage of not supporting payment use cases in the spec

index.html
kenchris: I am trying to wrap my head around what you wanted to say :)

"but the fact that a payment on the web may happen by the use of NFC technology, which adds the ability to perform secondary NFC actions at the same time"

zolkis: > "but the fact that a payment on the web may happen by the use of NFC technology, which adds the ability to perform secondary NFC actions at the same time"

now it's me who doesn't understand your sentence :D

@anssiko chime in?

anssiko: How about:

NFC can have multiple records that are read when touched, and Web NFC actions can be taken at the same time with the payment that may be implemented using native NFC technology.

anssiko: @zolkis please update the PR if this works for you, @kenchris approved the proposed text.

zolkis: Amended again, please check if it's better than now. The use case text is written in the context of host card emulation with secure element.

index.html
```diff
@@ -918,15 +918,21 @@
...
         In general, touching the device to the point of sales terminal receiver
         area will result in a transaction between the secure element from the
         device and the point of sales terminal. With the Web NFC API, if the
         user navigates to a web site before paying, there may be interaction
         with that site regarding the payment, e.g. the user could get points and
         discounts, or get delivered application or service specific data (e.g.
         tickets, keys, etc) to the device.
-        This use case is not supported in this version of the specification.
+      </p>
+      <p class="note">
```
anssiko: I suggest you make this note simply:

>This use case is not supported in this version of the specification.

zolkis: ok

index.html
```diff
@@ -918,14 +918,19 @@
     </section>
     <section> <h3>Payment scenarios</h3>
       <p>
-        The user buys goods in a store, and payments options include NFC.
+        Payment scenarios with Web NFC generally do not refer to supporting
+        the NFC payment process itself, but associating the payment status with
```
alexshalamov: Maybe rephrase it differently? NFC can be used as a 'companion' technology for payments, e.g., for EMV, thus, NFC is not a payment process.

> Payment scenarios with Web NFC generally do not refer to supporting the NFC payment process itself

Payment scenarios with Web NFC generally do not refer to supporting the payment process itself

index.html
```diff
@@ -918,14 +918,19 @@
...
+        a web page in order to have secondary actions. For instance,
+        the user buys goods in a store, and payments options include NFC.
```
alexshalamov: include contactless payment using NFC technology?

alexshalamov (approved):
few nits

index.html
zolkis: Thanks! Amended for the new changes.

alexshalamov (approved):

anssiko:
Thanks for the contributions, @zolkis you can merge this now.

We should keep on maintaining the use cases document along the specification. The use cases doc is to be considered a "living document". It is normal that the initial assumptions made in the beginning may change as we collectively gain more insight.

PR #143, opened 2018-04-02
Use `[SecureContext]` instead of manually throwing an exception

romandev:


kenchris:
@alexshalamov are you fine with this? I assume this is the same you do in generic sensors?

index.html
```diff
@@ -1462,6 +1463,9 @@ <h2>The <dfn>NFCRecordType</dfn> enum</h2>
     cancelable Promises</a> used with <code>push()</code> and
     <code>watch()</code>, respectively.
   </p>
+  <p class="note">
```
alexshalamov: I'm not sure whether this extra information adds something, localhost considered to be secure, that is sufficient for development needs, thus UA don't need to handle special case.

index.html
```diff
@@ -1765,18 +1769,6 @@ <h2>The <dfn>NFCWatchMode</dfn> enum</h2>
             Return a new <a><code>Promise</code></a> <var>promise</var>, and
             then continue running this algorithm <a>in parallel</a>.
           </li>
-          <li>
-            If any exception occurs while running these steps, reject
```
alexshalamov: This block is not related to secure context, we can keep it as it is.

index.html
```diff
@@ -2656,18 +2640,6 @@ <h2>The <dfn>NFCWatchMode</dfn> enum</h2>
             Return <var>promise</var> and continue the following steps
             <a>in parallel</a>.
           </li>
-          <li>
-            If any exception occurs while running these steps, reject
```
alexshalamov: Same as above, let's remove custom handling of secure context and remove only following blocks:

```
-          <li>
-            If the <a>incumbent settings object</a> is not a
-            <a>secure context</a>, reject <var>promise</var> with
-            <code>"<a>SecurityError</a>"</code> and abort these steps.
-            <div class="note">
-              Browsers may ignore this rule for development purposes only.
-            </div>
-          </li>
```

alexshalamov (changes requested):

romandev:
@alexshalamov Thank you for review. I addressed your comments.

jyasskin (approved):

alexshalamov (approved):

alexshalamov:
@romandev Thanks!

Issue #149, opened 2018-08-08
NFCWatchOptions's recordType member should probably not be nullable


foolip:
See discussion at https://github.com/heycam/webidl/issues/213#issuecomment-411399248

The member doesn't need to have a default value, but the prose should describe what happens when the member is missing. The current wording 'The default value "" means that no matching happens.' appears to be copy-pasted as "" isn't one of the enum values.

PR #150, opened 2018-08-09
Improve description of NFCWatchOptions.recordType attribute

alexshalamov:
This PR clarifies how NFC watch algorithm treats absence of NFCWatchOptions.recordType dictionary member.

Fixes: #149

### DIFF
--- a/index.html
+++ b/index.html
@@ -252,9 +252,11 @@
     <a href="http://heycam.github.io/webidl/#idl-ArrayBuffer">
       <code><dfn>ArrayBuffer</dfn></code></a>,
     <a href="http://heycam.github.io/webidl/#common-BufferSource">
-      <code><dfn>BufferSource</dfn></code></a> and
+      <code><dfn>BufferSource</dfn></code></a>,
     <a href="http://heycam.github.io/webidl/#idl-any">
-      <code><dfn>any</dfn></code></a>
+    <code><dfn>any</dfn></code></a> and
+      <a href="https://heycam.github.io/webidl/#dfn-present">
+    <code><dfn>not present</dfn></code></a>
     are defined in [[!WEBIDL]].
   </p>
   <p>
@@ -1662,7 +1664,7 @@ navigator.nfc.push({ records: [
       <pre class="idl">
         dictionary NFCWatchOptions {
           USVString url = "";
-          NFCRecordType? recordType;
+          NFCRecordType recordType;
           USVString mediaType = "";
           NFCWatchMode mode = "web-nfc-only";
         };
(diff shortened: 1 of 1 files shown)


alexshalamov:
@foolip could you take a look?

index.html
```diff
@@ -1675,11 +1675,13 @@ <h2>The <dfn>NFCPushTarget</dfn> enum</h2>
...
       <p>
         The <dfn>NFCWatchOptions.recordType</dfn> property
-        denotes the string value which is used for matching the
+        denotes the enum value which is used for matching the
         <code>
         <a href="#idl-def-nfcrecordtype">type</a></code> property of each
         <code><a>NFCRecord</a></code> object in a <a>Web NFC message</a>.
-        The default value <code>""</code> means that no matching happens.
+        If the dictionary member is absent or its value is <code>null</code>,
+        then it is not used by the
```
kenchris: or "ignored" ?

foolip:
@alexshalamov can you make the dictionary member not nullable? Then, you need only deal with the "is absent" case. (I filed https://github.com/w3c/web-nfc/issues/149 about nullability primarily.)

kenchris:
I don't see the point in having it nullable either. You can make it absent or add undefined instead.

Actually if you mark something as undefined, it will still be iterable.

```
a = { hest: null, nonhest: undefined }

for (var entry in a) { console.log(entry) }
VM214:1 hest
VM214:1 nonhest
```

alexshalamov:
@foolip fixed comments. I'm not sure why it was nullable. This change introduces small break, e.g., `null` will not be one of allowed values for the `NFCWatchOptions.recordType`.

index.html
```diff
@@ -1675,11 +1677,12 @@ <h2>The <dfn>NFCPushTarget</dfn> enum</h2>
       </p>
       <p>
         The <dfn>NFCWatchOptions.recordType</dfn> property
-        denotes the string value which is used for matching the
+        denotes the enum value which is used for matching the
         <code>
         <a href="#idl-def-nfcrecordtype">type</a></code> property of each
         <code><a>NFCRecord</a></code> object in a <a>Web NFC message</a>.
-        The default value <code>""</code> means that no matching happens.
+        If the dictionary member is <a>not present</a>, then it will be ignored by the
```
foolip, line 1684: Are some changes in the "NFC watch algorithm" algorithm also needed?

alexshalamov: @foolip Improved dispatch algorithm for handling NFCWatchOptions.recordType, as it was checking for empty string, instead of checking member presence.

foolip (approved):
Looks great, thanks!

foolip:
Is this ready to be merged?

alexshalamov:
@foolip rebased and merged. Thanks for feedback.

PR #153, opened 2018-09-29
Fix reference URLs

domenic:
Lots of references were mismashes of different outdated forks of the original spec, or things that have moved.


zolkis:
Wow! Lot of work. Thanks!

kenchris:
Thank you very much!

PR #162, opened 2018-11-27
Use AbortSignal instead of cancelWatch

kenchris:


kenchris:
Can you have a look @zolkis - I will change the push as well afterwards

zolkis:
Looks nice. Thanks!

PR #164, opened 2018-11-29
Fix #133: add Type 5 mention in Introduction

### LINKED ISSUE #133: Type5 support in web NFC APIs
The WEB NFC APIs specs do not explicitly mention the type5.
The Introduction talks about type1 to type4.
The type5 is now officially part of the NFC forum specs.
Given the current implementation in chromium the Web NFC APIs work fine on Android devices with Type5 tags. Indeed the implementation is tech agnostic (using NDEF APIs from Android).
Moreover, beginning of june Apple mentioned that iOs 11 will open NFC APis (In a first step to read NDEF messages). Actually iOs 11 in beta version do support the type5, 
I do not know if Safari 'll support the WebNFC Apis,  but type 5 is definitively an important actor of the NFC ecosystem (providing, for instance better performances in read range).
I think that it should now be added to these specs. It would be a bug to ignore the type5

zolkis:

### DIFF
--- a/index.html
+++ b/index.html
@@ -2647,7 +2647,7 @@ navigator.nfc.push({ records: [
             Let <var>options</var> be the second callback.
           </li>
           <li>
-            Let <var>signal</var> be the options’ dictionary member of the same
+            Let <var>signal</var> be the <var>options</var>’ dictionary member of the same
             name if present, or null otherwise.
           </li>
           <li>
--- a/index.html
+++ b/index.html
@@ -590,6 +590,15 @@
         service. Supports three different communication speeds 106 or 212 or
         424 Kbits/s.
       </li>
+      <li>
+        <b>NFC Forum Type 5</b>: This tag is based on the NFC-V RF
+        technology which is based on ISO/IEC 15693 and allows reading and
+        writing an NDEF message on a ISO/IEC 15693 RF tag that is accessible by
+        long range RFID readers as well. The NFC communication is limited to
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -590,6 +590,15 @@
         service. Supports three different communication speeds 106 or 212 or
         424 Kbits/s.
       </li>
+      <li>
+        <b>NFC Forum Type 5</b>: This tag is based on the NFC-V RF
+        technology which is based on ISO/IEC 15693 and allows reading and
+        writing an NDEF message in a ISO/IEC 15693 RF tag that is accessible by
```
kenchris: on a ?

index.html
```diff
@@ -2638,7 +2647,7 @@ <h2>The <dfn>NFCWatchMode</dfn> enum</h2>
             Let <var>options</var> be the second callback.
           </li>
           <li>
-            Let <var>signal</var> be the options’ dictionary member of the same 
+            Let <var>signal</var> be the options’ dictionary member of the same
```
Honry:
```suggestion
            Let <var>signal</var> be the <var>options</var>’ dictionary member of the same
```

index.html
zolkis: resolved

kenchris:
Why is there a commit from Honry (unrelated) here?

kenchris:
OK, lets just merge this, but please don't mix commits in the future :-)

PR #165, opened 2018-11-30
Replace cancelPush with AbortSignal

### LINKED ISSUE #147: Use AbortController/*Signal instead of cancelWatch(optional  long id);
https://developer.mozilla.org/en-US/docs/Web/API/AbortController

That is a much more modern way of doing this, but one that didn't exist when `cancelWatch` was added to the spec

kenchris:
Closes #147

### DIFF
--- a/index.html
+++ b/index.html
@@ -1438,7 +1438,7 @@ navigator.nfc.push({ records: [
   The <dfn>NFC</dfn> object provides a way for the <a>browsing context</a> to
   use NFC functionality.
   It allows for pushing <a>Web NFC message</a>s to <a>NFC tag</a>s
-  or <a>NFC peer</a>s within range, and to set up and cancel <a>NFC watch</a>es
+  or <a>NFC peer</a>s within range, and to set up and abort <a>NFC watch</a>es
   to handle incoming <a>Web NFC message</a>s either from an <a>NFC tag</a> or an
   <a>NFC peer</a>.
   <pre class="idl">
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1561,12 +1560,20 @@ <h2>The <dfn>NFCRecordType</dfn> enum</h2>
...
-      steps with <code>"peer"</code> as parameter.
+      For each pending <a>NFC.push</a>:
+      <ol>
+        <li>
+          Stop the instance's <var>timer</var> if it is active.
+        </li>
+        <li>
+          If the instance has already initiated NFC data transfer,
+          continue.
+        </li>
```
zolkis, line 1571: If we let the transfer continue, I would not stop its timer. 
If a transfer has started and we let it continue, it should finish regularly and perhaps we could fail the abort, then the client code could try later.

index.html
```diff
@@ -1755,200 +1767,228 @@ <h2>The <dfn>NFCWatchMode</dfn> enum</h2>
...
+              <ol>
+                <li>
+                  Stop the instance's <var>timer</var> if it is active.
+                </li>
+                <li>
+                  If the instance has already initiated NFC data transfer,
+                  reject <var>p</var> with
+                  <code>"<a>NoModificationAllowedError</a>"</code> and abort these
+                  steps.
+                </li>
```
zolkis, line 1809: Here as well, I would put the stop timer step after this one.

zolkis (commented):
Nice, thank you. We should double check timer management during aborts.

index.html
kenchris: Continue means move to the next item in the foreach...

zolkis: yes, but a transfer is left ongoing and we have no timer left; perhaps it's fine, we just lose the explicit possibility to automatically abort it (using this algorithm), but we don't have much effect over the underlying NFC transfer any more. So the steps are probably "best effort".

kenchris: yes that is how I see it...

zolkis (approved):
We need to think over the timer management, but the PR is ok in the sense that it's describing a "best effort" abort.

PR #173, opened 2018-12-04
Modernize examples

kenchris:


anssiko (approved):

index.html
```diff
@@ -762,11 +763,8 @@
     }]
   };
 
-  navigator.nfc.push(message).then(() => {
-    console.log('Progress stored!');
-  }).catch((error) => {
-    console.log('Failure, please try again.');
-  });
+  await navigator.nfc.push(message);
+  console.log("Pushed message");
```
zolkis, line 767: error path?

kenchris: shown in multiple other examples, so a bit repetitive, which is why I left it out here

zolkis: ok, fine, it's just an example

Issue #152, opened 2018-09-27
Use addEventListener instead of watch


kenchris:
I don't recall why we invented the watch, but it feels a bit like its own thing, instead of being aligned with the platform. We could go with a `startNotifications()` approach like WebBluetooth.

```js
navigator.nfc.addEventListener('exchange', message => {
  for (let record of message.records) {
    let article =/[aeio]/.test(record.data.title) ? "an" : "a";
    console.log(`${record.data.name} is ${article} ${record.data.title}`);
  }
},{
  url: document.baseURI,
  recordType: "json"
});

navigator.nfc.startNotifications();
```

This means that all of our values in the options dict needs to qualify as a separate handler (ie if you want to remove you need the same options dict and same function

zolkis:
We could add a filter to `watch()`, i.e. options.
It would be nice if `addEventListener()` supported options.
Next possibility is Observables (not that it would be aligned with the platform, either).
Or just do away with options.

kenchris:
`addEventListener()` does support options today :-)

https://dom.spec.whatwg.org/#dictdef-addeventlisteneroptions

zolkis:
But not _those_ NFC options. Is it better to extend a standard dictionary with NFC-specific properties (making it _apparently_ aligned), or have a separate call with its own, admittedly different semantics?

kenchris:
I think it is better extending. Creating something that looks similar but has edge cases that works differently isn't really good.

Other option would be (feel free to bikeshed names)

```js
const filter = navigator.nfc.createReadFilter({
  url: document.baseURI,
  recordType: "json"
});

const onMessage = message => {
  for (let record of message.records) {
    let article =/[aeio]/.test(record.data.title) ? "an" : "a";
    console.log(`${record.data.name} is ${article} ${record.data.title}`);
  }
};

filter.addEventListener('exchange', onMessage, {capture: false});
filter.startNotifications();
```

zolkis:
> I think it is better extending. 

Maybe so, but what is the general consensus or mechanism on extending standard dictionaries in the platform? Do we need to register those somewhere? (imagine multiple specs adding properties which may clash)

> Creating something that looks similar but has edge cases that works differently isn't really good.

Extending a standard dictionary looks exactly like this. So it's a counter-argument.

> `const filter = navigator.nfc.createReadFilter({`

So we'd need to specify a new NFCFilter interface that contains the `startNotifications()` method? 
Isn't that more complex than just using a `watch()` method?

[...]

PR #177, opened 2018-12-05
Move push to a new NFCWriter object

kenchris:
The easy part of issue #152

### DIFF
--- a/index.html
+++ b/index.html
@@ -640,7 +640,8 @@
       peers.
     </p>
     <pre class="highlight">
-navigator.nfc.push(
+const writer = new NFCWriter();
+writer.push(
   '{ "prop1": "value1", "prop2": "value2" }'
 ).then(() => {
   console.log("Message pushed.");
@@ -657,7 +658,8 @@ navigator.nfc.push(
       and thus, no data is pushed when the user taps a tag.
     </p>
     <pre class="highlight">
-navigator.nfc.push(
+const writer = new NFCWriter();
+writer.push(
   "Text meant for peers only", { target: "peer" }
 ).then(() => {
   console.log("Message pushed.");
@@ -672,7 +674,8 @@ navigator.nfc.push(
       In order to push an NDEF record of URL type, simply use NFCMessage.
     </p>
     <pre class="highlight">
-navigator.nfc.push({
+const writer = new NFCWriter();
+writer.push({
   records: [{ recordType: "url", data: "https://w3c.github.io/web-nfc/" }]
 }).then(() => {
   console.log("Message pushed.");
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1182,7 +1190,7 @@
     <p>
       The <dfn>NFCMessage.url</dfn>
       property represents the <a>Web NFC Id</a> of a received
-      <a>Web NFC message</a>. When used in the <code><a>NFC.push</a>()</code>
+      <a>Web NFC message</a>. When used in the <code><a>NFCWriter.push</a>()</code>
```
littledan, line 1193: Nit: Consider writing something like NFCWriter.prototype.push (fine to do this in another PR)

kenchris: Sure. I hope that works with respec :-)

index.html
```diff
@@ -1564,7 +1580,7 @@ <h2>The <dfn>NFCRecordType</dfn> enum</h2>
       Set <var>nfc</var>.[[\suspended]] to <code>true</code>.
     </li>
     <li>
-      For each pending <a>NFC.push</a>:
+      For each pending <a>NFCWriter.push</a>:
```
littledan, line 1583: I guess this is talking about making the NFCWriter object disappear, and we're doing this for each push pending on the NFCWriter object, right? (Since presumably this will be called for all the NFCWriter objects.)

We should probably have a little layering change in HTML's [unloading document cleanup steps](https://html.spec.whatwg.org/#unloading-document-cleanup-steps) to let all platform objects register "make disappear" steps. (And, going crazy with it, maybe we should have a JS API for this as well?)

(Doesn't need to be clarified in this PR)

index.html
```diff
@@ -1448,17 +1456,25 @@ <h2>The <dfn>NFCRecordType</dfn> enum</h2>
   <pre class="idl">
     typedef (DOMString or ArrayBuffer or NFCMessage) NFCMessageSource;
 
+    [Constructor(), SecureContext, Exposed=Window]
+    interface NFCWriter {
+      Promise&lt;void&gt; push(NFCMessageSource message, optional NFCPushOptions options);
+    };
+
     [SecureContext]
```
littledan, line 1464: The NFC interface should either have `[Exposed=Window]` or `[NoInterfaceObject]` (I guess the latter?). Anyway, the extended attributes for NFCWriter LGTM, and I guess NFC will go away when you add NFCReader.

index.html
```diff
@@ -1448,17 +1456,25 @@ <h2>The <dfn>NFCRecordType</dfn> enum</h2>
...
     interface NFC {
-      Promise&lt;void&gt; push(NFCMessageSource message, optional NFCPushOptions options);
       Promise&lt;void&gt; watch(MessageCallback callback, optional NFCWatchOptions options);
     };
 
     callback MessageCallback = void (NFCMessage message);
   </pre>
+  <p>
+    The <dfn>NFCWriter</dfn> is an object used for writing data to NFC devices
```
littledan, line 1472: Nit: Instances of NFCWriter are used to write data to NFC devices (this is not a singleton class)

littledan (approved):
This looks fine to land; all comments I have are separate things to look into changing.

index.html
littledan: Added https://github.com/w3c/web-nfc/pull/178/commits/459a51a551b309278171f014f32ed9c2b39f4466 to clarify these things (I had previously missed the connection below, and seems like we don't need a layering change).

PR #178, opened 2018-12-05
Editorial: Move state from the NFC interface to the settings object

### LINKED ISSUE #156: Do not use "script execution environment"
"script execution environment" is not a well-defined term. The spec links to https://html.spec.whatwg.org/#script-execution-environment which does not exist. It then seems to try to define it itself, but in a way that doesn't actually say what it is.

I'm not sure what the point of this term is, but it seems to be only used in the non-normative security policies section, plus once redundantly elsehwere ("Each Window object connected to the script execution environment" can just be "Each Window object") so hopefully it can be removed.

littledan:
This refactoring should make it easier to change the interface to
NFCReader, to address #152

Along the way, closes #156.

### DIFF
--- a/index.html
+++ b/index.html
@@ -216,16 +216,16 @@
 
     <a href="https://html.spec.whatwg.org/multipage/#the-iframe-element">
     <dfn>iframe</dfn></a>.
+
+    <a href="https://html.spec.whatwg.org/multipage/#settings-object">
+    <dfn>settings object</dfn></a>.
   </p>
   <p>
     A <a>browsing context</a> refers to the environment in which
     <a>Document</a> objects are presented to the user. A given
     <a>browsing context</a> has a single <code><a>WindowProxy</a></code> object,
     but it can have many <code>Document</code> objects, with their associated
-    <code><a>Window</a></code> objects. The <dfn>script execution environment</dfn>
-    associated with the <i>browsing context</i> identifies the entity which
-    invokes this API, which can be a <i>web app</i>, a <i>web page</i>, or an
-    <a>iframe</a>.
+    <code><a>Window</a></code> objects.
   </p>
   <p>
     The term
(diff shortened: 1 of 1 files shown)


littledan:
@domenic is this interacting with HTML correctly?

index.html
```diff
@@ -1478,21 +1465,26 @@ <h2>The <dfn>NFCRecordType</dfn> enum</h2>
...
         A <a>set</a> of <a>NFC watch</a>es initially set to the empty <a>set</a>.
       </td>
      </tr>
     </tbody>
   </table>
+  <p class="note">
+    Internal slots are used only as a notation in this specification, and
+    implementations do not necessarily have to map them to explicit internal
+    properties.
+  </p>
```
kenchris, line 1503: Great note

kenchris (approved):

index.html
littledan: (It was there before; I just moved it.)

littledan: I was wondering, should we add a note that there's only one NFC device per UA?

kenchris: There could potentially be multiple, it would just push a write to both in parallel, as well as set up the same watch for both

index.html
```diff
@@ -1117,12 +1104,11 @@
...
       </li>
       <li>
-        When pushing <a>Web NFC content</a>, the
-        <a>serialized origin</a> and the <a>URL path</a> of
-        the <a>incumbent settings object</a> of the
-        <a>script execution environment</a> requesting the operation must be
-        recorded in each sent <a>NDEF message</a>'s <a>Web NFC record</a>.
-        For details see the <a>Writing or pushing content</a> section.
+        When pushing <a>Web NFC content</a>, the <a>serialized origin</a> and
+        the <a>URL path</a> of the <a>incumbent settings object</a> when
```
domenic, line 1108: Separate issue, but incumbent settings object is strongly deprecated...

index.html
```diff
@@ -216,16 +216,16 @@
 
     <a href="https://html.spec.whatwg.org/multipage/#the-iframe-element">
     <dfn>iframe</dfn></a>.
+
+    <a href="https://html.spec.whatwg.org/multipage/#settings-object">
+    <dfn>settings object</dfn></a>.
```
domenic, line 221: "Settings object" is an internal slot of scripts. "Environment settings object" is the type.

index.html
```diff
@@ -1480,10 +1466,11 @@ <h2>The <dfn>NFCRecordType</dfn> enum</h2>
     The <dfn>MessageCallback</dfn> is a callback function that must be
     provided to <code><a href="#dom-nfc-watch">watch()</a></code> metod.
   </p>
-  <section><h3>Internal slots of NFC</h3>
+  <section><h3>NFC state associated with the settings object</h3>
   <p>
-    The <a>NFC</a> object is created with the <a>internal slots</a>
-    described in the following table:
+    The <a>settings object</a> of a <a>browsing context</a> which supports
```
domenic, line 1471: Browsing contexts don't have (environment) settings objects.

Browsing contexts have active documents, which themselves have relevant settings objects.

littledan:
Thanks for the review and refresher on the basics, @domenic ! https://github.com/w3c/web-nfc/pull/182 tries to fix up the errors you noted.

PR #179, opened 2018-12-06
Introduce NFCReader object with event

kenchris:
@littledan can you have a look? Please look at wording as well, I would want it as simple as possible, but respec requires me to create dfns for everything

### DIFF
--- a/index.html
+++ b/index.html
@@ -683,24 +683,19 @@ writer.push({
     </p>
     <pre class="highlight">
 const writer = new NFCWriter();
+const reader = new NFCReader();
 
-navigator.nfc.watch((message) => {
+reader.onreading = message => {
   if (message.records[0].recordType == 'empty') {
     writer.push({
       url: "/custom/path",
       records: [{ recordType: "text", data: 'Hello World' }]
     });
-  } else {
-    console.log(`Read message written by ${message.url}`);
-    processMessage(message);
+    return;
   }
-}).then(() => {
-  console.log("Added a watch.");
-}).catch((error) => {
-  console.log(`Adding watch failed: ${error.name}`);
-});
 
-function processMessage(message) {
+  console.log(`Read message written by ${message.url}`);
+
   for (let record of message.records) {
     switch (record.recordType) {
       case "text":
@@ -725,7 +720,9 @@ function processMessage(message) {
         break;
     }
   }
-}
+};
+
+reader.watch();
     </pre>
   </aside>
 
@@ -743,13 +740,13 @@ function processMessage(message) {
       whichever one is tapped first.
     </p>
     <pre class="highlight">
-navigator.nfc.watch(reader, { url: "https://mygame.com/mypath/mygame" });
-
-async function reader(message) {
+const reader = new NFCReader();
+reader.watch({url: "https://mygame.com/mypath/mygame"});
+reader.onreading = async message => {
   console.log(`Source:     ${ message.url }`);
   console.log(`Game state: ${ JSON.stringify(message.records) }`);
 
-  const message = {
+  const newMessage = {
     url: "/mypath/mygame/update",
     records: [{
       recordType: "json",
@@ -758,9 +755,9 @@ async function reader(message) {
     }]
   };
   const writer = new NFCWriter();
-  await writer.push(message);
+  await writer.push(newMessage);
   console.log("Pushed message");
-}
+};
     </pre>
   </aside>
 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1447,24 +1431,44 @@ <h2>The <dfn>NFCRecordType</dfn> enum</h2>
       Promise&lt;void&gt; push(NFCMessageSource message, optional NFCPushOptions options);
     };
 
-    [SecureContext]
-    interface NFC {
-      Promise&lt;void&gt; watch(MessageCallback callback, optional NFCWatchOptions options);
+    [Constructor(), SecureContext, Exposed=Window]
+    interface NFCReader : EventTarget {
+      Promise&lt;void&gt; watch(optional NFCWatchOptions options);
```
littledan, line 1435: Bikeshedding: Should we have a watch method, or should these be arguments for the constructor?

index.html
```diff
@@ -770,7 +772,8 @@
       Storing and receiving JSON data is easy with serialization and deserialization.
     </p>
     <pre class="highlight">
-navigator.nfc.watch((message) => {
+const reader = new NFCReader();
+reader.watch(message => {
```
littledan: Is this supposed to be `reader.onreading = message => ...` ?

index.html
```diff
@@ -743,7 +744,8 @@
       whichever one is tapped first.
     </p>
     <pre class="highlight">
-navigator.nfc.watch(reader, { url: "https://mygame.com/mypath/mygame" });
+const reader = new NFCReader();
+reader.watch(reader, { url: "https://mygame.com/mypath/mygame" });
```
littledan: Delete the first argument?

index.html
```diff
@@ -683,8 +683,9 @@
     </p>
     <pre class="highlight">
 const writer = new NFCWriter();
+const reader = new NFCReader();
 
-navigator.nfc.watch((message) => {
+reader.watch(message => {
```
littledan: Ditto from https://github.com/w3c/web-nfc/pull/179/files#r239426031

index.html
```diff
@@ -1447,24 +1431,44 @@ <h2>The <dfn>NFCRecordType</dfn> enum</h2>
       Promise&lt;void&gt; push(NFCMessageSource message, optional NFCPushOptions options);
     };
 
-    [SecureContext]
-    interface NFC {
-      Promise&lt;void&gt; watch(MessageCallback callback, optional NFCWatchOptions options);
+    [Constructor(), SecureContext, Exposed=Window]
+    interface NFCReader : EventTarget {
+      Promise&lt;void&gt; watch(optional NFCWatchOptions options);
+      attribute EventHandler onreading;
```
littledan, line 1436: Bikeshedding: I'm not sure about the "reading" name; I will have to think about it more. If others like it, I'm fine with it.

index.html
littledan: To answer my own stupid question: It's because watch returns a promise when the initialization is complete.

To continue bikeshedding: Could this still take the options as an argument, and have a `readonly attribute Promise<void> loaded` to represent the return value of `watch`, and let people detect failure? (We should maybe have examples of the failure case.)

index.html
```diff
@@ -807,7 +810,8 @@
       property to <code>false</code> and set up a watch before the push.
     </p>
     <pre class="highlight">
-navigator.nfc.watch((message) => {
+const reader = new NFCReader();
+reader.watch(message => {
```
littledan: https://github.com/w3c/web-nfc/pull/179/files#r239426031

index.html
littledan: I see, "reading" is used all over the generic sensor API--makes sense here.

index.html
littledan: I guess if you want to match the generic sensor API, there would just be `onerror` and `onactivate` events for this purpose. Seems like another decent option.

kenchris: Yes, these are follow up things, I also need the NFCReader ctor to take the options dict

index.html
kenchris: yes! Good catch

index.html
kenchris: Yup that is why I choose it

index.html
kenchris: OK will fix the samples!

index.html
kenchris: I don't think we need onactivate as we don't have start/stop (@reillyeon wanted to avoid that)

index.html
kenchris: done

index.html
kenchris: done

littledan (approved):

PR #180, opened 2018-12-07
Move watch options to NFCReader ctor

kenchris:
This is an incremental change, as `watch` will be removed in future PR

### DIFF
--- a/index.html
+++ b/index.html
@@ -730,7 +730,7 @@ reader.watch();
        class="example">
     <p>
       Filtering of relevant data sources can be done by the use of
-      the <a>NFCWatchOptions</a>. Below we accept URL with <code>
+      the <a>NFCReaderOptions</a>. Below we accept URL with <code>
       "/mypath/mygame/"</code> in its path from <code>"mygame.com"</code>
       domain and its subdomains. When we read the data, we immediately update
       the game progress by issueing a push with a custom NDEF data layout.
@@ -1430,10 +1430,11 @@ writer.push({ records: [
       Promise&lt;void&gt; push(NFCMessageSource message, optional NFCPushOptions options);
     };
 
-    [Constructor(), SecureContext, Exposed=Window]
+    [Constructor(optional NFCReaderOptions options), SecureContext, Exposed=Window]
     interface NFCReader : EventTarget {
-      Promise&lt;void&gt; watch(optional NFCWatchOptions options);
       attribute EventHandler onreading;
+
+      Promise&lt;void&gt; watch();
     };
 
     [Constructor(DOMString type, NFCReadingEventInit nfcReadingEventInitDict), SecureContext, Exposed=Window]
@@ -1678,28 +1679,28 @@ writer.push({ records: [
     </dl>
   </section> <!-- NFCPushTarget enum -->
 
-  <section> <h3>The <dfn>NFCWatchOptions</dfn> dictionary</h3>
+  <section> <h3>The <dfn>NFCReaderOptions</dfn> dictionary</h3>
       <p>
         To describe which messages an application is interested in, the
-        <a>NFCWatchOptions</a> dictionary is used:
+        <a>NFCReaderOptions</a> dictionary is used:
       </p>
       <pre class="idl">
-        dictionary NFCWatchOptions {
+        dictionary NFCReaderOptions {
           USVString url = "";
           NFCRecordType recordType;
           USVString mediaType = "";
-          NFCWatchMode mode = "web-nfc-only";
+          NFCAccept accept = "web-nfc-only";
           AbortSignal? signal;
         };
       </pre>
       <p>
-        The <dfn>NFCWatchOptions.url</dfn> property
+        The <dfn>NFCReaderOptions.url</dfn> property
         denotes the <a>URL pattern</a> which is used for matching the
         <a>Web NFC Id</a> of <a>Web NFC message</a>s which are being read.
         The default value <code>""</code> means that no matching happens.
       </p>
       <p>
-        The <dfn>NFCWatchOptions.recordType</dfn> property
+        The <dfn>NFCReaderOptions.recordType</dfn> property
         denotes the enum value which is used for matching the
         <code>
         <a href="#idl-def-nfcrecordtype">recordType</a></code> property of each
@@ -1708,19 +1709,19 @@ writer.push({ records: [
         <a href="#steps-watch">NFC watch algorithm</a>.
       </p>
       <p>
-        The <dfn>NFCWatchOptions.mediaType</dfn> property
+        The <dfn>NFCReaderOptions.mediaType</dfn> property
         denotes the <a>match pattern</a> which is used for matching the
         <code><a href="#dom-nfcrecord-mediatype">mediaType</a></code> property of each
         <code><a>NFCRecord</a></code> object in a <a>Web NFC message</a>.
         The default value <code>""</code> means that no matching happens.
       </p>
       <p>
-        The <dfn>NFCWatchOptions.mode</dfn> property denotes the
-        <code><a>NFCWatchMode</a></code> value telling whether only
+        The <dfn>NFCReaderOptions.accept</dfn> property denotes the
+        <code><a>NFCAccept</a></code> value telling whether only
         <a>Web NFC content</a> or any <a>NFC content</a> will be watched.
       </p>
       <p>
-        The <dfn>NFCWatchOptions.signal</dfn> property allows to abort
+        The <dfn>NFCReaderOptions.signal</dfn> property allows to abort
         the watch.
       </p>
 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1771,7 +1782,37 @@ <h2>The <dfn>NFCWatchMode</dfn> enum</h2>
...
+          Let <var>reader</var> be a new <a><code>NFCReader</code></a> object.
+        </li>
+        <li>
+          Let <var>options</var> be first argument to constructor.
+        </li>
+        <li>
+          <a>For each</a> <var>key</var> → <var>value</var> of <var>options</var>:
+          <ul>
+            <li>
+              If the <a>NFCReaderOptions</a>
```
kenchris: This should probably be NFCReader, because otherwise I will copy `signal` as well. Maybe I should just list our the attributes? @anssiko

index.html
```diff
@@ -1430,10 +1430,16 @@ <h2>The <dfn>NFCRecordType</dfn> enum</h2>
...
     interface NFCReader : EventTarget {
-      Promise&lt;void&gt; watch(optional NFCWatchOptions options);
+      readonly attribute USVString url;
+      readonly attribute NFCRecordType recordType;
+      readonly attribute USVString mediaType;
+      readonly attribute NFCAccept accept;
+
       attribute EventHandler onreading;
+
+      Promise&lt;void&gt; watch();
```
zolkis, line 1437: IMHO removing the argument of `watch()` and passing it to the constructor is half-breed solution. 
I would use either `watch()` with the options, or constructor with the function passed together with options.

index.html
```diff
@@ -1678,28 +1689,28 @@ <h2>The <dfn>NFCPushTarget</dfn> enum</h2>
...
+        <a>NFCReaderOptions</a> dictionary is used:
       </p>
       <pre class="idl">
-        dictionary NFCWatchOptions {
+        dictionary NFCReaderOptions {
           USVString url = "";
           NFCRecordType recordType;
           USVString mediaType = "";
-          NFCWatchMode mode = "web-nfc-only";
+          NFCAccept accept = "web-nfc-only";
```
zolkis, line 1692: Isn't `accept` a bit overloaded? I feel that `mode` was better compromise for capturing what it is an accept-policy then just `accept`. Unless `accept` has one well defined meaning and usage in other specs.

index.html
```diff
@@ -1742,18 +1753,18 @@ <h2>The <dfn>NFCPushTarget</dfn> enum</h2>
...
+    <h2>The <dfn>NFCAccept</dfn> enum</h2>
     <p>
       This enum defines the set of known values denoting whether
       <code><a href="#dom-nfcreader-watch">watch()</a></code> operation should
       <a>dispatch NFC content</a> for <a>Web NFC content</a> or any
       <a>NFC content</a>
     </p>
     <pre class="idl">
-      enum NFCWatchMode {
+      enum NFCAccept {
```
zolkis, line 1757: Same comment as above with `accept` being overloaded term in comms.

index.html
kenchris: This is an incremental change to make sure it all works, `watch()` will be removed soon enough and it will be activated when event listener is attached

index.html
kenchris: It is used by http at least. Maybe we can find a better name, but for now, WatchMode doesn't make so much sense as I am removing watch

index.html
```diff
@@ -1469,6 +1475,11 @@ <h2>The <dfn>NFCRecordType</dfn> enum</h2>
     The <dfn>NFCReader.onreading</dfn> is an EventHandler which is called to notify
     that new reading is available.
   </p>
+  <p>
+    The <dfn>NFCReader.url</dfn>, <dfn>NFCReader.recordType</dfn>,
+    <dfn>NFCReader.mediaType</dfn> and <dfn>NFCReader.accept</dfn> correlates to
+    the members of <a>NFCReaderOptions</a>.
```
littledan: Nit: s/correlates the/have the values of the corresponding/ (when correlates is used as a transitive verb, the subject is the person doing the analysis)

index.html
littledan: `accept` SGTM; the analogy makes sense

index.html
littledan: This name isn't web-exposed, so I don't think it matters so much to decide the perfect one right now.

index.html
littledan: Nit: If you have a normative definition here, then lines 1478-1482 should be a note. But, in WebIDL, attributes are getter/setters, not own properties, so their definition isn't determined by setting the attribute to a value, but saying that, when you read the attribute, you get that value.

index.html
```diff
@@ -1771,7 +1782,37 @@ <h2>The <dfn>NFCWatchMode</dfn> enum</h2>
...
+        </li>
+        <li>
+          Let <var>options</var> be first argument to constructor.
+        </li>
+        <li>
+          <a>For each</a> <var>key</var> → <var>value</var> of <var>options</var>:
+          <ul>
+            <li>
+              If the <a>NFCReaderOptions</a>
+              <a data-lt="list-contain">contains</a> <var>key</var>,
```
littledan: WebIDL will already filter out the irrelevant keys, so I don't think this conditional is necessary. Or did you mean to check whether NFCReader contains the key, to filter out signal?

index.html
```diff
@@ -2633,7 +2674,7 @@ <h2>The <dfn>NFCWatchMode</dfn> enum</h2>
       </p>
       <p>
         An <dfn>NFC watch</dfn> is referring to a tuple of a <code><a>Promise</a></code>,
```
littledan: English nit: s/is referring to/refers to/

index.html
```diff
@@ -2710,7 +2748,7 @@ <h2>The <dfn>NFCWatchMode</dfn> enum</h2>
                 <code>"NotSupportedError"</code> and return <var>p</var>.
               </li>
               <li>
-                If the <var>options</var>.url is not an empty <code>String</code>
+                If the <var>reader_instance</var>.url is not an empty <code>String</code>
```
littledan: Hmm, somehow this text should clarify, you don't mean `Get(reader_instance, "url")` (since that could be modified using Object.defineProperty), but you mean the original value. You could specify this using internal slots, or a single big internal slot for the options dictionary.

index.html
littledan: Incremental changes SGTM. I like the plan of activating when the event listener is attached.

@zolkis You might want to avoid using analogies like "half-breed", which could be considered demeaning.

index.html
```diff
@@ -1430,10 +1430,16 @@ <h2>The <dfn>NFCRecordType</dfn> enum</h2>
...
     };
 
-    [Constructor(), SecureContext, Exposed=Window]
+    [Constructor(optional NFCReaderOptions options), SecureContext, Exposed=Window]
     interface NFCReader : EventTarget {
-      Promise&lt;void&gt; watch(optional NFCWatchOptions options);
+      readonly attribute USVString url;
+      readonly attribute NFCRecordType recordType;
+      readonly attribute USVString mediaType;
+      readonly attribute NFCAccept accept;
```
littledan: Just curious, what is the motivation for exposing these to read out? It seems like this will just be the options that were passed in, so I don't really see the use case. Did you just add these attributes so that they were stored internally in the NFCReader? You  might want to use internal slots for that purpose, if so.

littledan (commented):
The biggest thing I'd want clarified before approving is whether you intended to expose the options as attributes; besides that, all comments are editorial and would be fine to clean up later.

index.html
kenchris: This was based of generic sensor spec (geolocation sensor)

index.html
kenchris: Maybe you are right... I was thinking about it being nice to store if you had multiple readers but I guess we can just drop that.

index.html
```diff
@@ -2974,39 +3013,36 @@ <h3>Receiving and parsing content</h3>
...
+            algorithm, providing <var>reader_instance</var>.[[\Url]] as a
+            <a>URL pattern</a> and
             <var>message</var>.url as a <code><a>Web NFC Id</a></code>. If
             <a href="#dfn-match-web-nfc-id-with-url-pattern">URL pattern match</a>
             algorithm returns <code>false</code>, skip to the next <a>NFC watch</a>.
           </li>
           <li>
-            If <var>options</var>.recordType is <a href=#dfn-not-present>present</a> and
-            it is not equal to any <var>record</var>.recordType where <var>record</var>
+            If <var>reader_instance</var>.[[\RecordType] is <a href=#dfn-not-present>present</a> and
```
littledan: Nit: missing ]

index.html
```diff
@@ -2974,39 +3013,36 @@ <h3>Receiving and parsing content</h3>
...
+            it is not equal to any <var>record</var>.[[\recordType]] where <var>record</var>
```
littledan: Where is the type of these records defined?

Probably either both this line and line 3041 should use internal slot notation or property notation, but not one of each.

littledan (approved):

index.html
kenchris: These are internal slots on the NFCReader. But the NFCRecord (the data we read) has these as properties

kenchris: http://localhost:8000/#dom-nfcrecord

kenchris: Maybe we need to define the internal slots for the NFCReader somewhere. Can be a follow-up

PR #181, opened 2018-12-07
Now that we use event listeners for watches, remove AbortSignal

kenchris:


littledan (commented):
I don't see any logic to remove the watch when the EventListener is removed. Probably that should be added.

kenchris:
The watch method will be removed soon at it will all depend on add / removing event listeners like @reillyeon suggested (same as geolocation spec)

littledan (approved):
OK, the change looks good as long as the logic is added back in a follow-on patch

kenchris:
Sure, will do that.

PR #182, opened 2018-12-10
Fix up references to the environment settings object

littledan:
- Reference the current settings object, not the incumbent settings object
- Fix various "type errors" where a nonsensical thing was being referenced

### DIFF
--- a/index.html
+++ b/index.html
@@ -184,8 +184,8 @@
     <a href="https://html.spec.whatwg.org/multipage/#global-object">
     <dfn>global object</dfn></a>,
 
-    <a href="https://html.spec.whatwg.org/multipage/#incumbent-settings-object">
-    <dfn>incumbent settings object</dfn></a>,
+    <a href="https://html.spec.whatwg.org/multipage/#current-settings-object">
+    <dfn>current settings object</dfn></a>,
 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1569,14 +1575,14 @@ <h2>The <dfn>NFCRecordType</dfn> enum</h2>
...
-    To <dfn>release NFC</dfn> on a <a>settings object</a>, perform the following steps:
+    To <dfn>release NFC</dfn> on an <a>environment settings object</a>, perform the following steps:
   </p>
   <ol id="steps-nfc-release">
     <li>
       <a>Suspend NFC</a>.
     </li>
     <li>
-      For each pending <a>NFCWriter.push</a> associated with the <a>settings object</a>:
+      For each pending <a>NFCWriter.push</a> associated with the <a>current settings object</a>:
```
domenic: This uses current but the algorithm takes a different ESO as an argument. Is that intentional?

domenic (commented):
LGTM with question

index.html
littledan: I don't think "current" makes much sense here; probably better to use the one associated with the document. Pushed a patch to clarify.

kenchris (approved):

PR #183, opened 2018-12-11
As event listeners cannot have side-effects, intro start/stop

kenchris:
This follows the Generic Sensor approach, except `start` returns a promise which
can be fixed in a future PR.


littledan (approved):
Aligning with the sensor API sounds like a great plan, and this patch takes us closer to that in a clean way.

PR #184, opened 2018-12-12
Introduce error event handler and make start not return a promise

kenchris:


zolkis:
I thought Promises were excellent for modeling NFC interactions. But I guess this is harmonization with Generic Sensor. Even though the modeled interactions are different.

index.html
```diff
@@ -2680,43 +2703,89 @@ <h2>The <dfn>NFCAccept</dfn> enum</h2>
...
-            an <a>NFC peer</a> or <a>NFC tag</a> in proximity range, reject
-            <var>p</var> with <code>"NotSupportedError"</code> and return <var>p</var>.
-          </li>
-          <li>
-            Add <var>reader_instance</var> to the <a>activated reader objects</a>.
+            an <a>NFC peer</a> or <a>NFC tag</a> in proximity range, then
+            <ol>
+              <li>
+                Let <var>e</var> be the result of <a>creating</a> a
+                <code>"NotSupportedError"</code> DOMException.
```
kenchris: Maybe this should be moved to ctor?

index.html
```diff
@@ -2680,43 +2703,89 @@ <h2>The <dfn>NFCAccept</dfn> enum</h2>
...
             <ol>
               <li>
                 If the <a>obtain reading permission</a> steps return
-                <code>false</code>, then reject <var>p</var> with
-                <code>"SecurityError"</code> and return <var>p</var>.
+                <code>false</code>, then
+                <ol>
+                  <li>
+                    Let <var>e</var> be the result of <a>creating</a> a
+                    <code>"SecurityError"</code> DOMException.
```
kenchris: Maybe this should be moved to ctor?

kenchris:
> I thought Promises were excellent for modeling NFC interactions. But I guess this is harmonization with Generic Sensor. Even though the modeled interactions are different.

Yes and at TPAC in Lisbon it was decided that start() should not return a promise.

The sensors are new in the way that they are constructable objects (fits more with Node.js etc) and care has been taken to minimize fingerprinting.

index.html
```diff
@@ -3141,7 +3220,7 @@ <h3>
 <section> <h2>Acknowledgements</h2>
   <p>
     The editors would like to thank Jeffrey Yasskin, Anne van Kesteren,
-    Anssi Kostiainen, Domenic Denicola, Jonas Sicking,
+    Anssi Kostiainen, Domenic Denicola, Daniel Ehrenberg, Jonas Sicking,
```
littledan, line 3223: Aww, thanks

littledan (approved):
LGTM; probably will make sense to add an onactivate event as well (since this patch takes away the ability to see when that happens).

littledan:
@zolkis How are the interactions different in NFC vs Sensor, in a way that makes sense to reflect in an API?

zolkis:
@littledan 
> How are the interactions different in NFC vs Sensor, in a way that makes sense to reflect in an API?

IMHO:
- An NFC interaction is defined by the user consciously approaching an NFC device to another (tag, peer). This is a finite one-time interaction with zero or one successful data transactions. There is one success or error outcome. Modeling with Promises makes sense.
- A Sensor interaction can be one-off, multiple-off, manual/automatic/wakeup activated, various sampling and reporting rate etc. Modeling with Events plus stuff makes sense.

IMHO even in the Web Platform, "API harmonization" should not mean "one style force-fits all".

Anyway in this specific case, the resulting API is bearable even if feels like overkill, besides it's already merged, so it's fine if it's good for developers.

PR #186, opened 2019-01-02
Improve "Create a Web NFC Id"

### LINKED ISSUE #155: "Creating a Web NFC Id" not very rigorous
> Let id be the ASCII serialized origin of the browsing context, appended with url. 

Which browsing context?

"x appended with y" is confusing to me. Is that `xy` or `yx`?

> If id is a not a valid URL, throw a "SyntaxError" exception and abort these steps. 

What is the algorithm for determining "valid URL"? Is it https://url.spec.whatwg.org/#valid-url-string ? My guess is you want to apply the basic URL parser and throw an exception if parsing fails.

Also, the name "url" is very confusing. It should be "urlPath", I think.

> Return id, encoded in UTF-16. 

Assuming UTF-16 is necessary for interop with legacy systems (we prefer UTF-8 on the web), you should use https://encoding.spec.whatwg.org/#decode. And specify UTF-16BE or UTF-16LE.

Note that even then a UTF-8 BOM would force UTF-8 encoding. Not sure if that's a problem or not...

Honry:
Fixed #155, unified the terms of Encoding as well.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2514,8 +2514,8 @@ writer.push({ records: [
         run these steps:
         <ol>
           <li>
-            Let <var>id</var> be the <a>serialized origin</a> of the current
-            <a>browsing context</a>.
+            Let <var>id</var> be the <a>serialized origin</a> of the
+            <a>current settings object</a>.
           </li>
           <li>
             Append <var>urlPath</var> to <var>id</var>.
--- a/index.html
+++ b/index.html
@@ -166,6 +166,7 @@
     <dfn data-cite="!URL#concept-url-host">URL host</dfn>,
     <dfn data-cite="!URL#concept-url-path">URL path</dfn>,
     <dfn data-cite="!URL#concept-url">URL record</dfn>,
+    <dfn data-cite="!URL#valid-url-string">valid URL string</dfn>,
     <dfn data-cite="!URL#absolute-url-string">absolute-URL string</dfn>,
     <dfn data-cite="!URL#path-absolute-url-string">path-absolute-URL string</dfn> and
     <dfn data-cite="!URL#concept-basic-url-parser">basic URL parser</dfn> are defined in [[!URL]].
@@ -192,7 +193,8 @@
     <dfn data-cite="!HTML#eventhandler">EventHandler</dfn>,
     <dfn data-cite="!HTML#focus">focus</dfn>,
     <dfn data-cite="!HTML#handler-onblur">lose focus</dfn>,
-    <dfn data-cite="!HTML#unloading-document-cleanup-steps">unloading document cleanup steps</dfn>.
+    <dfn data-cite="!HTML#unloading-document-cleanup-steps">unloading document cleanup steps</dfn>,
+    <dfn data-cite="!HTML#utf-16-encoding">UTF-16 encoding</dfn>.
   </p>
   <p>
     A <a>browsing context</a> refers to the environment in which
@@ -272,8 +274,9 @@
     are defined in [[!DOM]].
   </p>
   <p>
-    The algorithms <dfn data-cite="!ENCODING#utf-8-encode">utf-8 encode</dfn>, and
-    <dfn data-cite="!ENCODING#utf-8-decode">utf-8 decode</dfn> are defined in [[!ENCODING]].
+    <dfn data-cite="!ENCODING#utf-8">UTF-8 encoding</dfn>,
+    <dfn data-cite="!ENCODING#encode">encode</dfn>, and
+    <dfn data-cite="!ENCODING#decode">decode</dfn> are defined in [[!ENCODING]].
   </p>
   <p>
     <dfn>IANA media type</dfn>s (formerly known as MIME types) are defined in
@@ -2299,8 +2302,8 @@ writer.push({ records: [
                   <li>
                     If <var>record</var>.mediaType contains the parameter
                     "charset=UTF-8" (e.g. "text/plain; charset=UTF-8;"),
-                    then set bit 7 (MSB) to 0 (UTF-8 encoding),
-                    otherwise to 1 (UTF-16 encoding).
+                    then set bit 7 (MSB) to 0 (<a>UTF-8 encoding</a>),
+                    otherwise to 1 (<a>UTF-16 encoding</a>).
                   </li>
                   <li>Set bit 6 to 0.</li>
                   <li>
@@ -2316,8 +2319,9 @@ writer.push({ records: [
               <li>
                 Set the consecutive payload octets to
                 <var>record</var>.data in the encoding set by the
-                MSB bit of the first payload octet (by default UTF-16,
-                or UTF-8 on special request).
+                MSB bit of the first payload octet (by default
+                <a href="#dfn-utf-16-encoding">UTF-16</a>,
+                or <a href="#dfn-utf-8-encoding">UTF-8</a> on special request).
               </li>
             </ol>
           </li>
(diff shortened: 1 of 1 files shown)


kenchris:
Hi @domenic as you filed the issue, would you mind having a look?

index.html
```diff
@@ -2299,8 +2302,8 @@ <h2>The <dfn>NFCAccept</dfn> enum</h2>
                   <li>
                     If <var>record</var>.mediaType contains the parameter
                     "charset=UTF-8" (e.g. "text/plain; charset=UTF-8;"),
-                    then set bit 7 (MSB) to 0 (UTF-8 encoding),
-                    otherwise to 1 (UTF-16 encoding).
+                    then set bit 7 (MSB) to 0 (<a>UTF-8 encoding</a>),
+                    otherwise to 1 (<a>UTF-16 encoding</a>).
```
domenic, line 2306: Which UTF-16 encoding, BE or LE?

index.html
```diff
@@ -2339,7 +2343,7 @@ <h2>The <dfn>NFCAccept</dfn> enum</h2>
             and abort these steps.
           </li>
           <li>
-            If the <var>record</var>.data is not a valid <a>URL</a>, throw a
+            If the <var>record</var>.data is not a <a>valid URL string</a>, throw a
```
domenic: As I mentioned in the issue, I find it highly unlikely browsers will implement the valid URL string algorithm (which is meant for conformance checkers). Instead, you should probably parse the URL string into a URL record, and throw if the parsing fails.

index.html
```diff
@@ -2479,12 +2483,12 @@ <h2>The <dfn>NFCAccept</dfn> enum</h2>
...
             Let <var>payload</var> be the result of invoking
-            <a>create a Web NFC Id</a> given <var>url</var>.
+            <a>create a Web NFC Id</a> given <var>urlPath</var>.
             If this throws an exception, re-throw it.
           </li>
           <li>
-            Set <var>ndef.PAYLOAD</var> to the <a>URL path</a> of the
-            <a>browsing context</a>, encoded as UTF-16.
+            Set <var>ndef.PAYLOAD</var> to the <a>URL path</a> of the current
+            <a>browsing context</a>, <a href="#dfn-encode">encoded</a> as <a>UTF-16 encoding</a>.
```
domenic, line 2495: @annevk, from what I can tell it's not possible to use the encoding standard to encode to UTF-16. But it seems like we have a web spec that needs that, presumably for interop with legacy systems? Your help and review would be appreciated.

index.html
```diff
@@ -2502,23 +2506,20 @@ <h2>The <dfn>NFCAccept</dfn> enum</h2>
...
       <p>
-        To <dfn>create a <a>Web NFC Id</a></dfn> given <var>url</var>,
+        To <dfn>create a <a>Web NFC Id</a></dfn> given <a>URL path</a> <var>urlPath</var>,
         run these steps:
         <ol>
           <li>
-            Let <var>id</var> be the <a>serialized origin</a> of the
-            <a>browsing context</a>, appended with <var>url</var>.
+            Let <var>id</var> be the <a>serialized origin</a> of the current
+            <a>browsing context</a>, append <var>urlPath</var> to <var>id</var>.
```
domenic: What is "the current browsing context"? That's not a defined term.

domenic: These appear to be two separate operations, which should be separate steps, not a comma-joined single sentence.

domenic (changes requested):
Still some big issues, especially around encoding which I'm not even sure what the right thing is anymore. Perhaps someone could start by explaining why you need to encode a string to UTF-16? Is there some part of the NFC spec that requires that, or similar? Could you point to it?

index.html
annevk: Does NFC define some high-level way to set the PAYLOAD field from scalar values?

Which UTF-16 encoding is in use here? This would indeed require reverting https://github.com/whatwg/encoding/commit/8360f775c8df145f649047c7d59c5ff733ade112 somewhat. Perhaps an initial step would be to put those algorithms in NFC with an issue against whatwg/encoding for potential upstreaming.

Looking over this PR and skimming through Web NFC it seems like there's multiple places that are rather hand-wavy about how bytes and code points are handled. Those all need tidying up.

kenchris:
@zolkis could you comment as you wrote those parts of the spec?

index.html
Honry: Eh, I didn't find any definition for "the current browsing context" in HTML, and all instances in HTML are like this, 

e.g. https://html.spec.whatwg.org/#dom-opener

index.html
Honry: Thanks! I will fix it in next commit.

index.html
Honry: OK, I will separate them.

index.html
Honry: @zolkis, could you comment on this?

index.html
domenic: That's different; that's a property of `Window`, for which it's a bit more clear. (But still not great.) You have no such `Window` object here.

Honry: Aha, I got the point. How about using "[current settings object](https://html.spec.whatwg.org/multipage/#current-settings-object)"?

index.html
hsivonen: From the current spec and PR, it's very unclear what constraints lead to the notion that Web NFC needs to encode into UTF-16BE or UTF-16LE (which one?). At present, I think we aren't at the point where such a need is well enough motivated to change the Encoding Standard.

However, if there is a genuine need to enable taking a DOMString and generating bytes in UTF-16BE or UTF-16LE, it should probably be done outside the general framework of the Encoding Standard to make it absolutely clear that it doesn't apply to form submission, URL queries or the `TextEncoder` API.

zolkis: A few remarks based on the NDEF spec:
- NDEF mandates no specific processing model for NDEF messages. The usage of the payload types is entirely at the discretion of the user application.
- Encoding of URI characters which fall outside the US-ASCII range is left to the NDEF
application. Therefore, an NDEF parser must not assume any particular encoding for this field.
- To indicate the payload type, the TYPE field (max 255 bytes) defines it with the semantics set by the TNF field (e.g. URI, IANA media type, NFC Forum External Type, unknown, NFC Forum Well Known Type etc.).
- Transmission order is big-endian (most significant octet is transmitted first, and leftmost bit is the most significant bit).
- I don't remember any more why we specified UTF-16 at that time. WFIW `neard` supported UTF-8 and UTF-16 at that time. The encoding support might be limited by the implementation and otherwise should be handled by the application (from NFC pov the implementation is an application, so it can well define the required encoding; hence the Web NFC spec could also use/mandate a given encoding if there was a use case for that). For Web NFC we should comply with/refer to the (Web) Encoding Standard.

zolkis: (BTW apologies for the late answer, jut got back from holidays).

annevk: If you could always produce UTF-8 and consume UTF-16(LE/BE?) as needed, that'd be ideal. It sounds like that might be doable?

(It might be good for someone to split this into a couple of issues as tracking all the various things in one PR will make it easy to overlook things.)

domenic: Should it even be a string at all? It sounds like from your comments the application should probably be doing any encoding/decoding, and the Web NFC spec might want to just provide a mechanism for exchanging raw bytes.

zolkis: @domenic That is right, the spec should just provide a bitpipe to the app. For convenience we also have the freedom to pick a certain number of supported encodings but that would be a restriction from NFC point of view and would add complexity to the spec/implementations.

Honry: @zolkis, so do we have a conclusion here? Should I create a separate issue to track encoding/decoding problem?

zolkis: Yes, we should track this in a separate issue. The conclusion is that we need mechanisms/algorithms for how to handle encoding. It's not a simple property.

zolkis:
I suggest we merge this and track the encoding in #187

kenchris (approved):
Approved

PR #188, opened 2019-02-13
Remove NFCAccept as you can accomplish the same by not doing url filtering

kenchris:
If you don't filter by URL, eg. the "url" field of the dictionary is equal to "", then all NFC devices are accepted.

If the "url" field is different from "", that means it will be matched against the Web NFC Record, thus the presence of such is required.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1686,7 +1686,8 @@ writer.push({ records: [
     </dl>
   </section> <!-- NFCPushTarget enum -->
 
-  <section> <h3>The <dfn>NFCReaderOptions</dfn> dictionary</h3>
+  <section>
+    <h3>The <dfn>NFCReaderOptions</dfn> dictionary</h3>
       <p>
         To describe which messages an application is interested in, the
         <a>NFCReaderOptions</a> dictionary is used:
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1696,7 +1697,6 @@ <h2>The <dfn>NFCPushTarget</dfn> enum</h2>
           USVString url = "";
           NFCRecordType recordType;
           USVString mediaType = "";
-          NFCAccept accept = "web-nfc-only";
```
zolkis, line 1699: Do you mean that if reader options is specified (recordType or mediaType is present), then by default it's web-nfc-only and otherwise not?

Shouldn't this be explicitly mentioned in the spec?

kenchris: If you add any filter (URL) it will require the NFC record

zolkis: OK, but I think then this needs to be stated in prose in the spec and if so, it should be in this PR. Could you add a clarifying statement?

kenchris: I could add a note, but it is clear from the algorithm. You want me to add a comment to the commit message?

zolkis: Yes, it is implied by the algorithm, but I wonder people would really realize the behavior here. Adding a note would be good IMHO, at least for now.

kenchris:
OK, fixed @zolkis comments

PR #189, opened 2019-02-15
Support opting into supporting reading and writing to all NDEF devices

### LINKED ISSUE #160: Create error message when using proprietary tags
The most common feedback we have gotten is the lack of support for the super popular Mifare Classic tags. These are NDEF compatible but only supported by NXP readers.

We should support reading/writing to these on compatible hardware as an explicit opt-in

kenchris:
The most common feedback we have gotten is the lack of support for
the super popular Mifare Classic tags. These are NDEF compatible but
not part of the NFC spec and only supported by NXP reader chips.

This adds an explicit opt-in to reading and writing to these tags

Fixes #160

### DIFF
--- a/index.html
+++ b/index.html
@@ -1614,6 +1614,46 @@ writer.push({ records: [
   </p>
   </section> <!-- release NFC -->
 
+  <section data-dfn-for="NDEFCompatibility">
+      <h3>The <dfn>NDEFCompatibility</dfn> enum</h3>
+        <p>
+          To describe what NDEF compatible devices are accepted as
+          vendor specific tags exist that support NDEF but which
+          are not universally supported by all NFC readers, or by
+          the NFC standard.
+        </p>
+        <pre class="idl">
+          enum NDEFCompatibility {
+            "nfc-forum",
+            "vendor",
+            "any"
+          };
+        </pre>
+        <p>
+          <dl>
+            <dt><dfn>nfc-forum</dfn></dt>
+            <dd>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1614,13 +1614,54 @@ <h2>The <dfn>NFCRecordType</dfn> enum</h2>
...
+          To describe what NDEF compatible devices are accepted as
+          vendor specific tags exist that support NDEF but which
+          are not universally supported by all NFC readers, or by
+          the NFC standard.
+        </p>
+        <pre class="idl">
+          enum NDEFCompatibility {
+            "universal",
+            "vendor",
+            "all"
```
zolkis: IMHO ```"all"``` is not suggestive for its stated meaning
```The enum value representing all NDEF compatible devices that the reader chip can read.```
I think "ndef" would be better.

Then, "universal" could become "all-nfc" or just "nfc"?
```The enum value representing all active and passive NFC devices, supported by the NFC standard.```

kenchris: Maybe "any" instead of "all" - that is actually what I meant.

What do you think about "nfc-forum" ?

zolkis: "any" sounds good, in the sense that any device/tag the reader can communicate with. But the text specifies it has to be NDEF based, so it actually means "any NDEF". If we really mean NDEF-only, then IMHO we should use "ndef", otherwise we can use "any".

"nfc-forum" is fine.

kenchris:
Fixed @zolkis ' comments

PR #191, opened 2019-02-15
Prefix NDEF related objects with NDEF instead of NFC

kenchris:
This fits other specs like Web Bluetooth that also uses GATT in API names


zolkis:
Ship it :)

PR #192, opened 2019-02-18
Make use of the MIME sniff spec

### LINKED ISSUE #158: Use MIMESNIFF instead of RFC2046 for some things
Although RFC2046 defines MIME types in general, and is good at that, for rigorous parsing, serialization, classification, and discussion of the MIME type's components, we have https://mimesniff.spec.whatwg.org/.

In particular https://w3c.github.io/web-nfc/#dfn-json-compatible-iana-media-type seems like it can be replaced with https://mimesniff.spec.whatwg.org/#json-mime-type . Any talk of "validity" can be replaced with https://mimesniff.spec.whatwg.org/#parse-a-mime-type or https://mimesniff.spec.whatwg.org/#parse-a-mime-type-from-bytes , which then gives you a specification data structure to talk about. E.g. https://w3c.github.io/web-nfc/#mapping-string-to-ndef could become much more rigorous, doing parsing and then introspecting the components.

kenchris:
Fixes #158

### DIFF
--- a/index.html
+++ b/index.html
@@ -200,8 +200,13 @@
     <code><a>Window</a></code> objects.
   </p>
   <p>
-    The term <dfn data-cite="!MIMESNIFF#valid-mime-type">valid MIME type string</dfn>
-    is defined in [[!MIMESNIFF]].
+    The term 
+    <dfn data-cite="!MIMESNIFF#mime-type">MIME type</dfn>, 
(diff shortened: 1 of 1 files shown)


kenchris:
@domenic could you have a look at this? thanks

kenchris:
OK, going to merge this, if there are comments I can fix those in follow up patches

PR #193, opened 2019-02-19
Use standard JSON parsing/serializing primitives

### LINKED ISSUE #154: Consider using standard JSON parsing/serializing primitives
In particular, https://infra.spec.whatwg.org/#json

Currently the spec appears to use UTF-16 encoding, which is quite unfortunate as that's legacy on the web platform. Maybe there is some domain-specific reason why it's necessary, in which case that should be noted at the same point the serializing/parsing is done. And probably similar rigor should be applied as is done in the Infra spec, instead of just saying "serializing" but linking to the ES spec's JSON.stringify definition.

kenchris:
- JSON is stored on NFC tags as UTF-8 now (not UTF-16BE)
- MIME types are parsed and serialized using standard primitives

Fixes #154

### DIFF
--- a/index.html
+++ b/index.html
@@ -200,8 +200,8 @@
     <code><a>Window</a></code> objects.
   </p>
   <p>
-    The term 
-    <dfn data-cite="!MIMESNIFF#mime-type">MIME type</dfn>, 
+    The term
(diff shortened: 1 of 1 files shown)


zolkis (approved):
Much better now, thanks.

PR #196, opened 2019-02-20
Tighten up the language of the algorithms

### LINKED ISSUE #194: Byte and byte sequence vs octet/octets
Hi @annevk and @domenic 

The current spec uses octets a lot (as written by @zolkis) but I notice that INFRA uses byte and byte sequences. Any recommendation here?

kenchris:
- Tighten up the language and align them with modern standards
- Use byte instead of octet

Fixes #194

### DIFF
--- a/index.html
+++ b/index.html
@@ -228,6 +228,7 @@
     [[!RFC5234]].
   </p>
   <p>
+    <dfn data-cite="!WEBIDL#dfn-throw"><code>throw</code></dfn>,
     <dfn data-cite="!WEBIDL#dfn-create-exception"><code>creating</code></dfn>,
     <dfn data-cite="!WEBIDL#idl-DOMString"><code>DOMString</code></dfn>,
     <dfn data-cite="!WEBIDL#idl-dictionary"><code>Dictionary</code></dfn>,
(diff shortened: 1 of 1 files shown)


zolkis (approved):
Thanks!

index.html
```diff
@@ -1854,18 +1856,16 @@ <h3>The <dfn>NFCReaderOptions</dfn> dictionary</h3>
           </li>
           <li>
             If there is no underlying <a>NFC Adapter</a>, or if a connection cannot
-            be established, then reject <var>p</var> with a <code>"<a>NotReadableError</a>"</code>
-            <code><a>DOMException</a></code> and return <var>p</var>.
+            be established, then reject <var>p</var> with a <code><a>NotReadableError</a></code>
+            and return <var>p</var>.
```
annevk, line 1860: This seems like a regression to me, FWIW.

kenchris: This fits with now it is done in WHATWG URL spec

kenchris: image

Or you want to make clear that it is a DOMException when used with a promise and not just thrown?

annevk: A `TypeError` is not a `DOMException`, it's a separate class. Look at https://fetch.spec.whatwg.org/ for a specification that throws both `TypeError` and "`AbortError`" `DOMException`.

PR #198, opened 2019-02-20
Change default encoding to UTF-8 and make sure we can still read UTF-16BE

### LINKED ISSUE #187: Specify encoding/decoding algorithms
As discussed in #186, we need to fix the way encoding/decoding is currently specified.

kenchris:
Fixes #187

### DIFF
--- a/index.html
+++ b/index.html
@@ -189,8 +189,7 @@
     <dfn data-cite="!HTML#eventhandler">EventHandler</dfn>,
     <dfn data-cite="!HTML#focus">focus</dfn>,
     <dfn data-cite="!HTML#handler-onblur">lose focus</dfn>,
-    <dfn data-cite="!HTML#unloading-document-cleanup-steps">unloading document cleanup steps</dfn>,
-    <dfn data-cite="!HTML#utf-16-encoding">UTF-16 encoding</dfn>.
+    <dfn data-cite="!HTML#unloading-document-cleanup-steps">unloading document cleanup steps</dfn>.
   </p>
   <p>
     A <a>browsing context</a> refers to the environment in which
@@ -258,8 +257,10 @@
   </p>
   <p>
     <dfn data-cite="!INFRA#ascii-decode">ASCII decode</dfn>,
+    <dfn data-cite="!INFRA#ascii-lowercase">ASCII lowercase</dfn>,
     <dfn data-cite="!INFRA#string">string</dfn>,
     <dfn data-cite="!INFRA#byte">byte</dfn>,
+    <dfn data-cite="!INFRA#byte-sequence">byte sequence</dfn>,
     <dfn data-cite="!INFRA#ordered-set">set</dfn>,
     <dfn data-cite="!INFRA#map-exists">exists</dfn>,
     <dfn data-cite="!INFRA#list">list</dfn>,
@@ -282,6 +283,7 @@
   </p>
   <p>
     <dfn data-cite="!ENCODING#utf-8">UTF-8 encoding</dfn>,
+    <dfn data-cite="!ENCODING#utf-8-decode">UTF-8 decode</dfn>,
     <dfn data-cite="!ENCODING#encode">encode</dfn>, and
     <dfn data-cite="!ENCODING#decode">decode</dfn> are defined in [[!ENCODING]].
   </p>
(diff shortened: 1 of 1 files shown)


zolkis:
AFAICT we could just use raw byte stream, the app doing the conversion.
See https://github.com/w3c/web-nfc/pull/186#discussion_r245924367
and https://github.com/w3c/web-nfc/pull/186#discussion_r246036580 
and https://github.com/w3c/web-nfc/issues/187#issuecomment-452809060

kenchris:
I am not sure that I want to force people to have to do conversion for just writing text (Well-known) - if they want to do fancy stuff, they can encode however they want using the opaque field (external type).

This makes UTF-8 strings default and you will have to set charset explicitly if you want to use UTF-16BE - we can of course remove that and only make sure that we can read it - but we can take that as a follow up. This pretty much fits with https://github.com/w3c/web-nfc/pull/186#discussion_r245924367

The not being a string (the next two comments) fits with external type and that is already handed, but that seems quite weird with well-known/text which is supposed to be only for strings (UTF-8 or UTF-16BE)

index.html
```diff
@@ -2314,16 +2294,20 @@ <h3>The <dfn>NFCReaderOptions</dfn> dictionary</h3>
                 <ol>
                   <li>
                     If <var>MIME type</var>.parameters[<code>"charset"</code>]
-                    <a>exists</a>, set <var>charset</var> to the value.
+                    <a>exists</a>, set <var>charset</var> to the value, <a>lowercased</a>.
+                  </li>
+                  <li>
+                    If <var>charset</var> is not equal to either <code>"utf-8"</code>,
```
annevk: either?

index.html
```diff
@@ -2314,16 +2294,20 @@ <h3>The <dfn>NFCReaderOptions</dfn> dictionary</h3>
                 <ol>
                   <li>
                     If <var>MIME type</var>.parameters[<code>"charset"</code>]
-                    <a>exists</a>, set <var>charset</var> to the value.
+                    <a>exists</a>, set <var>charset</var> to the value, <a>lowercased</a>.
```
annevk: ASCII lowercase?

index.html
```diff
@@ -2514,7 +2496,7 @@ <h3>The <dfn>NFCReaderOptions</dfn> dictionary</h3>
           </li>
           <li>
             Set <var>ndef.PAYLOAD</var> to the <a>URL path</a> of the current
-            <a>browsing context</a>, <a href="#dfn-encode">encoded</a> as <a>UTF-16 encoding</a>.
+            <a>browsing context</a>.
```
annevk, line 2500: There's no such thing as current browsing context. A browsing context also has no URL path.

index.html
```diff
@@ -2948,7 +2941,8 @@ <h3>Receiving and parsing content</h3>
               <li>
                 Let <var>MIME type</var>.parameters[<code>"lang"</code>] be the
                 result of invoking <a>ASCII decode</a> on the consecutive
-                <var>offset</var> number of <a>bytes</a>.
+                <var>offset</var> number of <a>bytes</a>
+                (From second <a>byte</a> and <var>offset</var> <a>bytes</a> ahead).
```
annevk: You want something like from the second to the offset + 1 byte, inclusive, of ndef.PAYLOAD, no?

annevk (commented):
I'm also a little surprised with the record dot field syntax used. I don't think that's standardized?

kenchris:
> I'm also a little surprised with the record dot field syntax used. I don't think that's standardized?

You mean that it is just used for internal slots? I can look at this later. In the manifest we access them as record["field"].

I filed an issue for this https://github.com/w3c/web-nfc/issues/201

index.html
kenchris: yes. Everything shifted by 1 byte

zolkis: To my understanding, from the current "pointer" counting ```offset``` number of bytes, i.e. including the first byte from the pointer, which is the second byte in the algorithm.

index.html
kenchris: We only allow this in top-level browsing context.

How do you suggest that I rewrite this?

index.html
kenchris: `as ASCII lowercase`?  or `to the result of running ASCII lowercase on value?`

index.html
kenchris: fixed

index.html
annevk: So you need to figure out what object you want the URL from and find a way from where you are now to there. I don't really see how browsing contexts relate here since they don't carry URLs.

index.html
annevk: "ASCII lowercased" would be fine, was mostly wondering what lowercased meant.

kenchris: fixed

index.html
kenchris: Fixed

index.html
kenchris: OK I will file an issue for this so we can fix it afterwards - this is a couple of other places

kenchris: Filed: https://github.com/w3c/web-nfc/issues/200

zolkis (approved):
Let's merge now and fix the rest later.

PR #205, opened 2019-02-22
Fix Web NFC id algorithm, fixes #200

### LINKED ISSUE #200: There's no such thing as current browsing context. A browsing context also has no URL path.
There's no such thing as current browsing context. A browsing context also has no URL path.

kenchris:


zolkis (approved):
Good catch.

zolkis:
Well, dot notation is more intuitive and more exact than saying "some object's something". Now we need to standardize genitivus.

kenchris:
But that is what all other specs use. Also this allows us to link to the definitions

PR #210, opened 2019-03-04
Fix bug in examples

Honry:
the retval of `onreading` should be an `NFCReadingEvent` rather than `message`.

### DIFF
--- a/index.html
+++ b/index.html
@@ -679,8 +679,8 @@ writer.push({
 const writer = new NFCWriter();
 const reader = new NFCReader();
 
-reader.onreading = message => {
-  if (message.records[0].recordType == 'empty') {
+reader.onreading = event => {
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -679,18 +679,18 @@
 const writer = new NFCWriter();
 const reader = new NFCReader();
 
-reader.onreading = message => {
-  if (message.records[0].recordType == 'empty') {
+reader.onreading = event => {
+  if (event.message.records[0].recordType == 'empty') {
```
kenchris, line 683: I think it will look nicer to create a local variable.

const message = event.message;

kenchris: At least in this case

kenchris (approved):
Let's merge this now as it is a mistake

PR #234, opened 2019-06-13
Editorial fixes

zolkis:


kenchris (commented):
Look at wake lock, Respec can do most of these automatically today

PR #240, opened 2019-07-16
Change NDEF{Record, Message} to interface

### LINKED ISSUE #232: NFCReadingEvent has an attribute with a type of dictionary?
http://w3c.github.io/web-nfc/#dom-nfcreadingevent defines 

`interface NFCReadingEvent : Event {
  readonly attribute NDEFMessage message;
};
`

but https://heycam.github.io/webidl/#idl-dictionaries says

> Dictionaries must not be used as the type of an attribute or constant.

Honry:
Because dictionary could not be used as the type of an attribute.

Fixed #232

### DIFF
--- a/index.html
+++ b/index.html
@@ -698,17 +698,17 @@ reader.onreading = event => {
   for (let record of message.records) {
     switch (record.recordType) {
       case "text":
-        console.log(`Text: ${record.data}`);
+        console.log(`Text: ${record.data()}`);
         break;
       case "url":
-        console.log(`URL: ${record.data}`);
+        console.log(`URL: ${record.data()}`);
         break;
       case "json":
-        console.log(`JSON: ${record.data.myProperty.toString()}`);
+        console.log(`JSON: ${record.data().myProperty.toString()}`);
         break;
       case "opaque":
         if (record.mediaType.startsWith('image/') {
-          const blob = new Blob([record.data], {type: record.mediaType});
+          const blob = new Blob([record.data()], {type: record.mediaType});
 
           const img = document.createElement("img");
           img.src = URL.createObjectURL(blob);
@@ -770,8 +770,8 @@ const reader = new NFCReader({url: document.baseURI, recordType: "json"});
 
 reader.addEventListener("reading", event => {
   for (let record of event.message.records) {
-    let article =/[aeio]/.test(record.data.title) ? "an" : "a";
-    console.log(`${record.data.name} is ${article} ${record.data.title}`);
+    let article =/[aeio]/.test(record.data().title) ? "an" : "a";
+    console.log(`${record.data().name} is ${article} ${record.data().title}`);
   }
 });
 
@@ -812,7 +812,7 @@ reader.onreading = event => {
   for (let record of event.message.records) {
     console.log("Record type:  " + record.recordType);
     console.log("MIME type:    " + record.mediaType);
-    console.log("=== data ===\n" + record.data);
+    console.log("=== data ===\n" + record.data());
   }
 };
 reader.start();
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1392,14 +1414,14 @@ <h2>The <dfn>NDEFRecordType</dfn> enum</h2>
...
+    typedef (DOMString or ArrayBuffer or NDEFMessageInit) NDEFMessageSource;
 
     [Constructor(), SecureContext, Exposed=Window]
     interface NFCWriter {
-      Promise&lt;void&gt; push(NDEFMessageSource message, optional NFCPushOptions options);
+      Promise&lt;void&gt; push(NDEFMessageSource message, optional NFCPushOptions options={});
     };
 
-    [Constructor(optional NFCReaderOptions options), SecureContext, Exposed=Window]
+    [Constructor(optional NFCReaderOptions options={}), SecureContext, Exposed=Window]
```
kenchris, line 1424: if you add ={} I don't think you need the optional any longer?

index.html
```diff
@@ -1142,67 +1142,89 @@
...
+      };
+
+      dictionary NDEFRecordInit {
         NDEFRecordType recordType;
         USVString mediaType;
         NDEFRecordData data;
       };
     </pre>
     <p>
       The <dfn>NDEFRecordData</dfn> is a union type representing data types allowed
```
kenchris, line 1202: Shouldn't you write/define that in Web IDL then?

index.html
```diff
@@ -1142,67 +1142,89 @@
...
     </p>
     <pre class="idl">
       typedef (DOMString or unrestricted double or ArrayBuffer or Dictionary) NDEFRecordData;
 
-      dictionary NDEFRecord {
+      [Constructor(NDEFRecordInit recordInit), Exposed=Window]
+      interface NDEFRecord {
+        readonly attribute NDEFRecordType recordType;
+        readonly attribute USVString mediaType;
+        NDEFRecordData data();
```
kenchris, line 1192: Wasn't the idea that we would have multiple getters depending on data type? like `json()`?

index.html
Honry: This was a new ReSpec error:
`WebIDL validation error: Optional dictionary arguments must have a default value of {}`

Do you have any solution?

index.html
kenchris: a bit like (copied from other spec)

```
[NewObject] Promise<any> json();
[NewObject] Promise<USVString> text();
[NewObject] Promise<ArrayBuffer> arrayBuffer();
```

index.html
kenchris: OK, maybe that is new, let's see what @marcoscaceres says

index.html
Honry: Eh... Sorry, I didn't get your point.  `NDEFRecordData` is already defined in [here](https://github.com/w3c/web-nfc/pull/240/files#diff-eacf331f0ffc35d4b482f1d15a887d3bR1186).

index.html
Honry: I saw you've already commented your concern on the original issue thread, let's just move to there.

Honry:
@kenchris, many thanks for your comments! @zolkis, @leonhsl, please help review it as well. Thanks!

(@leonhsl is very busy on spec implementation, so he asked me to address the issue. :D )

index.html
marcoscaceres: Yeah, you still need the optional :)

index.html
marcoscaceres: Does this json() need to be async? If no, please use `default toJSON()` instead.

kenchris: I am not sure. So non async methods doing potential conversion are called toSomething()?

marcoscaceres: Yep. And the default toJSON() has special free behavior defined WebIDL. Not sure if it’s suitable, so double check.

kenchris: We should have this in the design principles doc

zolkis (approved):
With the suggested change (`default toJSON()`) LGTM.

index.html
leonhsl: I assume the toJSON() discussion is for the solution below, right? That'd be our future improvement so can we just accept `NDEFRecordData data()` in this PR for now?

> [NewObject] Promise<any> json();
> [NewObject] Promise<USVString> text();
> [NewObject] Promise<ArrayBuffer> arrayBuffer();

kenchris: Yes

leonhsl:
lgtm, Thanks!

leonhsl (approved):

Honry:
@zolkis, @kenchris, can we land this now? :) I will address the suggested improvement for `NDEFRecordData data()` in a follow-up.

PR #241, opened 2019-07-24
Ignore abort signal if push operation can't be aborted, fixes #239.

### LINKED ISSUE #239: Should we just ignore the Abort signal and wait for push response for cannot cancel case?
In the SPEC, step 8.2 in the [push method](https://w3c.github.io/web-nfc/#the-push-method):

"If the instance has already initiated NFC data transfer, reject p with "NoModificationAllowedError" DOMException and abort these steps."

If we fail to cancel the push request,  push request may succeed or fail with a different error. If we just reject the promise with  a "NoModificationAllowedError" exception, the developer may lose detailed imformation.  So shall we just rely on the push response if the the operation cannot be canceled?

Actually, for current case(the NfcImpl.java on Android), if we fail to cancel the push request, the push response will always be called before cancel response. So "NoModificationAllowedError" exception should never occur.

DonnaWuDongxia:
@kenchris @zolkis  would you please take a look at this? In current implementation, we just rely on the response of push operation if it can't be aborted. As the abort signal will reject the same JS promise with the response to the push method, so if we reject the promise when the operation can't be aborted, we will failed to notice the developer about the real response to the push operation. So, Reilly and me suggest we just ignore the Abort signal when the push operation can't be aborted. 
This is described in #239 .

### DIFF
--- a/index.html
+++ b/index.html
@@ -242,8 +242,7 @@
     <dfn data-cite="!WEBIDL#notsupportederror"><code>NotSupportedError</code></dfn>,
     <dfn data-cite="!WEBIDL#networkerror"><code>NetworkError</code></dfn>,
     <dfn data-cite="!WEBIDL#notreadableerror"><code>NotReadableError</code></dfn>,
-    <dfn data-cite="!WEBIDL#timeouterror"><code>TimeoutError</code></dfn>,
-    <dfn data-cite="!WEBIDL#nomodificationallowederror"><code>NoModificationAllowedError</code></dfn>, and
+    <dfn data-cite="!WEBIDL#timeouterror"><code>TimeoutError</code></dfn>, and
     <dfn data-cite="!WEBIDL#notallowederror"><code>NotAllowedError</code></dfn>,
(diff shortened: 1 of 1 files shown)


zolkis:
Yes, that's the desired behavior. LGTM and thanks!
(before AbortSignal was used, the old steps made sense since canceling had its own Promise).
It's not too nice we just ignore AbortSignal with no explanation, but that's what we can do.

index.html
```diff
@@ -1894,13 +1893,11 @@ <h3>The <dfn>NFCReaderOptions</dfn> dictionary</h3>
             <a>add the following abort steps</a> to <var>signal</var>:
               <ol>
                 <li>
-                  Stop the instance's <var>timer</var> if it is active.
+                  If the instance has already initiated NFC data transfer,
+                  and it can't be aborted, just ignore the abort signal and return.
```
kenchris: I would remove "just"

DonnaWuDongxia: Done.

PR #243, opened 2019-07-29
Replace NDEFRecordData data() with multiple getters

Honry:
```
USVString toText();
[NewObject] ArrayBuffer toArrayBuffer();
[NewObject] object toJSON();
```

- Define a concept named "bytes" for NDEFRecord object
aim to replace the "record's data" in NFC reading algorithm
as which was removed from spec.
- Add algorithm for converting the NDEFRecord object's
associated bytes to a given type for above getters.
- Update examples and miscs.

### DIFF
--- a/index.html
+++ b/index.html
@@ -771,8 +771,8 @@ const reader = new NFCReader({url: document.baseURI, recordType: "json"});
 
 reader.addEventListener("reading", event => {
   for (let record of event.message.records) {
-    let json = record.toJSON();
-    let article =/[aeio]/.test(json.title) ? "an" : "a";
+    const json = record.toJSON();
+    const article =/[aeio]/.test(json.title) ? "an" : "a";
     console.log(`${json.name} is ${article} ${json.title}`);
   }
 });
@@ -1203,9 +1203,9 @@ writer.push("Pushing data is fun!", {target: "tag", ignoreRead: false});
       };
     </pre>
     <p>
-      <a>NDEFRecord</a> objects have an associated <dfn id="ndefrecord-bytes">bytes</dfn>
-      (a <a>byte sequence</a>) set on creation that represents the payload data of the
-      <a>NDEF record</a>, which is `null` if there was no data in the
+      <a>NDEFRecord</a> objects have associated <dfn id="ndefrecord-bytes">bytes</dfn>
+      (a <a>byte sequence</a>), set on creation that represents the payload data of the
+      <a>NDEF record</a>, which is `null` if there were no data in the
       <a>NDEF record</a>.
     </p>
     <p>
@@ -1224,7 +1224,7 @@ writer.push("Pushing data is fun!", {target: "tag", ignoreRead: false});
     <p>
       The <dfn data-dfn-for="NDEFRecord">toText()</dfn> method, when invoked, MUST return the
       result of running <a>convert NDEFRecord's bytes</a> with an <a>NDEFRecord</a> object
-      and a `text` type, unless there's other specific illustration in step 7 of
+      and a `text` type, unless there's another specific illustration in step 7 of
       <a href="#the-nfc-reading-algorithm">The NFC reading algorithm</a>.
     </p>
     <p>
@@ -1264,7 +1264,7 @@ writer.push("Pushing data is fun!", {target: "tag", ignoreRead: false});
           <dt>text</dt>
           <ol>
             <li>
-              If the <var>recordType</var> value is not equal to <code>"empty"</code>, then
+              If the <var>recordType</var> value is not equal to "`empty`", then
               return the result of running <a>UTF-8 decode</a> on <var>bytes</var>.
             </li>
             <li>
@@ -1274,8 +1274,8 @@ writer.push("Pushing data is fun!", {target: "tag", ignoreRead: false});
           <dt>arrayBuffer</dt>
           <ol>
             <li>
-              If the <var>recordType</var> value is equal to <code>"json"</code> or
-              <code>"opaque"</code>, then return an {{ArrayBuffer}} whose contents are the
+              If the <var>recordType</var> value is equal to "`json`" or
+              "`opaque`", then return an {{ArrayBuffer}} whose contents are the
               <var>bytes</var>. Re-throw any exceptions.
             </li>
             <li>
@@ -1285,8 +1285,8 @@ writer.push("Pushing data is fun!", {target: "tag", ignoreRead: false});
           <dt>JSON</dt>
           <ol>
             <li>
-              If the <var>recordType</var> value is equal to <code>"json"</code> or
-              <code>"opaque"</code>, then return the result of running <a>parse JSON from bytes</a>
+              If the <var>recordType</var> value is equal to "`json`" or
+              "`opaque`", then return the result of running <a>parse JSON from bytes</a>
               on <var>bytes</var>. Re-throw any exceptions.
             </li>
             <li>
--- a/index.html
+++ b/index.html
@@ -445,7 +445,7 @@
     The <dfn>Web NFC message origin</dfn> is a <a>serialized origin</a>
     with <code>"https"</code> scheme, stored in the <a>Web NFC record</a>.
     For <a>NFC content</a> that is not a <a>Web NFC message</a>, it is
-    <code>null</code>.
+    `null`.
   </p>
   <p>
     The <dfn id="web-nfc-id">Web NFC Id</dfn> is an <a>absolute-URL string</a>,
@@ -771,8 +771,9 @@ const reader = new NFCReader({url: document.baseURI, recordType: "json"});
 
 reader.addEventListener("reading", event => {
   for (let record of event.message.records) {
-    let article =/[aeio]/.test(record.toJSON().title) ? "an" : "a";
-    console.log(`${record.toJSON().name} is ${article} ${record.toJSON().title}`);
+    let json = record.toJSON();
+    let article =/[aeio]/.test(json.title) ? "an" : "a";
+    console.log(`${json.name} is ${article} ${json.title}`);
   }
 });
 
@@ -1190,9 +1191,9 @@ writer.push("Pushing data is fun!", {target: "tag", ignoreRead: false});
       interface NDEFRecord {
         readonly attribute NDEFRecordType recordType;
         readonly attribute USVString mediaType;
-        [NewObject] USVString toText();
+        USVString toText();
         [NewObject] ArrayBuffer toArrayBuffer();
-        [NewObject] any toJSON();
+        [NewObject] object toJSON();
       };
 
       dictionary NDEFRecordInit {
@@ -1204,7 +1205,7 @@ writer.push("Pushing data is fun!", {target: "tag", ignoreRead: false});
     <p>
       <a>NDEFRecord</a> objects have an associated <dfn id="ndefrecord-bytes">bytes</dfn>
       (a <a>byte sequence</a>) set on creation that represents the payload data of the
-      <a>NDEF record</a>, which is <code>null</code> if there was no data in the
+      <a>NDEF record</a>, which is `null` if there was no data in the
       <a>NDEF record</a>.
     </p>
     <p>
@@ -1212,27 +1213,29 @@ writer.push("Pushing data is fun!", {target: "tag", ignoreRead: false});
       for <a href="#dom-ndefrecordinit-data">NDEFRecordInit.data</a> property.
     </p>
     <p>
-      The <dfn>NDEFRecord.mediaType</dfn>
+      The <dfn data-dfn-for="NDEFRecord">mediaType</dfn>
       property represents the <a>MIME type</a> of the <a>NDEF record</a>
       payload.
     </p>
     <p>
-      The <dfn>NDEFRecord.recordType</dfn> property represents the
+      The <dfn data-dfn-for="NDEFRecord">recordType</dfn> property represents the
       <a>NDEF record</a> types.
     </p>
     <p>
-      The <dfn>NDEFRecord.toText()</dfn> method, when invoked, MUST return the
-      result of running <a>convert an NDEFRecord object's bytes</a> with <i>text</i>,
-      unless there's other specific illustration in step 7 of
+      The <dfn data-dfn-for="NDEFRecord">toText()</dfn> method, when invoked, MUST return the
+      result of running <a>convert NDEFRecord's bytes</a> with an <a>NDEFRecord</a> object
+      and a `text` type, unless there's other specific illustration in step 7 of
       <a href="#the-nfc-reading-algorithm">The NFC reading algorithm</a>.
     </p>
     <p>
-      The <dfn>NDEFRecord.toArrayBuffer()</dfn> method, when invoked, MUST return the
-      result of running <a>convert an NDEFRecord object's bytes</a> with <i>arrayBuffer</i>.
+      The <dfn data-dfn-for="NDEFRecord">toArrayBuffer()</dfn> method, when invoked, MUST return the
+      result of running <a>convert NDEFRecord's bytes</a> with an <a>NDEFRecord</a> object
+      and an `arrayBuffer` type.
     </p>
     <p>
-      The <dfn>NDEFRecord.toJSON()</dfn> method, when invoked, MUST return the
-      result of running <a>convert an NDEFRecord object's bytes</a> with <i>JSON</i>.
+      The <dfn data-dfn-for="NDEFRecord">toJSON()</dfn> method, when invoked, MUST return the
+      result of running <a>convert NDEFRecord's bytes</a> with an <a>NDEFRecord</a> object
+      and a `JSON` type.
     </p>
     <p>
       The <dfn>NDEFRecordInit</dfn> dictionary is used to initialize an <a>NDEF record</a>
@@ -1245,12 +1248,16 @@ writer.push("Pushing data is fun!", {target: "tag", ignoreRead: false});
       and <a>Writing or pushing content</a> sections.
     </p>
     <p>
-      To <dfn>convert an NDEFRecord object's bytes</dfn> to a given <var>type</var>, run these steps:
+      To <dfn>convert NDEFRecord's bytes</dfn>, pass a <var data-type="NDEFRecord">record</var>
+      and a <var>type</var>, run these steps:
     </p>
     <ol>
-      <li>Let <var>bytes</var> be the <a>NDEFRecord</a> object's associated <a>bytes</a>.
+      <li>
+        Let <var>bytes</var> be the <var>record</var>'s <a>bytes</a>.
       </li>
-      <li>Let <var>recordType</var> be the value of <a>NDEFRecord</a> object's recordType attribute.
+      <li>
+        Let <var data-type="NDEFRecordType">recordType</var> be the value of <var>record</var>'s
+        recordType attribute.
       </li>
       <li>Switch on <var>type</var>:
         <dl>
@@ -1258,10 +1265,10 @@ writer.push("Pushing data is fun!", {target: "tag", ignoreRead: false});
           <ol>
             <li>
               If the <var>recordType</var> value is not equal to <code>"empty"</code>, then
-              return the result of running <a>UTF-8 decode</a> on the <var>bytes</var>.
+              return the result of running <a>UTF-8 decode</a> on <var>bytes</var>.
             </li>
             <li>
-              Otherwise, return <code>null</code>.
+              Otherwise, return `null`.
             </li>
           </ol>
           <dt>arrayBuffer</dt>
@@ -1269,11 +1276,11 @@ writer.push("Pushing data is fun!", {target: "tag", ignoreRead: false});
             <li>
               If the <var>recordType</var> value is equal to <code>"json"</code> or
               <code>"opaque"</code>, then return an {{ArrayBuffer}} whose contents are the
-              <var>bytes</var>. Exceptions thrown during the creation of the {{ArrayBuffer}}
-              object are re-thrown.
+              <var>bytes</var>. Re-throw any exceptions thrown by the creation of the
+              {{ArrayBuffer}} object.
             </li>
             <li>
-              Otherwise, return <code>null</code>.
+              Otherwise, return `null`.
             </li>
           </ol>
           <dt>JSON</dt>
@@ -1284,7 +1291,7 @@ writer.push("Pushing data is fun!", {target: "tag", ignoreRead: false});
               on <var>bytes</var>. Re-throw any exceptions thrown by <a>parse JSON from bytes</a>.
             </li>
             <li>
-              Otherwise, return <code>null</code>.
+              Otherwise, return `null`.
             </li>
           </ol>
         </dl>
@@ -1368,7 +1375,7 @@ writer.push("Pushing data is fun!", {target: "tag", ignoreRead: false});
       <td><dfn>"json"</dfn></td>
       <td><a>JSON MIME type</a></td>
       <td>
-        <code>null</code> or <code>DOMString</code> or <code>Number</code> or
+        `null` or <code>DOMString</code> or <code>Number</code> or
         <code>Dictionary</code>
       </td>
       <td><a>MIME type</a> (<a>TNF</a>=2) record with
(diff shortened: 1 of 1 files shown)


Honry:
[`[Default] object toJSON()`](https://heycam.github.io/webidl/#default-tojson-operation) will covert all attributes of the interface into a JSON object, which is not suitable for `NDEFRecord` interface.

index.html
```diff
@@ -698,17 +699,17 @@
...
+        console.log(`URL: ${record.toText()}`);
         break;
       case "json":
-        console.log(`JSON: ${record.data().myProperty.toString()}`);
+        console.log(`JSON: ${record.toJSON().myProperty.toString()}`);
         break;
       case "opaque":
         if (record.mediaType.startsWith('image/') {
-          const blob = new Blob([record.data()], {type: record.mediaType});
+          const blob = new Blob([record.toArrayBuffer()], {type: record.mediaType});
```
kenchris, line 712: do you still need the [] here? you are turning an array buffer into an array and then into a blob

index.html
```diff
@@ -770,8 +771,8 @@
 
 reader.addEventListener("reading", event => {
   for (let record of event.message.records) {
-    let article =/[aeio]/.test(record.data().title) ? "an" : "a";
-    console.log(`${record.data().name} is ${article} ${record.data().title}`);
+    let article =/[aeio]/.test(record.toJSON().title) ? "an" : "a";
```
kenchris: I would factor it out here

const json = record.toJSON();

kenchris: You are generating a new object each time...

index.html
```diff
@@ -1212,9 +1221,18 @@
       <a>NDEF record</a> types.
     </p>
     <p>
-      The <dfn>NDEFRecord.data()</dfn> method returns the
-      payload data of the <a>NDEF record</a> with an appropriate [[!ECMAScript]]
-      type, which depends on the <a>MIME type</a>.
+      The <dfn>NDEFRecord.toText()</dfn> method, when invoked, MUST return the
+      result of running <a>convert an NDEFRecord object's bytes</a> with <i>text</i>,
+      unless there's other specific illustration in step 7 of
```
kenchris: Hardcoding step numbers is very dangerous... it can easily get out of sync. @marcoscaceres do you have ideas how to get around this?

index.html
```diff
@@ -1212,9 +1221,18 @@
...
+      <a href="#the-nfc-reading-algorithm">The NFC reading algorithm</a>.
+    </p>
+    <p>
+      The <dfn>NDEFRecord.toArrayBuffer()</dfn> method, when invoked, MUST return the
+      result of running <a>convert an NDEFRecord object's bytes</a> with <i>arrayBuffer</i>.
```
kenchris: "convert an NDEFRecord object's bytes" is very long. I wonder if we could just call it "convert NDEFRecord's bytes"

index.html
```diff
@@ -1226,6 +1244,52 @@
       <a href="#steps-receiving">Receiving and parsing content</a>
       and <a>Writing or pushing content</a> sections.
     </p>
+    <p>
+      To <dfn>convert an NDEFRecord object's bytes</dfn> to a given <var>type</var>, run these steps:
+    </p>
+    <ol>
+      <li>Let <var>bytes</var> be the <a>NDEFRecord</a> object's associated <a>bytes</a>.
```
kenchris: Do we really need to say object here? @marcoscaceres?

index.html
```diff
@@ -1226,6 +1244,52 @@
...
+      </li>
+      <li>Let <var>recordType</var> be the value of <a>NDEFRecord</a> object's recordType attribute.
+      </li>
+      <li>Switch on <var>type</var>:
```
kenchris, line 1259: I have never seen "switch" used in specs. We usually would say "If type is A, ..., Otherwise, if type is B,..."

@marcoscaceres sorry for needing your input so much on this :-)

index.html
```diff
@@ -1226,6 +1244,52 @@
...
+        <dl>
+          <dt>text</dt>
+          <ol>
+            <li>
+              If the <var>recordType</var> value is not equal to <code>"empty"</code>, then
+              return the result of running <a>UTF-8 decode</a> on the <var>bytes</var>.
```
kenchris: on bytes should be fine... removing "the"

index.html
```diff
@@ -1226,6 +1244,52 @@
...
+            </li>
+            <li>
+              Otherwise, return <code>null</code>.
```
kenchris: There are newer respec features making this easier to spec, like just use backtick instead of `<code>` in most cases. Check wakelock spec for examples.

Also please add yourself to the Acknowledgement section

index.html
```diff
@@ -1226,6 +1244,52 @@
...
+            </li>
+          </ol>
+          <dt>arrayBuffer</dt>
+          <ol>
+            <li>
+              If the <var>recordType</var> value is equal to <code>"json"</code> or
+              <code>"opaque"</code>, then return an {{ArrayBuffer}} whose contents are the
+              <var>bytes</var>. Exceptions thrown during the creation of the {{ArrayBuffer}}
+              object are re-thrown.
```
kenchris: s/are/should be?

index.html
```diff
@@ -1226,6 +1244,52 @@
...
+            <li>
+              Otherwise, return <code>null</code>.
+            </li>
+          </ol>
+          <dt>JSON</dt>
+          <ol>
+            <li>
+              If the <var>recordType</var> value is equal to <code>"json"</code> or
+              <code>"opaque"</code>, then return the result of running <a>parse JSON from bytes</a>
+              on <var>bytes</var>. Re-throw any exceptions thrown by <a>parse JSON from bytes</a>.
```
kenchris: Here the prose for the rethrowing is different that above. Please be consistent

index.html
```diff
@@ -1334,38 +1398,42 @@ <h2>The <dfn>NDEFRecordType</dfn> enum</h2>
       <th>NDEF record type</th>
       <th>NDEFRecord recordType</th>
       <th>NDEFRecord mediaType</th>
-      <th>NDEFRecord data</th>
+      <th>NDEFRecord data getters</th>
```
kenchris, line 1431: As these are functions, I am not sure we should call them getters. but I am unsure about this

index.html
Honry: Here's the example I referenced from: https://www.w3.org/TR/push-api/#dfn-extract-a-byte-sequence

index.html
Honry: More simplified, I like it. :)

index.html
```diff
@@ -1189,7 +1190,9 @@
       interface NDEFRecord {
         readonly attribute NDEFRecordType recordType;
         readonly attribute USVString mediaType;
-        NDEFRecordData data();
+        [NewObject] USVString toText();
```
marcoscaceres: You might want to check this, but `[NewObject]` might not apply to `USVString`... I'm not sure  tho.

index.html
Honry: Yes, we still need it, as the first argument of Blob constructor should be `optional sequence<BlobPart> blobParts`, see https://w3c.github.io/FileAPI/#dfn-Blob

index.html
```diff
@@ -1189,7 +1190,9 @@
...
+        [NewObject] ArrayBuffer toArrayBuffer();
+        [NewObject] any toJSON();
```
marcoscaceres: Can you use `[Default] object toJSON();`?  If no... can you define a dictionary? 

```suggestion
        [NewObject] object toJSON();
```

marcoscaceres: Also, there is a IDL `record` type... NDEFRecord being a "Record" seems like a there might be overlap? Or maybe "Record" is a misnomer?

index.html
marcoscaceres: (Strings are primitives, so they are always passed by value... which is why NewObject seems suspect/redundant here)

index.html
```diff
@@ -1212,9 +1221,18 @@
...
+      unless there's other specific illustration in step 7 of
+      <a href="#the-nfc-reading-algorithm">The NFC reading algorithm</a>.
+    </p>
+    <p>
+      The <dfn>NDEFRecord.toArrayBuffer()</dfn> method, when invoked, MUST return the
+      result of running <a>convert an NDEFRecord object's bytes</a> with <i>arrayBuffer</i>.
+    </p>
+    <p>
+      The <dfn>NDEFRecord.toJSON()</dfn> method, when invoked, MUST return the
+      result of running <a>convert an NDEFRecord object's bytes</a> with <i>JSON</i>.
```
marcoscaceres: `<i>JSON</I>`... italics doesn't convey any semantics? Avoid using that.

index.html
```diff
@@ -1212,9 +1221,18 @@
...
+      result of running <a>convert an NDEFRecord object's bytes</a> with <i>text</i>,
+      unless there's other specific illustration in step 7 of
+      <a href="#the-nfc-reading-algorithm">The NFC reading algorithm</a>.
+    </p>
+    <p>
+      The <dfn>NDEFRecord.toArrayBuffer()</dfn> method, when invoked, MUST return the
+      result of running <a>convert an NDEFRecord object's bytes</a> with <i>arrayBuffer</i>.
+    </p>
+    <p>
+      The <dfn>NDEFRecord.toJSON()</dfn> method, when invoked, MUST return the
```
marcoscaceres:
```suggestion
      The <dfn>toJSON()</dfn> method, when invoked, MUST return the
```

And add `data-dfn-for="NDEFRecord"` above somewhere.

index.html
marcoscaceres: I like the switch also, tbh. I've also used:

> If it happens to be the case that X is "foo"

However, your HTML might  now be invalid below... can an`ol` be a child of  `dl`? If no, then wrap the `ol` in a `dd`:

```
<dl>
  <dd><ol><li></dd>
</dl>
```

marcoscaceres: You get the idea :)

index.html
```diff
@@ -1334,38 +1398,42 @@ <h2>The <dfn>NDEFRecordType</dfn> enum</h2>
...
-      <th>NDEFRecord data</th>
+      <th>NDEFRecord data getters</th>
     </tr>
     <tr>
       <td>Empty (<a>TNF</a>=0) record</td>
       <td>"empty"</td>
       <td>""</td>
-      <td><code>null</code></td>
+      <td>
+        <a href="#dom-ndefrecord-totext">toText()</a> or
```
marcoscaceres:
```suggestion
        <a>toText()</a> or
```

index.html
```diff
@@ -1334,38 +1398,42 @@ <h2>The <dfn>NDEFRecordType</dfn> enum</h2>
...
       <th>NDEFRecord mediaType</th>
-      <th>NDEFRecord data</th>
+      <th>NDEFRecord data getters</th>
     </tr>
     <tr>
       <td>Empty (<a>TNF</a>=0) record</td>
       <td>"empty"</td>
       <td>""</td>
-      <td><code>null</code></td>
+      <td>
```
marcoscaceres, line 1437:
```suggestion
      <td data-link-for="NDEFRecord">
```

index.html
marcoscaceres: Same below... never use IDs directly... let ReSpec handle that.

index.html
```diff
@@ -1334,38 +1398,42 @@ <h2>The <dfn>NDEFRecordType</dfn> enum</h2>
...
+        <a href="#dom-ndefrecord-tojson">toJSON()</a> or
+        <a href="#dom-ndefrecord-toarraybuffer">toArrayBuffer()</a>
+      </td>
     </tr>
     <tr>
       <td>NFC Forum <a>well-known type</a> (<a>TNF</a>=1) record with type <i>Text</i></td>
       <td>"text"</td>
       <td>"text/plain"</td>
-      <td><code>DOMString</code></td>
+      <td><a href="#dom-ndefrecord-totext">toText()</a></td>
```
marcoscaceres: as above

index.html
marcoscaceres: Oh, you should change the algorithm to take a variable for the NDEFRecord. 

So:

```
To <dfn>convert an `NDEFRecord`</dfn>, pass a <var data-type="NDEFRecord">record</var>` and a <var>type</var>,  and run these steps:
```

marcoscaceres: Add the `data-type` for `<var>type</var>` too, if you can.

index.html
Honry: There're lots of `<code>` in the spec need to trim, I would suggest to make the improvement in a separate PR. 

And very appreciate for the acknowledgement!

index.html
Honry: > Can you use [Default] object toJSON();? If no... can you define a dictionary?

No, I don't think so, as it only try to convert the bytes of the `NDEFRecord` interface to `JSON` object. And the bytes is even just a concept of the `NDEFRecord` interface.

index.html
Honry: It presents the parameter passed to the `<a>convert an NDEFRecord object's bytes</a>` algorithm. So we don't need the italics?

index.html
Honry: Thank you! Lesson learned!

index.html
Honry: OK, I will remove the `[NewObject]`.

index.html
marcoscaceres: See my other comment... that should be a `<var>` or something else? Definitely no use case for `<i>` so something is not right here.

index.html
marcoscaceres: Just: ". Re-throw any exceptions."

index.html
Honry: Thanks, fixed.

index.html
Honry: Thanks, done.

index.html
Honry: Done.

index.html
Honry: Thanks, done.

index.html
Honry: Thanks, fixed.

index.html
Honry: Done.

index.html
Honry: Done.

index.html
Honry: Done.

index.html
Honry: I will use the back-tick. :)

index.html
```diff
@@ -770,8 +771,9 @@
 
 reader.addEventListener("reading", event => {
   for (let record of event.message.records) {
-    let article =/[aeio]/.test(record.data().title) ? "an" : "a";
-    console.log(`${record.data().name} is ${article} ${record.data().title}`);
+    let json = record.toJSON();
```
kenchris: if you never modify the value, use const not let

index.html
```diff
@@ -1198,23 +1202,40 @@
         NDEFRecordData data;
       };
     </pre>
+    <p>
+      <a>NDEFRecord</a> objects have an associated <dfn id="ndefrecord-bytes">bytes</dfn>
+      (a <a>byte sequence</a>) set on creation that represents the payload data of the
```
kenchris: comma before set, otherwise it sounds like >bytes set< 

Also, bytes is plural, so 'an bytes' is wrong. Just drop an, I would say.

index.html
```diff
@@ -1198,23 +1202,40 @@
...
+      <a>NDEF record</a>, which is `null` if there was no data in the
```
kenchris: data is plural, so I think it should be "were"

index.html
```diff
@@ -1198,23 +1202,40 @@
...
+      The <dfn data-dfn-for="NDEFRecord">recordType</dfn> property represents the
       <a>NDEF record</a> types.
     </p>
     <p>
-      The <dfn>NDEFRecord.data()</dfn> method returns the
-      payload data of the <a>NDEF record</a> with an appropriate [[!ECMAScript]]
-      type, which depends on the <a>MIME type</a>.
+      The <dfn data-dfn-for="NDEFRecord">toText()</dfn> method, when invoked, MUST return the
+      result of running <a>convert NDEFRecord's bytes</a> with an <a>NDEFRecord</a> object
+      and a `text` type, unless there's other specific illustration in step 7 of
```
kenchris: I don't understand "there's other specific illustration in step 7" 

other makes it sound like plural, so maybe you want to say "another"? but what is specific illustration in ? Also @marcoscaceres can we somehow refer to that instead of hardcoding 7 here?

index.html
```diff
@@ -1198,23 +1202,40 @@
...
     <p>
-      The <dfn>NDEFRecord.recordType</dfn> property represents the
+      The <dfn data-dfn-for="NDEFRecord">recordType</dfn> property represents the
       <a>NDEF record</a> types.
     </p>
     <p>
-      The <dfn>NDEFRecord.data()</dfn> method returns the
-      payload data of the <a>NDEF record</a> with an appropriate [[!ECMAScript]]
-      type, which depends on the <a>MIME type</a>.
+      The <dfn data-dfn-for="NDEFRecord">toText()</dfn> method, when invoked, MUST return the
```
marcoscaceres: Note: you can put the `data-dfn-for` on the nearest section element and ReSpec will automatically know what to do.

index.html
```diff
@@ -1226,6 +1247,55 @@
...
+            </li>
+            <li>
+              Otherwise, return `null`.
+            </li>
+          </ol>
+          <dt>arrayBuffer</dt>
+          <ol>
+            <li>
+              If the <var>recordType</var> value is equal to <code>"json"</code> or
+              <code>"opaque"</code>, then return an {{ArrayBuffer}} whose contents are the
```
kenchris: This is how we write these in the Wakelock spec using latest respec

```
"`wake-lock`"
```

index.html
```diff
@@ -1226,6 +1247,55 @@
...
+              <var>bytes</var>. Re-throw any exceptions.
+            </li>
+            <li>
+              Otherwise, return `null`.
+            </li>
+          </ol>
+          <dt>JSON</dt>
+          <ol>
+            <li>
+              If the <var>recordType</var> value is equal to <code>"json"</code> or
```
kenchris: Using latest respec you can use ` | | ` instead of `<var>` and add type info like

```
|recordType:NDEFRecordType|
```

kenchris: We can fix these in future updates

index.html
marcoscaceres: If the illustration is a figure element, give the figure an id, then you can reference the figure like: [[[#someid]]] - which will automatically link and write “figure X” for you. However, we don’t currently have a means to reference steps.

index.html
```diff
@@ -1226,6 +1247,55 @@
...
+      and a <var>type</var>, run these steps:
+    </p>
+    <ol>
+      <li>
+        Let <var>bytes</var> be the <var>record</var>'s <a>bytes</a>.
+      </li>
+      <li>
+        Let <var data-type="NDEFRecordType">recordType</var> be the value of <var>record</var>'s
+        recordType attribute.
+      </li>
```
marcoscaceres, line 1258: Nit: link recordType attribute.

index.html
marcoscaceres: Are these enum values? If so, you can link to them.

index.html
```diff
@@ -1334,68 +1404,85 @@ <h2>The <dfn>NDEFRecordType</dfn> enum</h2>
...
       <th>NDEFRecord mediaType</th>
-      <th>NDEFRecord data</th>
+      <th>NDEFRecord data getters</th>
     </tr>
     <tr>
       <td>Empty (<a>TNF</a>=0) record</td>
       <td>"empty"</td>
       <td>""</td>
-      <td><code>null</code></td>
+      <td data-link-for="NDEFRecord">
```
marcoscaceres: You can move data-link-for to the table element. ReSpec will then know what to link.

index.html
```diff
@@ -3228,7 +3312,7 @@ <h3>Receiving and parsing content</h3>
   <p>
     The editors would like to thank Jeffrey Yasskin, Anne van Kesteren,
     Anssi Kostiainen, Domenic Denicola, Daniel Ehrenberg, Jonas Sicking,
-    Don Coleman, Salvatore Iovene and Rijubrata Bhaumik for their
+    Don Coleman, Salvatore Iovene, Rijubrata Bhaumik and Wanming Lin for their
```
marcoscaceres, line 3325: Cough🤧 Wanming Lin should be co-editor cough ... 🤧... 😇

index.html
Honry: I meant this step: 
> Otherwise, set the result of invoking record's toText() method to the result of running decode on buffer with encoding set to "utf-16be".

Which is specified for `utf-16be` decode. I don't have better idea, so help need. :)

index.html
Honry: Thanks, fixed.

index.html
Honry: Done.

index.html
Honry: Done.

index.html
Honry: Sure!

index.html
Honry: Thanks, done.

index.html
kenchris: So what you are referring to is 

              <li>
                If <var>charset</var> is equal to <code>"utf-8"</code>,
                set <var>record</var>'s data to the result of running
                <a>UTF-8 decode</a> on <var>buffer</var>.
                set <var>record</var>'s associated <a>bytes</a> to <var>buffer</var>.
              </li>
              <li>
                Otherwise, set <var>record</var>'s data to the result of running
                Otherwise, set the result of invoking <var>record</var>'s
                <a href="#dom-ndefrecord-totext">toText()</a> method to the result of running
                <a>decode</a> on <var>buffer</var> with <var>encoding</var> set
                to <code>"utf-16be"</code>.
              </li>

Honry: Exactly.

kenchris: Sounds to me that the "convert NDEFRecord's bytes" algorithm should handle the case that the data is encoded as UTF-16

index.html
kenchris: We might get there :-)

index.html
```diff
@@ -287,6 +287,7 @@
     are defined in [[!DOM]].
   </p>
   <p>
+    <dfn data-cite="!ENCODING#concept-stream-push">push</dfn>,
```
marcoscaceres: Note: you don't need "!" in front of citations. ReSpec knows when something is normative or informative ("artificial intelligence" lol). 

```suggestion
    <dfn data-cite="ENCODING#concept-stream-push">push</dfn>,
```

index.html
kenchris: I understand the issue now. We loose info about the charset in the reading step.

Maybe we could do

> 2. Let charset be "utf-8" if bit 7 (most significant bit) of header is equal to the value 0, or else "utf-16be". 

> 3, Let MIME type's parameters["charset"] be charset, in the case it is not equal to "utf-8"

index.html
```diff
@@ -3012,11 +3099,11 @@ <h3>Receiving and parsing content</h3>
...
               <li>
                 If <var>charset</var> is equal to <code>"utf-8"</code>,
-                set <var>record</var>'s data to the result of running
-                <a>UTF-8 decode</a> on <var>buffer</var>.
+                set <var>record</var>'s associated <a>bytes</a> to <var>buffer</var>.
               </li>
               <li>
-                Otherwise, set <var>record</var>'s data to the result of running
+                Otherwise, set the result of invoking <var>record</var>'s
+                <a href="#dom-ndefrecord-totext">toText()</a> method to the result of running
```
kenchris: Yeah this part is really weird. You are replacing how calls to that future in the future will work. Much better to just set bytes to buffer in all cases (would simplify this algorithm) and then in the toText() method, check the mimetype for the charset and then choose the right decoding method

index.html
Honry: The header is the first byte of ndef's PAYLOAD field, and we didn't pass the header to the record's bytes. So we don't know what the charset is.

Maybe we could set the record's bytes to the whole byte sequence of ndef's PAYLOAD field if the recordType is 'text', then we could calculate the charset. Does that make sense?

kenchris: I think it does

index.html
```diff
@@ -1304,7 +1374,7 @@ <h2>The <dfn>NDEFRecordType</dfn> enum</h2>
       <td><dfn>"json"</dfn></td>
       <td><a>JSON MIME type</a></td>
       <td>
-        <code>null</code> or <code>DOMString</code> or <code>Number</code> or
+        `null` or <code>DOMString</code> or <code>Number</code> or
```
kenchris, line 1401: Shouldn't these be {{DOMString}}, {{Number}} or similar?

index.html
Honry: Nice, I will improve the rest in a follow-up.

index.html
Honry: Yes, I'm going to make a unified improvement in a follow-up.

index.html
Honry: Done, PTAL again. Thanks!:)

kenchris:
@Honry is there anywhere I can reach you on chat or so?

Honry:
Skype online: Wanming Lin

kenchris (approved):

Honry:
@kenchris, @marcoscaceres, really appreciate for your exhaustive review!

index.html
marcoscaceres: There is no such thing as Number, I think? 🤔

marcoscaceres:
No problem. Keep up the great work!

index.html
Honry: Right, it should be `UnrestrictedDouble`, I will fix it later. :)

PR #255, opened 2019-08-02
Editorial: Remove useless statement in conformance section

Honry:
- Fixed two ReSpec error BTW


Honry:
@kenchris, PTAL. Thanks!

kenchris (approved):

PR #281, opened 2019-08-07
Compress the Web NFC specific record

kenchris:
The record is now specced better and stored more efficiently
by using short records, no chunking, using a shorter type name
and not storing "https://".

It is also renamed to author record as it is more saying and
more in line with the type. Also before the name Web NFC Message
was used for something that is a record and not a message as
such.

***
<a href="https://pr-preview.s3.amazonaws.com/kenchris/web-nfc/pull/281.html" title="Last updated on Aug 7, 2019, 9:50 AM UTC (a4382ee)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/281/f6b043b...kenchris:a4382ee.html" title="Last updated on Aug 7, 2019, 9:50 AM UTC (a4382ee)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -213,6 +213,9 @@
     multiple records etc.
   </p>
   <p>
+    A <dfn>Web NFC message</dfn> consists of a sequence of <a>NDEF record</a>s.
+  </p>
+  <p>
     A generic record looks like the following:
     <ndef-record class="ndef"
       header="*,*,*,*,*,*"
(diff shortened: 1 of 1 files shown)


zolkis:
I thought "Web NFC Id" was not perfect, but more direct and more exact than saying "author". The id was used for expressing the author, but when skipping that step and referring directly to "author" is more confusing.

However, if "author" is more in line with other specs (any links?), we can use that.

kenchris:
I thought a lot and hard about it the last week, and I think it is better. Id was also confusing with each record already having an id, as well as our url being confusing with url records and the like. We have already had issues about that before but never made any change.

The record is really about who created/wrote it and for that author seems like a good choice.

I could call it 'message author id' if you find that better

kenchris:
OK, going to merge this, we can rename it later

zolkis:
OK, let's use this and see how it works. Thanks.

PR #284, opened 2019-08-07
Fixes #248

### LINKED ISSUE #248: Absolute URL TNF=3 should read from TYPE field and not PAYLOAD
image

I am not sure we are able to actually read these as Android most likely will just try opening them in Chrome. Would be good to verify. On Android these are considered "action URLs" basically URLs that the OS will try launching. I assume they could be Android app URIs as well (good to verify).

Even if we cannot read these (should probably be noted in the spec) then we should probably be able to write them.

kenchris:
URL is stored in TYPE field and not PAYLOAD field

***
<a href="https://pr-preview.s3.amazonaws.com/kenchris/web-nfc/pull/284.html" title="Last updated on Aug 7, 2019, 3:29 PM UTC (91ce700)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/284/cd40617...kenchris:91ce700.html" title="Last updated on Aug 7, 2019, 3:29 PM UTC (91ce700)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3272,7 +3272,7 @@ writer.push("Pushing data is fun!", {target: "tag", ignoreRead: false});
             If |ndef|'s <a>TNF field</a> is `3` (<a>absolute-URL record</a>), then
             set |record|'s recordType to "`url`",
             set |record|'s mediaType to "`text/plain`" and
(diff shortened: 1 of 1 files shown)


zolkis:
Right. Thanks.

riju:
:+1:

riju:
Ah, zoltan was quick :)

PR #285, opened 2019-08-08
Fixes #228

### LINKED ISSUE #228: Should number in input NDEF message be mapped to JSON type?
As Reilly pointed out in the code review[[1]] that: According to the specification[[2]] a Number should be expressed as a JSON type record. Also, since JSON cannot represent Infinity and -Infinity as much as it cannot represent NaN(NOTE-4 in 24.5.2 JSON.stringify section in [[3]]) the typedef should be updated to use "double" rather than "unrestricted double".

So here is the question: for NFCWriter::push(), if one of the input record has just a double number as its data feild, without "recordType" feild nor mediaType feild, should we map it to NDEFRecordType::text or NDEFRecordType::json?
According to "create Web NFC message " section in spec[[4]], NDEFRecordType::text will be deduced, but there is no possible number value in the "NDEFRecord data" column in data-mapping[[1]].

[1]: https://chromium-review.googlesource.com/c/chromium/src/+/1628587/6/third_party/blink/renderer/modules/nfc/nfc.cc#104
[2]: https://w3c.github.io/web-nfc/#data-mapping;
[3]: http://www.ecma-international.org/publications/files/ECMA-ST/Ecma-262.pdf;
[4]: https://w3c.github.io/web-nfc/#creating-web-nfc-message

kenchris:
The majority of WebIDL types are valid JSON, so use `any` as the type and drop special handling of unrestricted double

***
<a href="https://pr-preview.s3.amazonaws.com/kenchris/web-nfc/pull/285.html" title="Last updated on Aug 8, 2019, 11:34 AM UTC (cd34c4b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/285/a89fb55...kenchris:cd34c4b.html" title="Last updated on Aug 8, 2019, 11:34 AM UTC (cd34c4b)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -738,7 +738,7 @@ reader.onreading = event => {
         console.log(`URL: ${record.toText()}`);
         break;
       case "json":
-        console.log(`JSON: ${record.toJSON().myProperty.toString()}`);
+        console.log(`JSON: ${record.toJSON().myProperty}`);
         break;
       case "opaque":
         if (record.mediaType.startsWith('image/') {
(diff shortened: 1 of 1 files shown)


riju (approved):
LGTM.

Good to see `unrestricted double`  go and Numbers being handled as part of JSON. 
Need to check how encoding works on undefined though

zolkis:
Nice. Merging.

PR #292, opened 2019-08-12
Improve the definition of visibility and focus

### LINKED ISSUE #282: The link to 'focus' is not appropriate
The "focus" used in the spec links to https://html.spec.whatwg.org/multipage/interaction.html#focused which only applies to elements, rather than the web pages or browsing contexts.
e.g. 
>Reading an NFC tag containing other than Web NFC message, when a web page using the Web NFC API is open and in focus.

We need to find an another way to indicate the focus, maybe it could be something like:

> when a web page using the Web NFC API is open and contains an element which has [gained focus](https://html.spec.whatwg.org/multipage/interaction.html#gains-focus).

Thoughts?

Honry:
Fixed #282

***
<a href="https://pr-preview.s3.amazonaws.com/Honry/web-nfc/pull/292.html" title="Last updated on Aug 14, 2019, 9:03 AM UTC (e689cbd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/292/acdd289...Honry:e689cbd.html" title="Last updated on Aug 14, 2019, 9:03 AM UTC (e689cbd)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1112,7 +1112,7 @@ writer.push("Pushing data is fun!", {target: "tag", ignoreRead: false});
     </section>
     <section> <h4>Visible and focused document</h4>
       <p>
-        Web NFC API is only avaliable for the {{Document}} of the
+        Web NFC functionality is allowed only for the {{Document}} of the
         <a>top-level browsing context</a>, which must be <a>visible and focused</a>.
       </p>
       <p>
--- a/index.html
+++ b/index.html
@@ -869,18 +869,19 @@ writer.push("Pushing data is fun!", {target: "tag", ignoreRead: false});
     <section> <h3>Reading an <a>NFC tag</a></h3>
       <ol>
         <li>
-          Reading an <a>NFC tag</a> containing a <a>Web NFC message</a>, when a
-          web page using the Web NFC API is open and in <a data-cite="html#focused">focus</a>.
-          For instance, a web page instructs the user to tap an NFC tag, and
-          then receives information from the tag.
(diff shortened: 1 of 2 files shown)


Honry:
@kenchris, PTAL. Thanks!

kenchris:
@riju @anssiko and @marcoscaceres can you have a look as I am at the airport right now

kenchris (commented):
What does only available to mean? The API surface should be there so that people cannot fingerprint on availability, but it should just fail right

index.html
```diff
@@ -1101,38 +1102,62 @@ <h4>
...
-        with the {{Document}} contained by the
-        <a>top-level browsing context</a> using the Web NFC API must be
-        <a data-cite="page-visibility/#dom-visibilitystate-visible">visible</a>
-        and in <a data-cite="html#focused">focus</a>.
+      </p>
+    </section>
+    <section> <h4>Visible and focused document</h4>
+      <p>
+        Web NFC API is only avaliable for the {{Document}} of the
+        <a>top-level browsing context</a>, which must be <a>visible and focused</a>.
```
zolkis, line 1116: I suggest removing "only", it works without it.

Honry: Fine, or maybe we can use "NFC Content is only available for ..."

zolkis: Or just say "Web NFC functionality is allowed only for the ..."
then the algorithms will specify the steps on how that is handled.

Honry: OK, done.

PR #299, opened 2019-08-19
Change default compatibility value to `any`

beaufortfrancois:
I feel like having default compatibility value to `any` will help web developers getting started with the API as they may not know what means `nfc-forum` or `vendor`.

What do you think?

***
<a href="https://pr-preview.s3.amazonaws.com/beaufortfrancois/web-nfc/pull/299.html" title="Last updated on Aug 19, 2019, 11:41 AM UTC (36e1dfa)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/299/95ff627...beaufortfrancois:36e1dfa.html" title="Last updated on Aug 19, 2019, 11:41 AM UTC (36e1dfa)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1944,7 +1944,7 @@ writer.push("Pushing data is fun!", {target: "tag", ignoreRead: false});
         unrestricted double timeout = Infinity;
         boolean ignoreRead = true;
         AbortSignal? signal;
-        NDEFCompatibility compatibility = "nfc-forum";
+        NDEFCompatibility compatibility = "any";
       };
     </pre>
     <p>


kenchris:
On the other hand, that means that people will continue buying NFC stickers etc that won't be readable by all devices, which can be quite surprising to people

kenchris:
It might make more sense to explain this to developers when we introduce the feature.

"The NFC Forum defines a set of NFC tags that all support the NFC Data Exchange Format, aka NDEF. Apart from this there are other proprietary tag types that support NDEF, but are only readable by NFC devices from the same manufacturer as the tag. By default, Web NFC only reads and writes to standard tags that are readable/writable by all devices such as phones. If you need to interact with proprietary formats, please set compatibility to any"

beaufortfrancois:
From what I can read at https://www.shopnfc.com/en/content/7-nfc-compatibility, it looks like support for NDEF tags and Mifare UltraLight is almost the same.

I'm not sure how the **default** value will encourage people (not web developers) to buy NFC tags that are not readable by all devices.

kenchris:
Mifare is a brand. NXP Mifare Classic is only readable/writable by phone with a NXP chip. Other Mifare brands (I believe UltraLight) are actually NFC compatible

So to rephrase, Mifare Classic tags will work by default, but Mifare Classic (not NFC compatible and proprietary) won't

beaufortfrancois:
I still think* web developers want to simply interact, by default, with "any" NFC tag.
Is NXP Mifare Classic being only readable/writable by phone with a NXP chip the reason why every new web developer will have to use `compatibility: "any"`?

The more I think about it, we may want to rename `compatibility` to `technology`. WDYT?

```js
/* CURRENT */

// Will read NFC devices supported by the NFC standard.
const r = new NFCReader();

// Will read all NDEF compatible devices that the reader chip can read.
const r = new NFCReader({ compatibility: 'any' });


/* PROPOSED */

// Will read all NDEF compatible devices that the reader chip can read.
const r = new NFCReader();

// Will read NFC devices supported by the NFC standard.
const r = new NFCReader({ technology: 'nfc-forum' });

// Will read all vendor specific NFC tags that require specific reader chips.
const r = new NFCReader({ technology: 'vendor' });
```

kenchris:
> I still think* web developers want to simply interact, by default, with "any" NFC tag.

That might be, but it is a bit against the web, where we try to make solutions that work everywhere and not push proprietary formats. @reillyeon

> Is NXP Mifare Classic being only readable/writable by phone with a NXP chip the reason why every new web developer will have to use compatibility: "any"?

I don't follow. If developers buy actual NFC chips they don't don't need to worry or care, it will just work out of the box.

If they choose to buy Mifare Classic because of price or legacy systems etc, they should be aware that these tags will not be readable by all users and that might cause a lot of trouble - hense the current opt-in.

From working with NFC, regular devs have no idea that they cards or tags they bought might not actually be NFC compatible, and it is a costly affair finding out after that you have rolled out a product that certain phone users cannot interact with your cards/tags because they were proprietary in nature.

Even a bus system of a certain country that I know ended up using mifare classic public transportation cards and now that is delaying the whole transition to payment by phone.

In a way you could argue that being a web standard we should only support NFC Forum tags as that brings no tie-in to proprietary systems, but as Mifare Classic is deployed many places, we opted to add the 'compatibility' fields so that people could explicitly opt into using it, knowing all the downsides.

kenchris:
> The more I think about it, we may want to rename compatibility to technology. WDYT?

I am kind of OK with that, but I think technology refers to other things in the Android API (and maybe other native APIs) so it might be confusing - maybe you could check that

reillyeon:
The stated goal of the compatibility parameter is to make developers aware that they are taking a dependency on a non-standard technology. Changing the default to "any" defeats that goal. If the default is changed I will argue for the parameter to be removed entirely.

How about splitting the difference and renaming the parameter "compatibleTechnology"?

beaufortfrancois:
What do you mean with "splitting the difference"?

reillyeon:
Sorry for the idiom. I meant that if both "technology" and "compatible" are confusing then perhaps "compatibleTechnology" will make more sense.

beaufortfrancois:
Thanks for the idiom. I learn something new every day ;)

If the goal is to make developers aware of the technology used, how about adding the tag technology in `NFCReadingEvent` and remove it completely from `NFCScanOptions`. We already have the serial number available there.

```js
reader.scan();
reader.onreading = event => {
  switch (event.tagTechnology) {
    case 'nfc-forum': { /* ... */ };
    case 'mifare-classic': { /* ... */ };
    ...
  }
};
```

Note that since UA captures all NFC tags, regardless of their technology, it feels more accurate to not have it in the scan options.

kenchris:
I am OK with that.

Btw mifare classic is just called Mifare as tech and ultralight is actually NFC compatible

zolkis:
> Note that since UA captures all NFC tags, regardless of their technology, it feels more accurate to not have it in the scan options.

The use case for having it in the scan options was the app not having to check for every event, only for those types it is interested in. I think it makes sense to keep it in the scan options, but we can also expose the tech on the events.

One of the intents of this spec has been to be able to scan only for "Web NFC" stuff, that is, adapted for web apps to handle safely. So an app may want to be able to say 
- "I am interested in all tags the underlying platform can read"
- "I am interested in all NDEF tags"
- "I am interested only in Web NFC tags"
(further filtering being possible in the scan options).

zolkis:
I don't remember why did we discard "web-nfc" from the `NDEFCompatibility` enum?

Also, the name should be changed from NDEF since not all tags we can read are NDEF. It could be `NFCCompatibility` or `NFCFormat` or something like that.

Then, we could also change `"nfc-forum"` to `"ndef"` since it's more exact. If a _developer_ knows what NFC is, probably also knows what NDEF is, at least as much as knowing about NFC Forum.

In the current context, I agree making `"any"` the default value (for developer convenience), so the PR itself LGTM. 

However, for the record I note that `"web-nfc"` was meant to be the default value (because that's what this spec makes safe to use) and changing that would require some user input. But now that is missing.

kenchris:
web-nfc can work on both nfc-forum compatible ones and on mifare classic

kenchris:
I am leaning more and more towards just supporting mifare classic out of the box. I don't know any other legacy tag that has gained NDEF support and we probably won't see others. It is also being outfaced as people realize the issues (like the public transportation cards here are changing to Mifare Ultra from Classic).

I wish there was another way to warn, but maybe we don't need to and we just explain what the Mifare Classic is in the spec along with the other NFC types.

zolkis:
NFC Forum or MiFare classic are underlying tech: but we defined Web NFC specific content over these to be the default, for security reasons. 

We did have the early assumption that Web NFC comms where origin policies can be applied could be exempt from prompting. However, we realized that same-origin policy cannot be securely _enforced_ with NFC, so the obtaining permission would likely to be the same, regardless of the underlying tech.

But if there is legit justification for using the Web NFC content format (external record etc), IMHO it would make sense if that would be the default and I'd like developers to explicitly and consciously having to express their choice by specifying "any" or "vendor" if "web-nfc" is not enough.

beaufortfrancois:
With https://github.com/w3c/web-nfc/pull/328 now merged. I believe we can close this issue. Thanks!

PR #300, opened 2019-08-19
Replace start/stop with scan/AbortController

### LINKED ISSUE #225: Consider using AbortController rather than a stop() method
Somehow #147 became about `cancelPush()` rather than `cancelWatch()` (now `stop()`). `push()` now takes an `AbortSignal` but `NFCReader` still uses the `start()` and `stop()` pattern.

beaufortfrancois:
This is a first draft to see if I'm heading in the right direction.
Let me know what you think.

Fix #225

***
<a href="https://pr-preview.s3.amazonaws.com/beaufortfrancois/web-nfc/pull/300.html" title="Last updated on Aug 21, 2019, 8:10 AM UTC (730a28d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/300/acdd289...beaufortfrancois:730a28d.html" title="Last updated on Aug 21, 2019, 8:10 AM UTC (730a28d)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -755,7 +755,7 @@ reader.onreading = event => {
   }
 };
 
-reader.start();
+reader.scan();
     </pre>
   </aside>
 
@@ -774,7 +774,7 @@ reader.start();
     </p>
     <pre class="highlight">
 const reader = new NFCReader({url: "https://mygame.com/mypath/mygame"});
-reader.start();
+reader.scan();
 reader.onreading = async event => {
   console.log(`Source:     ${ event.message.url }`);
   console.log(`Game state: ${ JSON.stringify(event.message.records) }`);
@@ -810,7 +810,7 @@ reader.addEventListener("reading", event => {
   }
 });
 
-reader.start();
+reader.scan();
 
 const writer = new NFCWriter();
 writer.push({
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1638,8 +1638,7 @@ <h2>The <dfn>NDEFRecordType</dfn> string</h2>
       attribute EventHandler onreading;
       attribute EventHandler onerror;
 
-      void start();
-      void stop();
+      void scan(optional NFCScanOptions options={});
```
kenchris, line 1641: We probably want to move the ctor options into this

kenchris (commented):
How is the AbortError event dispatched? Shouldn't it be?

index.html
beaufortfrancois: Done.

beaufortfrancois:
@kenchris I thought it was done automatically like `Push` in the "add the following abort steps". Nope?

kenchris:
Ah maybe it is... If you could verify that would be great

zolkis:
Looks good.

beaufortfrancois:
According to https://dom.spec.whatwg.org/#abortsignal-abort-algorithms, 

> An AbortSignal object has associated abort algorithms, which is a set of algorithms which are to be executed when its aborted flag is set. 

> To signal abort, given a AbortSignal object signal, run these steps:
> 1. If signal’s aborted flag is set, then return.
> 2. Set signal’s aborted flag.
> 3. For each algorithm in signal’s abort algorithms: run algorithm.
> 4. Empty signal’s abort algorithms.
> 5. Fire an event named abort at signal.

PR #303, opened 2019-08-21
Remove focused requirement

beaufortfrancois:
As raised in https://bugs.chromium.org/p/chromium/issues/detail?id=996250, I think* we should remove the `focused` requirement for Document as it causes some confusion.

In the example for instance, an iframe with a focused "input" element causes WebNFC operations to be suspended.

***
<a href="https://pr-preview.s3.amazonaws.com/beaufortfrancois/web-nfc/pull/303.html" title="Last updated on Aug 21, 2019, 4:54 PM UTC (0d7e7fb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/303/9ded2e9...beaufortfrancois:0d7e7fb.html" title="Last updated on Aug 21, 2019, 4:54 PM UTC (0d7e7fb)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -893,17 +893,17 @@ setTimeout(() => controller.abort(), 3000);
         <li>
           Reading an <a>NFC tag</a> containing a <a>Web NFC message</a>, when the
           {{Document}} of the <a>top-level browsing context</a> using the Web NFC API
-          is <a>visible and focused</a>. For instance, a web page instructs the user
+          is <a>visible</a>. For instance, a web page instructs the user
           to tap an NFC tag, and then receives information from the tag.
         </li>
         <li>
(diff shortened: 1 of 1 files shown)


zolkis:
What we intend to mean is that the page should be the one the user is interacting with. I don't know what is the best way nowadays to express that. Visible is not enough IMHO.

kenchris:
Maybe @domenic has input

kenchris:
We agreed at TPAC that we should remove the focused requirement and rely on visibility change instead. Want to rebase @beaufortfrancois 

@anssiko @riju can you share the notes from that discussion?

beaufortfrancois:
@kenchris Does it need rebase?
image

kenchris:
Maybe not... let me merge, but maybe you should check whether our visibility spec text covers the requirements

PR #333, opened 2019-09-03
Spec toRecords() method

### LINKED ISSUE #312: Support reading embedded records: Add toRecords()
NDEF records can have embedded records, and actually the well-known Sp (Smart Poster) makes use of that.

Some example of how to read that on Android here: http://mynewandroidtutorial.blogspot.com/2015/03/nfc-android-read-ndef-tag-textlink.html

We should basically add a `FrozenArray<NDEFRecord> toRecords();` or similar.

Developers are allowed to store their own record types within a subrecord in the well-known type, with the restriction that the type name cannot start with an uppercase (lowercase or number thus is fine).

We should think a bit about how to expose these. Potentially we could create a "custom" recordType or similar, and then add a `customType` property, or we should just add it to the recordType, like "custom:userDefinedType001". We don't need to support this initially, but it is something to think about.

kenchris:
Fixes #312 

I verified this with the NFC specs and it all looks correct. This would work for properly written smart-poster records as well.

***
<a href="https://pr-preview.s3.amazonaws.com/kenchris/web-nfc/pull/333.html" title="Last updated on Sep 3, 2019, 10:37 AM UTC (b65af96)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/333/bfabe0b...kenchris:b65af96.html" title="Last updated on Sep 3, 2019, 10:37 AM UTC (b65af96)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1380,6 +1380,11 @@ setTimeout(() => controller.abort(), 3000);
       running <a>convert NDEFRecord.[[\PayloadData]] bytes</a> with an <a>NDEFRecord</a> object
       and a `JSON` type.
     </p>
+    <p>
+      The <dfn>toRecord()</dfn> method, when invoked, MUST return the result of
+      running <a>convert NDEFRecord.[[\PayloadData]] bytes</a> with an <a>NDEFRecord</a> object
+      and an `external` type.
+    </p>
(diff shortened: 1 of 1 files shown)


zolkis (approved):
Thanks Kenneth.

PR #334, opened 2019-09-03
Add sub-record example and fix idl

kenchris:
@beaufortfrancois done!

(Support for custom sub-record types will be added soon)

***
<a href="https://pr-preview.s3.amazonaws.com/kenchris/web-nfc/pull/334.html" title="Last updated on Sep 3, 2019, 1:32 PM UTC (e1fef8f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/334/92cd2b0...kenchris:e1fef8f.html" title="Last updated on Sep 3, 2019, 1:32 PM UTC (e1fef8f)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -909,6 +909,85 @@ reader.scan({ signal: controller.signal });
 setTimeout(() => controller.abort(), 3000);
     </pre>
   </aside>
+
+  <aside title="Read a external record with sub-records" class="example">
+      <p>
+        Read a custom made external record which contains sub-records.
+      </p>
+      <p>
+        Though using JSON can be quite convenient, it is not the most
+        storage efficient way of writing data. This is due to the fact
+        that it is string based and meant to be easily human readable.
+      </p>
+      <p>
+        For instance every property uses a string as a key, and values
+        such as booleans are also written out as "`true`" (4 bytes) or
+        "`false`" (5 bytes). In contract you can actually store up to
+        8 boolean values in one byte if every bit is used to represent one.
+      </p>
+      <p>
+        External type records can be used to create custom records
+        which may contain sub-records which can contain any NDEF record
+        type, plus custom types. This makes them quite good for
+        structuring data such as binary (say images) as well as text
+        and booleans in a quite size efficient format.
+      </p>
+      <p>
+        Instead of using property names as identifiers as in JSON,
+        seperate sub-records each with their own type, will be used.
+        Custom types need to be at least one letter long, meaning the
+        overhead per record can be as little as 4 bytes, which is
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -909,6 +909,85 @@ <h4>
...
+        Instead of using property names as identifiers as in JSON,
+        seperate sub-records each with their own type, will be used.
+        Custom types need to be at least one letter long, meaning the
+        overhead per record can be as little as 4 bytes, which is
+        similar to JSON, which uses 4-6 bytes per single letter property name,
+        depending on whether it is the first property or not (given , and
+        newline separators).
+      </p>
+      <p>
+        This example shows how to read an exernal record for social posts,
```
zolkis, line 945: typo: external

index.html
```diff
@@ -909,6 +909,85 @@ <h4>
 setTimeout(() => controller.abort(), 3000);
     </pre>
   </aside>
+
+  <aside title="Read a external record with sub-records" class="example">
+      <p>
```
zolkis, line 914: I contest the usefulness of this vs the added complexity.
If storage efficiency is of concern, an external type with binary (e.g. CBOR) would be a yet more efficient solution, without the arbitrary conventions made down this example.

kenchris: It is supported by NDEF and even used in standard records like smart poster.

You can of course contest using NDEF at all (and write data directly to the tag using the type specific protocols), but hey that is what we are spec'ing.

index.html
```diff
@@ -909,6 +909,85 @@ <h4>
 setTimeout(() => controller.abort(), 3000);
     </pre>
   </aside>
+
+  <aside title="Read a external record with sub-records" class="example">
```
zolkis, line 913: typo: an

kenchris: thx

index.html
zolkis: If I am reading correctly the NFC Forum Smart Poster spec, there is no word about "sub-record", only about records (of type Title, URI, Action, Icon, Size, Type). The term "sub-record" seems to be made up outside of NFC Forum, so it's an application level term from NFC point of view. Read my previous comment in that context: apps could as well choose a more efficient storage method inside the allowed NDEF records. If we expose an app-level mechanism on top of NDEF in the Web NFC spec that needs to go though the usual feature addition path: use case, issue, group discussion, PR.

kenchris: OK, it definitely embeds the records and that is also defined how to do in other NFC specs. Whether we call them sub records or embedded records, I don't care.

PR #335, opened 2019-09-03
Remove JSON comparison and fix Zolkis' comments

kenchris:
@zolkis better?

***
<a href="https://pr-preview.s3.amazonaws.com/kenchris/web-nfc/pull/335.html" title="Last updated on Sep 3, 2019, 1:58 PM UTC (2b2fe1e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/335/40d7c8c...kenchris:2b2fe1e.html" title="Last updated on Sep 3, 2019, 1:58 PM UTC (2b2fe1e)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -910,45 +910,28 @@ setTimeout(() => controller.abort(), 3000);
     </pre>
   </aside>
 
-  <aside title="Read a external record with sub-records" class="example">
-      <p>
-        Read a custom made external record which contains sub-records.
-      </p>
-      <p>
-        Though using JSON can be quite convenient, it is not the most
-        storage efficient way of writing data. This is due to the fact
-        that it is string based and meant to be easily human readable.
-      </p>
-      <p>
-        For instance every property uses a string as a key, and values
-        such as booleans are also written out as "`true`" (4 bytes) or
-        "`false`" (5 bytes). In contract you can actually store up to
-        8 boolean values in one byte if every bit is used to represent one.
-      </p>
-      <p>
-        External type records can be used to create custom records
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -910,45 +910,28 @@ <h4>
...
+      and booleans, in a quite size efficient format. Sub-records are
+      also used in the standardized smart poster record type.
+    </p>
+    <p>
+      This example shows how to read an external record for social posts,
+      which makes use of a sub-record consisting of a text record and a
+      custom "act" (action) record, just like smart posters.
+    </p>
+    <p>
+      Custom sub-record types start with a hyphen (`-`).
```
zolkis, line 931: It is not clear if this sentence/requirement comes from NFC Forum spec or other source. Reference?

kenchris: as we map types like T to text, we need another way to differentiate custom types. In well-known, they always start with lowercase or a number (no - for instance) and all standardized types start with upper case (like T and Sp).

We haven't exposed these yet, but as our web-like names like "json" "text" etc would conflict with custom ones with the same name, we need a way to differentiate. I choose - for now as that is often used on the web, like --variable, but we could choose something else like "custom:act".

@beaufortfrancois

zolkis: Right, so my point is that we are mixing NDEF terms with an own (invented in Web NFC) application level quick solution (for what?), which I consider dangerous. It's app-level thing and should stay like that. But it's fine for the context of an example, just make clear what comes from NDEF and what is the app-level part (i.e. an arbitrary example which may be done in a different way by apps).

Issue #329, opened 2019-09-02
Records may contain ID, expose that in our API


kenchris:
We should allow writing these as well as expose them using an `id` attribute (null or optional?). We also want to allow filtering for these.

This might allow us to get rid of the author records and be closer to actual NDEF specs.

PR #338, opened 2019-09-04
Allow reading and writing of the ID field

kenchris:
No filtering support yet

Also fixed most spec problems due to changed in Web IDL

***
<a href="https://pr-preview.s3.amazonaws.com/kenchris/web-nfc/pull/338.html" title="Last updated on Sep 4, 2019, 10:18 AM UTC (3fdbac7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/338/6e46b14...kenchris:3fdbac7.html" title="Last updated on Sep 4, 2019, 10:18 AM UTC (3fdbac7)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -449,7 +449,7 @@
           <a>TYPE field</a> set to "`w3.org:A`".
         </li>
         <li>
-          The payload contains the <a>message author</a> of the
+          The payload contains the <a>message identifier</a> of the
           <a>Web NFC message</a>.
         </li>
       </ul>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1787,21 +1804,27 @@ <h2>The <dfn>NDEFRecordType</dfn> string</h2>
   <pre class="idl">
     typedef (DOMString or ArrayBuffer or NDEFMessageInit) NDEFMessageSource;
 
-    [Constructor, SecureContext, Exposed=Window]
```
beaufortfrancois, line 1790: Why moving `Constructor` as part of this PR?

index.html
```diff
@@ -1336,8 +1331,10 @@ <h4>
       <a>NDEFMessage</a> interface:
     </p>
     <pre class="idl">
-      [Constructor(NDEFMessageInit messageInit), Exposed=Window]
+      [Exposed=Window]
       interface NDEFMessage {
+        constructor(NDEFMessageInit messageInit);
```
zolkis, line 1336: Just a question, is this a new canonical way to declare constructors?

index.html
riju, line 1790: what's the difference between having Constructor as an IDL attribute, compared to an explicit function in the interface. I have not seen this in chromium source code, maybe I need to git grep better.

beaufortfrancois (commented):
Can we have a JS example as well in how this could be used?

index.html
kenchris: This is the new way according to respect error. Just open the current spec and check the error which goes to a section explaining it

index.html
kenchris: Deprecated according to respect+webidl

zolkis (approved):
I am not sure the word "identifier" is better than origin or originating URI, but it's certainly better than author and not bad anyway. So I'm in :).

index.html
kenchris: Had to get rid of errors to see what was caused by me or not :-) feel free to submit a patch doing this separately and I can rebase

kenchris:
@zolkis, the NFC spec uses the term message identifier

index.html
zolkis: Right, I noticed this in other specs (WoT spec broken, too), that's why the question.

zolkis:
Right, I know about NDEF message id :), was referring to the meaning of it (which used to be author). 
Anyway, ship it.

riju:
Fixes https://github.com/w3c/web-nfc/issues/329 ?

beaufortfrancois:
@kenchris A JS example would still be nice to show how `id` could be used.

PR #340, opened 2019-09-04
Remove special Web NFC record and use id for filtering instead

kenchris:
@zolkis we should be able to remove author records after this

***
<a href="https://pr-preview.s3.amazonaws.com/kenchris/web-nfc/pull/340.html" title="Last updated on Sep 5, 2019, 8:42 AM UTC (064e965)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/340/1c4ac8c...kenchris:064e965.html" title="Last updated on Sep 5, 2019, 8:42 AM UTC (064e965)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -415,70 +415,15 @@
       The NFC Forum <dfn data-no-export="">external type record</dfn>s are for client specified
       data and must have a type name following the [[[NFC-RTD]]] standard.
     </p>
-    <p class=note>
-      Web NFC defines a special <a>external type record</a> that is not accessible
-      to client applications.
-    </p>
     <p>
       <ndef-record
         header="*,*,*,*,*,4 (EXTERNAL)"
-        content="*,*,*,EXTERNAL TYPE (eg. w3.org:A),*,*"
+        content="*,*,*,EXTERNAL TYPE (eg. w3.org:member),*,*"
         short>
       </ndef-record>
     </p>
     </section>
-    <section>
-    <h4>
-      Web NFC author records (TNF 4)
-    </h4>
-    <p>
-      When writing data with Web NFC, a special <dfn>author type record</dfn> is written
-      in addition to the other records. This record is specificed by this specification
-      and is specific to Web NFC. The record indicates that the containing
-      <a>NDEF message</a> is targeted for <a>browsing context</a>s using this API
-      and contains information useful for handling the <a>NDEF message</a> with
-      the algorithms defined in this specification.
-      The format is as follows:
-      <ul>
-        <li>
-          A short, non-chunked record with <a>IL field</a> set to `0` and
-          thus no <a>ID LENGTH field</a> and <a>ID field</a>.
-        </li>
-        <li>
-          Uses NFC Forum <a>external type record</a> with the
-          <a>TYPE field</a> set to "`w3.org:A`".
-        </li>
-        <li>
-          The payload contains the <a>message identifier</a> of the
-          <a>Web NFC message</a>.
-        </li>
-      </ul>
-      <ndef-record
-        header="*,*,0,1,0,4 (EXTERNAL)"
-        content="8,*,_,w3.org:A,_,*"
-        short>
-      </ndef-record>
-    </p>
 
-    <p>
-      This enables matching <a>Web NFC content</a> with <a>URL pattern</a>s
-      specified by {{NFCReader}}s. For <a>NDEF message</a>s that are not <a>Web NFC
-      message</a>s, the <a>message identifier</a> is `null`.
-    </p>
-    <p>
-      The term <dfn>Web NFC content</dfn> denotes all <a>Web NFC message</a>s
-      contained within an <a>NDEF message</a>, identified by the <a>message identifier</a>.
-      This version of the specification supports one <a>author type record</a> per
-      <a>NDEF message</a>.
-    </p>
-    <p class="note">
-      As part of the <a>NDEF record</a>, an <a>ID field</a> may be
-      present in each record for application specific usages. According to the
-      [[[NFC-STANDARDS]]] it contains a URL with the maximum length of 256 bytes.
-      This URL is used for identifying the <a>NDEF record</a> payload in an
-      application specific way.
-    </p>
-    </section>
     <section>
     <h4>
       Unknown type records (TNF 5)
@@ -788,7 +733,7 @@ reader.scan();
        class="example">
     <p>
       Filtering of relevant data sources can be done by the use of
-      the <a>NFCScanOptions</a>. Below we accept URL with
+      the <a>NFCScanOptions</a>. Below we accept the record identifier URL with
       "`/mypath/mygame/`" in its path from "`mygame.com`"
       domain and its subdomains. When we read the data, we immediately update
       the game progress by issueing a push with a custom NDEF data layout.
@@ -799,14 +744,14 @@ reader.scan();
     </p>
     <pre class="highlight">
 const reader = new NFCReader();
-reader.scan({ url: "https://mygame.com/mypath/mygame" });
+reader.scan({ id: "https://mygame.com/mypath/mygame" });
 reader.onreading = async event => {
   console.log(`Source:     ${ event.message.url }`);
   console.log(`Game state: ${ JSON.stringify(event.message.records) }`);
 
   const newMessage = {
-    url: "/mypath/mygame/update",
     records: [{
+      id: "/mypath/mygame/update",
       recordType: "json",
       mediaType: "application/json",
       data: { level: 3, points: 4500, lives: 3 }
@@ -1017,12 +962,12 @@ setTimeout(() => controller.abort(), 3000);
           </li>
           <li>
             Writing to an <a>NFC tag</a> which already contains a
-            <a>Web NFC message</a> with a different <a>message identifier</a>
+            <a>Web NFC message</a> with a different <a>record identifier</a>
             (i.e. overwriting a web-specific tag).
           </li>
           <li>
             Writing to an <a>NFC tag</a> which already contains a
-            <a>Web NFC message</a> with the same <a>message identifier</a>
+            <a>Web NFC message</a> with the same <a>record identifier</a>
             (i.e. updating own tag).
           </li>
           <li>
@@ -1168,7 +1113,7 @@ setTimeout(() => controller.abort(), 3000);
   <p>
     The integrity of <a>NFC content</a> SHOULD NOT be trusted when
     used for implementing security policies, for instance the authenticity of
-    <a>message identifier</a>, unless a <a>prearranged trust relationship</a> exists.
+    <a>record identifier</a>, unless a <a>prearranged trust relationship</a> exists.
   </p>
   </section>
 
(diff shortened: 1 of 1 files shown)


kenchris:
I still don't like that out filtering doesn't filter actual records but looks at all in a bulk, but that is another issue

index.html
```diff
@@ -799,14 +799,14 @@ <h4>
     </p>
     <pre class="highlight">
 const reader = new NFCReader();
-reader.scan({ url: "https://mygame.com/mypath/mygame" });
+reader.scan({ id: "https://mygame.com/mypath/mygame" });
```
zolkis, line 747: Just to understand it: `id` is an addition, and not a replacement for `url`, but in this example we test a URL as the `id`. Then, an app can specify any `id` when creating content? 
I think this is why I asked about the name, since NDEF message is an NDEF level thing, but message identifier is an app-level concept. The two need to be mapped to each other but now a casual reader may confuse them. Provided I understand the change correctly.

index.html
```diff
@@ -799,14 +799,14 @@ <h4>
...
 reader.onreading = async event => {
   console.log(`Source:     ${ event.message.url }`);
   console.log(`Game state: ${ JSON.stringify(event.message.records) }`);
 
   const newMessage = {
-    url: "/mypath/mygame/update",
     records: [{
+      id: "/mypath/mygame/update",
```
zolkis, line 754: I wonder what the NDEF message id should look like, probably the impl takes this as an input and changes it.
Originally we were giving a path here and the implementation made up the rest.
How to make that clear to devs from the syntax?

index.html
```diff
@@ -2169,7 +2176,7 @@ <h3>The <dfn>NFCScanOptions</dfn> dictionary</h3>
         title="Filter accepting only JSON content from https://www.w3.org"
         class="example highlight">
         const options = {
-          url: "https://www.w3.org/*",  // any path from the domain is accepted
+          id: "https://www.w3.org/*",  // any path from the domain is accepted
```
zolkis, line 2152: I am thinking we should continue using `url` in scan options, or otherwise we should make clear it's a pattern the `id` is tested against, but calling it `id` gives the false assumption that a full id is matched.

zolkis (commented):
I think replacing the author record was OK, but now we need to be careful:
- the internal slot can be `Id`
- for scan options I would still use `url` or `urlPattern`, or idPattern (which looks bad): `id` alone makes me want an exact id
- for creating a message I would not use `id` in the dictionary but something else that denotes the id hint the app gives to the implementation which in the end is responsible to generate the full `id`.

zolkis:
> I still don't like that out filtering doesn't filter actual records but looks at all in a bulk, but that is another issue

First the impl needs to be able to determine whether the content is interesting to the app at all.
We have mechanism for this filtering.
We do say in the dispatching algorithm that if a content type is of interest, depending on the origin policy, if the content type in any record matches the scan option, the content is presented to the app.
At the moment scan options are an OR filter and do work per record.

zolkis:
In addition to the above, we might offer convenience to filter records within the message but I'd say its usefulness depends - I would defer that feature.
The primary thing to decide is whether a page can read _any_ content from an NFC source - if yes, that content is presented to the app, e.g. as an NDEF message.

I would not like an API where you we can just cherry-pick NDEF records from a message: we'd lose context with that. I don't see it very useful per se, once the data has been read anyway by the impl.

index.html
kenchris: - "id" is an addition
- according to NDEF spec the generator (say NFCWriter) is responsible for the uniqueness of the id's meaning we enforce the same restrictions here as we do for the author record.
- the idea is to remove url and author record after this

index.html
kenchris: Documentation should help :-)

- it should be possible to add a absolute URL as long as it matches the origin of the browsing context
- relative urls will be made absolute by us

index.html
kenchris: - if you add an url, it will be an exact match
- only if you add * and the like it is more flexible - that is the same for mediaType currently.
- I don't see how url is any clearer than id (documentation will be needed anyway)

We could consider calling it idFilter and mediaTypeFilter - but that should be another GH issue

kenchris: or maybe `containsId` and `containsMediaType` - that seems to match the behavior

kenchris: Filed an issue for this

kenchris:
Can I merge this? You can file issues if you think somethings are not handled...

zolkis:
I'd prefer to have a consistent state after a PR, so if you plan to remove `url` (i.e. replace it with `id`), it should be done in this PR. Let's use the `id` assumption for now and discuss names in #341

kenchris:
OK let me remove it all then... will get big

index.html
```diff
@@ -2135,6 +2055,7 @@ <h3>The <dfn>NFCScanOptions</dfn> dictionary</h3>
       <pre class="idl">
         dictionary NFCScanOptions {
           USVString url = "";
```
zolkis: didn't you want to remove `url`?

index.html
```diff
@@ -3115,14 +3004,12 @@ <h3><dfn>Writing or pushing content</dfn></h3>
     </p>
     <p>
       Each {{NFCReader}} can filter the <a>NFC content</a> based on
-      data type, and the URL [= url/path =] of the
-      <a>browsing context</a> which has been saved to the <a>author type record</a>
-      of the <a>NFC content</a>.
+      data type, and a message identifier (URL) pattern.
     </p>
     <p>
       If you filter by URL [= url/path =], that means it will be matched against
```
zolkis: Consider reformulating this as well: perhaps remove adressing as "you", 
the sentence is not very clear to me.

index.html
kenchris: yes, I missed this one

index.html
zolkis: I mean the whole paragraph: what does it mean that all records are accepted

kenchris:
We may be able to remove the "Web NFC message" terminology

zolkis:
I was confused by message id vs record id.
The NDEF spec says the `id` is for payload, and payload is in a record, so yes it's correct, but the original NFC Forum intent was to associate an optional `id` to a _payload_.
Questions:
- should payload `id` (per record) be mandatory (to write) in Web NFC
- should we call it `payload id` instead of `message id` ? If the idea is that in a message every `id` is the same, then yes, we can use `message id`, but that would be an extra constraint over NDEF.

zolkis:
> We may be able to remove the "Web NFC message" terminology

Right, we can just use "NDEF message".

index.html
```diff
@@ -788,7 +733,7 @@ <h4>
        class="example">
     <p>
       Filtering of relevant data sources can be done by the use of
-      the <a>NFCScanOptions</a>. Below we accept URL with
+      the <a>NFCScanOptions</a>. Below we accept the message identifier URL with
```
zolkis: message or payload identifier?

index.html
```diff
@@ -1276,11 +1221,11 @@ <h4>
...
-        Pushing <a>Web NFC content</a> to an <a>NFC tag</a> does not need to
-        <a>obtain permission</a>, if the <a>message identifier host</a> of the
-        <a>author type record</a> on that <a>NFC tag</a> is equal to the
-        <a data-cite="dom#concept-host-serializer">serialized host</a>
-        of the <a>current settings object</a>'s origin.
+        Pushing an <a>NDEF message</a> to an <a>NFC tag</a> does not need to
+        <a>obtain permission</a>, if the existing <a>NDEF message</a> only
+        contains <a>NDEF record</a>s without <a>ID field</a>s, or with <a>
+        ID field</a>s matching the [=host/registrable domain=] of the
+        <a>current settings object</a>'s origin.
```
zolkis, line 1228: A tricky sentence from security pov. We'd want to encourage using the ID, not the contrary :).
I would require permission when writing a message in which the IDs per record (payload id's) are not specified.
To be security-reviewed?

kenchris:
> * should payload `id` (per record) be mandatory (to write) in Web NFC

I will say no. It is not mandatory by NFC specs and existing solutions might not be using it today nor waning to add it when moving to Web NFC

> * should we call it `payload id` instead of `message id` ?

It is more like a record ID though and a record might have it and no payload.

NFC specs call it message id, so we might at least need to clarify that in the spec, even if we call it record id.

> If the idea is that in a message every `id` is the same, then yes, we can use `message id`, but that would be an extra constraint over NDEF.

I don't see such a constrain, but in chunked records, it is only present in the first one (illegal to have in the other chunks).

Can we fix this afterward? Like file issue, then discuss and then do global rename

index.html
```diff
@@ -1291,16 +1236,6 @@ <h4>
         to the web page using the Web NFC API.
       </p>
     </section>
-    <section> <h4>Record URL host and path when pushing</h4>
```
zolkis, line 1294: what about not completely removing this, but transforming to the `id` parlance?

index.html
kenchris: I am totally OK with changing this. Please file issue and we take it from there

index.html
kenchris: Can you copy out the text and put into an issue...

It probably makes sense to write by default (but that then will require more spec changes), but it might also be an issue for some legacy systems so we might want a way to opt out (like maybe id: null)

zolkis: but now we lose this content and we should not... it would belong to this change round

kenchris: Do you mind adding that then? I have to pick up my daughter soon :-)

kenchris:
>Right, we can just use "NDEF message".

Filed issue for tracking that

index.html
kenchris: Filed issue about this

index.html
zolkis: well it's there, what we should do is not removing it... but reformulate with `id`
no need to do it now, but you have the right context. It should really be contained in this PR IMHO.

index.html
kenchris: Added a note in the spec

index.html
kenchris: OK readded

zolkis (approved):
Ship it, thanks. Will take the discussions from here to future PRs.

index.html
```diff
@@ -2457,23 +2393,26 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
                 <ol>
                   <li>
-                    Clears the <a>pending push tuple</a>.
-                  </li>
-                  <li>
-                    Removes <a>async write handlers</a>.
+                    If |p| was settled (fulfilled or rejected), then:
+                    <ol>
+                      <li>
+                        Clears the <a>pending push tuple</a>.
```
taralx, line 2402: I believe you want "Clear" and "Remove".

beaufortfrancois: @zolkis Does this suggestion make sense to you?

zolkis: I think that was the original intent of the text (assuming removal, too). For safety, will let @kenchris clarify.

zolkis: I guess the clarification is warranted, since we don't have any explicit steps for "clear the pending push tuple".

Issue #324, opened 2019-09-02
Missing descriptions on pushing "external type" records


leonhsl:
Here: https://w3c.github.io/web-nfc/#creating-web-nfc-message

@kenchris @zolkis @riju

zolkis:
We don't use external type in NFCRecortdType, see https://w3c.github.io/web-nfc/#data-mapping.

In turn, we miss/have incomplete definition for how to handle "smart-poster".

zolkis:
We don't use external type in NFCRecordType in a clear way, we have a "default" string that has to match a pattern - I don't like that.

We also miss/have incomplete definition for how to handle "smart-poster".

zolkis:
This part has changed quite a lot while I was inactive as editor, now wondering about a few other things as well. We used to have a `"web-nfc"` record type, too, a kind of special external record, also used in compatibility checks. It was one of the primary motivations of the Web NFC spec over the low level older web NFC spec. Maybe it has been discussed and dropped, I just need to dig up why.

@kenchris could you check this, handling external type and smart-poster? 
(I could do that only tomorrow).

Issue #277, opened 2019-08-05
9.10.1 The push() method doesn't mention how record data is verified


kenchris:
Like it just states that things should be of type NFCMessageInit, but there is no verification that for instance recordType isn't set to bogus values like "blah-blah". It seems pretty vague, and just trusts that people look at the data mapping table and known how to follow that, but it doesn't even refer to that today

I wonder how we implemented this

riju:
9.10.1 Step 10.14. .. ah too much nesting, basically this step ->

> Initiate data transfer to device using output as buffer, using the NFC adapter in communication range with (connected to) device.

Did you mean "overflow" errors? Like developer trying to push more than tag capacity ?
What else should platform verify?

leonhsl:
10.8 of https://w3c.github.io/web-nfc/#the-push-method states:

> Let output be the notation for the NDEF message to be created by UA, as the result of passing message to create Web NFC message. If this throws an exception, reject p with that exception and abort these steps.

And 'create Web NFC message' process verifies data of `NDEFMessageInit`, like an invalid record type as '`blablabla`'. See https://w3c.github.io/web-nfc/#creating-web-nfc-message

zolkis:
Right now the implementation (and the underlying platform) decide what to do with the input received from a web page.

I suggest we factor out from the push steps the "validate NDEFMessage" steps, eventually relying on "validate NDEFRecord" steps.

kenchris:
Please do! 👍

zolkis:
Might be still valid, reopening until confirmed.

beaufortfrancois:
@zolkis Can we close it or is this issue still valid?

zolkis:
As to the opening comment, I think the algorithms do deal with bogus input (since they react only on valid values and throw otherwise). But it would be good to have a second look at it, maybe @leonhsl could you check the "create" algorithms and close the issue if it's the case?

leonhsl:
https://w3c.github.io/web-nfc/#creating-ndef-message is verifying various user input, in this respect I think we can close this one.

PR #348, opened 2019-09-05
Consolidate the push steps

zolkis:
Add internal slots to writer.
Factor out the timer algo.
Factor out the async write handler.

Please use the pr-preview to read the rendered text, not only the diff.

***
<a href="https://pr-preview.s3.amazonaws.com/zolkis/web-nfc/pull/348.html" title="Last updated on Sep 9, 2019, 7:28 AM UTC (f910e4d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/348/064e965...zolkis:f910e4d.html" title="Last updated on Sep 9, 2019, 7:28 AM UTC (f910e4d)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1820,11 +1820,7 @@ setTimeout(() => controller.abort(), 3000);
     such as tags.
   </p>
   <p>
-    The <dfn>NFCReader</dfn> is an object used for reading data when a device,
-    such as a tag, is within the magnetic induction field.
-  </p>
-  <p>
-    A {{NFCReader}} object has the following <a data-cite=
+    An {{NFCWriter}} object has the following <a data-cite=
     "ECMASCRIPT#sec-object-internal-methods-and-internal-slots">
     internal slots</a>:
   </p>
@@ -1832,36 +1828,80 @@ setTimeout(() => controller.abort(), 3000);
     <thead>
      <tr>
       <th>Internal Slot</th>
+      <th>Initial value</th>
+      <th>Description (<em>non-normative</em>)</th>
+     </tr>
+    </thead>
+    <tbody data-link-for="NFCWriter">
+     <tr>
+      <td><dfn>[[\PushOptions]]</dfn></td>
+      <td>`null`</td>
+      <td>
+        The {{NFCPushOptions}} value for writer.
+      </td>
+     </tr>
+     <tr>
+      <td><dfn>[[\PushMessage]]</dfn></td>
+      <td>`null`</td>
+      <td>
+        The {{NDEFMessage}} to be written.
+        It is initially unset.
+      </td>
+     </tr>
+     <tr>
+      <td><dfn>[[\PushTimer]]</dfn></td>
+      <td>`null`</td>
+      <td>
+        A timer associated with the writer.
+      </td>
+     </tr>
+    </tbody>
+  </table>
+
+  <p>
+    The <dfn>NFCReader</dfn> is an object used for reading data when a device,
+    such as a tag, is within the magnetic induction field.
+  </p>
+  <p>
+    An {{NFCReader}} object has the following <a data-cite=
+    "ECMASCRIPT#sec-object-internal-methods-and-internal-slots">
+    internal slots</a>:
+  </p>
+  <table class="simple">
+    <thead>
+     <tr>
+      <th>Internal Slot</th>
+      <th>Initial value</th>
       <th>Description (<em>non-normative</em>)</th>
      </tr>
     </thead>
     <tbody data-link-for="NFCScanOptions">
      <tr>
       <td><dfn>[[\Id]]</dfn></td>
+      <td>An empty <a>string</a>.
       <td>
-        This slot holds the provided {{NFCScanOptions}}.<a>id</a> value.
-        It is initially an empty <a>string</a>.
+        The {{NFCScanOptions}}.<a>id</a> value.
       </td>
      </tr>
      <tr>
       <td><dfn>[[\RecordType]]</dfn></td>
+      <td>`undefined`</td>
       <td>
-        This slot holds the provided {{NFCScanOptions}}.<a>recordType</a> value.
-        It is initially unset.
+        The {{NFCScanOptions}}.<a>recordType</a> value.
       </td>
      </tr>
      <tr>
       <td><dfn>[[\MediaType]]</dfn></td>
+      <td>An empty <a>string</a>.
       <td>
-        This slot holds the provided {{NFCScanOptions}}.<a>mediaType</a> value.
-        It is initially an empty <a>string</a>.
+        The {{NFCScanOptions}}.<a>mediaType</a> value.
       </td>
       </tr>
       <tr>
         <td><dfn>[[\Signal]]</dfn></td>
+        <td>`undefined`</td>
         <td>
-          This slot holds the provided {{NFCScanOptions}}.<a>signal</a> value.
-          It is initially unset.
+          The {{NFCScanOptions}}.<a>signal</a> to abort the operation.
         </td>
       </tr>
     </tbody>
@@ -1887,12 +1927,14 @@ setTimeout(() => controller.abort(), 3000);
     <thead>
      <tr>
       <th>Internal Slot</th>
+      <th>Initial value</th>
       <th>Description (<em>non-normative</em>)</th>
      </tr>
     </thead>
     <tbody>
      <tr>
       <td><dfn>[[\Suspended]]</dfn></td>
+      <td>`false`</td>
       <td>
         A boolean flag indicating whether NFC functionality is
         <a href="#nfc-suspended">suspended</a> or not, initially
@@ -1901,15 +1943,17 @@ setTimeout(() => controller.abort(), 3000);
      </tr>
      <tr>
       <td><dfn>[[\ActivatedReaderList]]</dfn></td>
+      <td>empty <a>set</a></td>
       <td>
-        A <a>set</a> of {{NFCReader}} instances initially set to the empty <a>set</a>.
+        A <a>set</a> of {{NFCReader}} instances.
       </td>
      </tr>
      <tr>
       <td><dfn>[[\PendingPush]]</dfn></td>
+      <td>empty</td>
       <td>
-        A promise-writer tuple where the promise holds a pending {{Promise}} and
-        writer holds a {{NFCWriter}}. Initially the slot is empty.
+        A &lt;|promise:Promise|, |writer:NFCWriter|&gt; tuple where |promise|
+        holds a pending {{Promise}} and |writer| holds an {{NFCWriter}}.
       </td>
       </tr>
     </tbody>
@@ -1969,6 +2013,27 @@ setTimeout(() => controller.abort(), 3000);
     </p>
   </section> <!-- visibility & focus: the suspended state -->
 
+  <section><h3>Clearing the pending push tuple</h3>
+  <p>
+    To <dfn>clear the pending push tuple</dfn> on an
+    <a>environment settings object</a>, perform the following steps:
+    <ol class=algorithm>
+      <li>
+        If there is no <a>pending push tuple</a> |tuple|, abort these steps.
+      </li>
+      <li>
+        Let |writer| be the <a>pending push tuple</a>'s writer.
+      </li>
+      <li>
+        Stop |writer|.[[\PushTimer]] if active.
+      </li>
+      <li>
+        Clear |tuple|.
+      </li>
+    </ol>
+  </p>
+</section>
+
   <section><h3>Aborting pending push operation</h3>
   <p>
     To attempt to <dfn>abort a pending push operation</dfn> on an
@@ -1978,30 +2043,17 @@ setTimeout(() => controller.abort(), 3000);
         If there is no <a>pending push tuple</a> |tuple|, abort these steps.
       </li>
       <li>
-        Let |writer:NFCWriter| be |tuple|'s writer.
-      </li>
-      <li>
-        Let |p:Promise| be |tuple|'s promise.
-      </li>
-      <li>
-        Stop the |writer|'s |timer| if it is active.
-      </li>
-      <li>
-        If the |writer| has already initiated NFC data transfer,
+        If |tuple|'s writer has already initiated an ongoing NFC data transfer,
         abort these steps.
       </li>
       <li>
-        Clear the <a>pending push tuple</a>.
-      </li>
-      <li>
-        Reject |p| with an {{"AbortError"}} {{DOMException}}
+        Reject |tuple|'s promise with an {{"AbortError"}} {{DOMException}}
         and abort these steps.
+        <p class=note>
+          Rejecting the promise will clear the <a>pending push tuple</a>.
+        </p>
       </li>
     </ol>
-    <p class=note>
-      Rejecting |p| will clear the <a>pending push tuple</a> and
-      remove the <a>async write handlers</a>.
-    </p>
   </p>
   </section>
 
@@ -2193,10 +2245,6 @@ setTimeout(() => controller.abort(), 3000);
             Let |options:NFCPushOptions| be the second argument.
           </li>
           <li>
-            Let |signal:AbortSignal| be the |options|’ dictionary member
-            of the same name if present, or `null` otherwise.
-          </li>
-          <li>
             If there is no underlying <a>NFC Adapter</a>, or if a connection cannot
             be established, then reject |p| with a
             {{"NotSupportedError"}} {{DOMException}}
@@ -2209,9 +2257,13 @@ setTimeout(() => controller.abort(), 3000);
             and return |p|.
           </li>
           <li>
-            If pushing data is not supported by the underlying <a>NFC Adapter</a>, then reject
-            |p| with a {{"NotSupportedError"}}
+            If pushing data is not supported by the underlying
+            <a>NFC Adapter</a>, then reject |p| with a {{"NotSupportedError"}}
             {{DOMException}} and return |p|.
+          </li>
+          <li>
+            Let |signal:AbortSignal| be the |options|’ dictionary member
+            of the same name if present, or `null` otherwise.
           </li>
           <li>
             If |signal|’s [= AbortSignal/aborted flag =] is set, then reject |p|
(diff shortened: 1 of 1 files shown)


zolkis:
Stupid github. Hit https://stackoverflow.com/a/26986320?
Maybe need to delete my fork's gh-pages and sync from upstream.

This should be the only commit in this PR:

Fix push steps. Fixes #277 as well.
https://github.com/w3c/web-nfc/pull/348/commits/922adcddfb2a0d79e5d5f304653a8cd7e87bb2d8 

(Rebase is clean. It's just the PR view that's cluttered.)
> pick dcdaab4 Add algo for mapping external data to ndef, fixes #324
pick da6631b Improve the NDEF parsing algorithms
pick 3bc03d8 Remove compatibility and optionally support MIFARE
pick bfabe0b Set NDEFRecord#mediaType to "application/octet-stream" for a TNF_...
pick 92cd2b0 Spec toRecords() method
pick 40d7c8c Add sub-record example and fix idl
pick e452164 Remove JSON comparison and fix Zolkis' comments
pick 6e46b14 Reword example to use terminology closer to NFC specs
pick a33dbb1 Allow reading and writing of the ID field
pick 1c4ac8c Define internal slots for NFCReader object
pick 064e965 Accept filtering based on id instead of author record
pick 922adcd Fix push steps. Fixes #277 as well.

zolkis:
@kenchris @leonhsl @Honry @riju  PTAL

index.html
```diff
@@ -2209,225 +2268,249 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
+                  <a>environment settings object</a>.
                 </li>
               </ol>
           </li>
           <li>
             Run the following steps <a>in parallel</a>:
-            <ol>
+            <ul>
+              <li>
+                [=promise/React=] to |p|: if |p| was settled (fulfilled or
```
kenchris: Add <ul><li>...

index.html
```diff
@@ -2209,225 +2268,249 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
-                "`tag`", and once set to the value "`peer`";
-                otherwise run the following step once, with
-                |slot| set to the value of |target|.
-              </li>
-              <li>
-                Attempt to <a>abort a pending push operation</a>.
-              </li>
-              <li>
-                Associate |output| with |slot|.
+                [=promise/React=] to |p|: if |p| was settled (fulfilled or
```
kenchris: React here adds steps to run ASYNC when the promise changes state. That doesn't seem like what you want here

index.html
```diff
@@ -2209,225 +2268,249 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
               </ol>
           </li>
           <li>
             Run the following steps <a>in parallel</a>:
-            <ol>
+            <ul>
+              <li>
+                [=promise/React=] to |p|: if |p| was settled (fulfilled or
+                rejected), then clear the <a>pending push tuple</a> and
+                abort these steps.
```
kenchris: This is async so cannot abort these steps

index.html
```diff
@@ -2209,225 +2268,249 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
+        To <dfn>start the push timer</dfn>, run these steps:
+        <ol class=algorithm>
+          <li>
+            Let |options:NFCPushOptions| be the <a>pending push options</a>.
+          </li>
+          <p>
+            Let |p:Promise| be the <a>pending push tuple</a>'s promise.
+          </p>
+          <li>
+            Let |timeout:number| be |options|'s timeout.
```
kenchris: There is another issue about this should be limited to more than infinity

index.html
```diff
@@ -2209,225 +2268,249 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
+            </ul>
           </li>
         </ol>
       </p>
 
       <section><h3>Obtaining push permission</h3>
       <p>
-        To <dfn>obtain push permission</dfn>, run these steps:
+        To <dfn>obtain push permission</dfn> given , run these steps:
         <ol class=algorithm>
```
kenchris: Given?

index.html
kenchris: Move it outside the parallel steps just like the abort steps...

zolkis:
Addressed the comments above, except for timer checks.

index.html
```diff
@@ -2209,218 +2283,223 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
-                  Reject |p| with an {{"AbortError"}}
-                  {{DOMException}} and abort these steps.
+                  Run the <a>abort a pending push operation</a> on the
+                  <a>environment settings object</a>.
                 </li>
               </ol>
           </li>
           <li>
-            Run the following steps <a>in parallel</a>:
+            [=promise/React=] to |p|: if |p| was settled (fulfilled or
```
kenchris: I still like these using a <ul> for readability

zolkis:
Regarding the timer constraints, it's not touched in this PR, let's handle that in a new issue/PR.

index.html
```diff
@@ -1969,39 +2033,53 @@ <h2>The <dfn>NDEFRecordType</dfn> string</h2>
...
       <li>
-        Let |writer:NFCWriter| be |tuple|'s writer.
+        Stop the <a>pending push timer</a> if active.
       </li>
       <li>
-        Let |p:Promise| be |tuple|'s promise.
+        Stop the <a>pending push handler</a> if running.
+      </li>
+      <li>
+        Clear the <a>pending push tuple</a>.
```
kenchris: It's a bit weird that these internal slots is per NFCWriter object but you treat them as global... What if there are multiple NFCWriter objects

index.html
zolkis: Using what for readability? (not visible)

index.html
zolkis: Yes, they are per NFCWriter.
The named "pending ..." refer to the internal slots of the one and only "pending push tuple", as the text says. 
Alternatively they could be put in the "pending push tuple", but then it's weird to have the writer object there and some free hanging internal slots of the writer object.

kenchris: Then we should refer to the object itself and not to prose like 'pending push tuple' ie. |writer|.[[\PushTuple]] etc (don't remember what name you used)

kenchris: It doesn't seem to be per NFCWriter object as the writer is part of the tuple'.... Hard to see this on phone

index.html
```diff
@@ -1819,6 +1819,70 @@ <h2>The <dfn>NDEFRecordType</dfn> string</h2>
...
+        This slot holds the <a>pending push handler</a> function.
+        It is initially unset.
+      </td>
+     </tr>
+    </tbody>
+  </table>
+  <p>
+    The <dfn>pending push options</dfn> is the value of the
+    <a>[[\PushOptions]]</a> internal slot of the
+    <a>pending push tuple</a>'s writer.
```
kenchris: OK I see what you are doing here, but I find it more clear to always get the writer and then refer to the internal slots directly

zolkis: OK, not a big deal. Which if the following should I better do:
1. copy the the following from under the NFCWriter internal slots to under the settings object's internal slots where the other "pending ..." internal slots are defined

> The pending push options is the value of the [[PushOptions]] internal slot of the pending push tuple's writer.
The pending push message is the value of the [[PushMessage]] internal slot of the pending push tuple's writer.
The pending push timer is the value of the [[PushTimer]] internal slot of the pending push tuple's writer.
The pending push handler is the value of the [[PushHandler]] internal slot of the pending push tuple's writer.

2. Remove these definitions and refer to them as the pending tuple's `|writer|.[[PushMessage]]` - something that you tried to avoid for the settings object's internal slots, so rather made text definitions for them. But this might have been because it's lengthy to refer to the settings object.

kenchris: Yes that was done because they are globals and hard to refer too

zolkis: I infer you'd prefer option 2 from above?

index.html
```diff
@@ -1819,6 +1819,50 @@ <h2>The <dfn>NDEFRecordType</dfn> string</h2>
...
+     <tr>
+      <th>Internal Slot</th>
+      <th>Description (<em>non-normative</em>)</th>
+     </tr>
+    </thead>
+    <tbody data-link-for="NFCPushOptions">
+     <tr>
+      <td><dfn>[[\PushOptions]]</dfn></td>
+      <td>
+        This slot holds the provided {{NFCPushOptions}} value.
```
kenchris: why not follow the style used elsewhere?


Internal slot | Initial value | Description (non-normative)
[[PayloadData]] | Empty byte sequence. | A byte sequence representing the whole or a subset of the PAYLOAD field data.

kenchris: [[\PushOptions]] | `null` | The {{NFCPushOptions}} value for a pending push

"This slots hold..." seems unneeded

index.html
```diff
@@ -1969,39 +2017,56 @@ <h2>The <dfn>NDEFRecordType</dfn> string</h2>
...
+    <ol class=algorithm>
+      <li>
+        If there is no <a>pending push tuple</a> |tuple|, abort these steps.
       </li>
       <li>
         Let |p:Promise| be |tuple|'s promise.
       </li>
       <li>
-        Stop the |writer|'s |timer| if it is active.
+        Let |writer| be the <a>pending push tuple</a>'s writer.
```
kenchris: You are already referring to |tuple| above, so makes more sense to do the same here

Let |writer| be |tuple|'s writer OR {{NFCWriter}} instance

index.html
```diff
@@ -1969,39 +2017,56 @@ <h2>The <dfn>NDEFRecordType</dfn> string</h2>
...
       </li>
       <li>
-        If the |writer| has already initiated NFC data transfer,
-        abort these steps.
+        If |writer|.[[\PushHandler]] has already initiated an ongoing NFC data
+        transfer, abort these steps.
       </li>
       <li>
-        Clear the <a>pending push tuple</a>.
+        Run the <a>clear the pending push tuple</a> steps.
```
kenchris: Don't we have clean up steps associated the settlement of the promise? We reject it below so maybe this is unneeded

index.html
```diff
@@ -1969,39 +2017,56 @@ <h2>The <dfn>NDEFRecordType</dfn> string</h2>
...
       </li>
       <li>
         Reject |p| with an {{"AbortError"}} {{DOMException}}
         and abort these steps.
       </li>
     </ol>
-    <p class=note>
-      Rejecting |p| will clear the <a>pending push tuple</a> and
```
kenchris, line 2002: Yeah we even had a note about it

index.html
```diff
@@ -2209,218 +2270,232 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
+      </p>
+
+      <p id="steps-push-timer">
+        To <dfn>start the push timer</dfn>, run these steps:
+        <ol class=algorithm>
+          <p>
+            Let |p:Promise| be the <a>pending push tuple</a>'s promise.
+          </p>
+          <li>
+            Let |writer| be the <a>pending push tuple</a>'s writer.
```
kenchris: writer:NFCWriter

kenchris: I don't think we can say "'s writer" here because writer is not a known thing. 'promise' is, thought. Maybe say

Let |writer| be the pending push tuple's NFCWriter instance

index.html
```diff
@@ -2209,218 +2270,232 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
-              <li>
-                If |target| is "`any`", run the following
-                steps twice, once with |slot| set to the value
-                "`tag`", and once set to the value "`peer`";
-                otherwise run the following step once, with
-                |slot| set to the value of |target|.
-              </li>
-              <li>
-                Attempt to <a>abort a pending push operation</a>.
+                If |timeout| value is not equal to `Infinity`, abort these
```
kenchris: so it looks like we always abort here and never ends up with a valid timeout value

index.html
```diff
@@ -2209,218 +2270,232 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
+          </li>
+          <li>
+            If |timer| expires, reject |p| with {{"TimeoutError"}}
+            {{DOMException}}.
+          </li>
+        </ol>
+      </p>
+
+      <p id="steps-write-handler">
+        To <dfn>add write handler</dfn>, run these steps <a>in parallel</a>:
```
kenchris: add A writer handler

kenchris: I think we need to write this a bit like the promise/React and abort steps.

image

kenchris: We are not adding an object or value, but a set of steps, so the internal slot probably doesn't make so much sense for this.

"Whenever an NFC device comes within communication range, execute 'write algorithms'"

kenchris: Hmm, these are always the same steps, so should not be a PushHandler but just an algorithm instead.

Maybe just call it "The push algorithm"

NOTE: The push algorithm is executed whenever an NFC device comes within communication range.

 - if no tuple, abort these steps...
...

Maybe then rename "push a message algorithm:" to "request a message push algorithm" or similar

index.html
```diff
@@ -2209,218 +2270,232 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
+                        serialized host</a> of the <a>current settings object</a>'s
+                        origin, and the <a>obtain push permission</a> steps return
+                        `false`, then reject |p| with {{"NotAllowedError"}}
+                        {{DOMException}} and abort these steps.
                       </li>
-                    </ol>
+                    </ul>
+                  </li>
+                  <li>
+                    Let |output:NDEFMessage| be |writer|.[[\PushMessage]].
```
kenchris: It seems to me that today we have the (p, writer) tuple because of the async write handlers, but you are getting rid of that here, so with that we don't really need to store the writer at all.

Also all the internal slots you are adding to writer, are actually globals so make more sense on the NFC state record.

Just add "PushPromise, PushOptions, PushMessage, PushTimer" there.

Then we add react steps to the promise to clear PushOptions, PushMessage and PushTimer, so that things are always consistent when the promise settles (fulfills or rejects)

If a new push is scheduled, just check whether PushPromise exists, if it does, reject it which will synchnonously clean everything up, then add the new ones.

index.html
```diff
@@ -2209,218 +2270,232 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
+                steps.
               </li>
+            </ul>
+          </li>
+          <li>
+            Start a timer |timer| with the timeout value set to
+            |timeout| and set |writer|.[[\PushTimer]] to |timer|.
+          </li>
+          <li>
+            If |timer| expires, reject |p| with {{"TimeoutError"}}
```
kenchris: when?

index.html
```diff
@@ -1969,39 +2017,56 @@ <h2>The <dfn>NDEFRecordType</dfn> string</h2>
...
       </li>
       <li>
-        Let |writer:NFCWriter| be |tuple|'s writer.
+        Let |writer| be the <a>pending push tuple</a>'s writer.
+      </li>
+      <li>
+        Stop |writer|.[[\PushTimer]] if active.
+      </li>
+      <li>
+        Stop |writer|.[[\PushHandler]] if running.
```
kenchris: This is not possible... these steps are only executed when reading/writing and then we cannot abort it

index.html
zolkis: I copied it from NFCReader... so it follows the style used there. So we should then change both?

index.html
zolkis: These are the clean-up steps, they are called to react to conditions after NFC operations, not during them.

index.html
zolkis: OK. Maybe |tuple|'s writer, to follow the style used elsewhere

index.html
zolkis: AFAIK we don't have implicit cleanup steps after settling a promise, at least in this spec. Are they defined elsewhere? (that might explain the now deleted note about cleanup on promise rejection).

index.html
zolkis: Well, wasn't writer defined for the pending push tuple? Again, I copied the text elsewhere.
If we change style, it's OK, but then we need to modify in every similar place.

index.html
zolkis: Right, this is a copy-paste and don't check bug. Thanks for spotting.

index.html
zolkis: Either is fine for me, "if" seemed more unambiguous. "When" might imply that it _will_ happen, the only question is when.

index.html
zolkis: OK, makes sense.

index.html
zolkis: Hmm, I am fine with that. I didn't want to go that far to remove the tuple, but actually it would simplify things further.
I can do that tomorrow.

zolkis: And thanks for the review.

index.html
kenchris: That is what the [=promise/React=] does. It adds steps to be run whenever a promise resolves or rejects.

We used to use a transformed promise for that, but now WebIDL handles this

kenchris: It is much simpler

index.html
kenchris: Yes, because it was needed to reset the 'async write handlers" which is what you are cleaning up - but this really should be handled like I mentioned in other comments

index.html
kenchris: Maybe fit in separate issue then, or file bug

index.html
kenchris: If it expires...  Doesn't mention when to act.. if I know it will expire I can act immediately, which makes no sense

index.html
kenchris: You can not stop a running 'push handler' it only runs when doing nfc operations. But as said in other comments, these are steps/algorithms and not something you can attach to an internal slot

index.html
kenchris: Yeah I see now we have two styles, would be good to fix

index.html
zolkis: I've put that note back.

zolkis:
Squashed these commits so far. Review still ongoing.

kenchris:
OK, please ping me when you have incorporated my suggestions

kenchris:
This looks much better. I still think we should just make the new slots global (part of NFC state record) as there can only be one. Then that would allow us to get rid of the tuple.

We can do that in a separate PR thought

kenchris (approved):

zolkis:
We agreed on IRC to handle the tuple in a later PR.
No more feedback received, merging now.

Issue #357, opened 2019-09-25
Do we want to add support for writing sub records?


leonhsl:
As we know, a ndef record may contain a ndef message as payload, and web nfc spec already adds support for reading those sub records, i.e. `NDEFRecord#toRecords()`
Then, do we want to support writing as well?

zolkis:
Short answer: yes.

Long answer below.

The informative section 4.3.C. of the NFC Forum RTD spec says this is possible, but also that it's _defined by the application_, together with the use of ID as context or other internal referencing method. IOW they don't exclude that NDEF message format could be used in a recursive manner to describe message _trees_.

The Web NFC API being such an "application", we could make the private decision to support parsing and constructing message trees that since it may be convenient for developers.

I think we should reformulate the algo for [`toRecords()`](https://w3c.github.io/web-nfc/#dom-ndefrecord-torecords) to specify since I am not convinced that `toRecords()` should only operate recursively on external records (step 3.1. in the [convert payload](https://w3c.github.io/web-nfc/#dfn-convert-ndefrecord-payloaddata-bytes) steps). According to the RTD at least we can side any number of records and messages, only that we need to encompass them with the right payload length field into the containing record. FWIW "Sp" is well known type so the algo would not work on it (would fail in step 3.1). I understand we just piggyback-invoke the algo to cover the recursivity, but then we should either get rid of the check on TNF (which we don't want since we need to check on [proper ABNF](https://w3c.github.io/web-nfc/#dfn-external-type)), or otherwise reformulate the algo for [`toRecords()`](https://w3c.github.io/web-nfc/#dom-ndefrecord-torecords).

IMHO the API can already express defining a message tree by simply assigning `data` of [NDEFRecord](https://w3c.github.io/web-nfc/#the-ndefrecord-interface) to an [NDEFMessage](https://w3c.github.io/web-nfc/#the-ndefmessage-interface) object for instance.

However, we may want to validate that input (e.g. allow trees only on TNF=1 "Sp" and TNF=4). We could also do that in the [create Web NFC message](https://w3c.github.io/web-nfc/#dfn-create-web-nfc-message) steps and then we don't need Web IDL changes, just algorithm and implementation changes.

zolkis:
If you agree with the above (double-check for possible errors/omissions) I could make a PR with those changes.

leonhsl:
About `toRecords`(), yeah I agree we need to make our spec clarify that it's only for TNF=1 "Sp" and TNF=4.

[...]

PR #359, opened 2019-09-26
Specify how to handle message as payload 

zolkis:
Handle #357

***
<a href="https://pr-preview.s3.amazonaws.com/zolkis/web-nfc/pull/359.html" title="Last updated on Sep 30, 2019, 3:29 PM UTC (1d342d3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/359/748e682...zolkis:1d342d3.html" title="Last updated on Sep 30, 2019, 3:29 PM UTC (1d342d3)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1901,6 +1901,13 @@ setTimeout(() => controller.abort(), 3000);
     </tbody>
   </table>
 
+  <p class="note">
+    Note that the internal slots of {{NFCReader}} come from the
+    |options:NFCScanOptions| passed to <a>NFCReader.scan()</a>.
+    Therefore there is maximum one filter associated with any given
+    {{NFCReader}} object and successive invocations of <a>NFCReader.scan()</a>
+    with new |options:NFCScanOptions| will replace existing filters.
+  </p>
   <p>
     The <dfn data-dfn-for="NFCReader">onreading</dfn> is an {{EventHandler}} which is called to notify
     that new reading is available.
@@ -3251,10 +3258,6 @@ setTimeout(() => controller.abort(), 3000);
         incoming <a>NFC content</a>.
       </p>
       <p>
-        Multiple consecutive calls to the <a data-link-for="NFCReader">scan()</a>
-        method from the same <a>origin</a> create filters which are in OR relationship.
-      </p>
-      <p>
         When the <dfn>NFCReader.scan</dfn> method is invoked, the UA
         MUST run the following
         <dfn id="steps-listen">NFC listen algorithm</dfn>:
@@ -3269,7 +3272,7 @@ setTimeout(() => controller.abort(), 3000);
             [= list/For each =] |key| → |value| of |options|:
             <ol>
               <li>
-                If |key| equals "`signal`", set
+                If |key| equals "`signal`" and |value| is not `undefined`, set
                 |reader|.<a>[[\Signal]]</a> to |value|.
               </li>
               <li>
--- a/index.html
+++ b/index.html
@@ -1304,8 +1304,6 @@ setTimeout(() => controller.abort(), 3000);
       <a>NDEFRecord</a> interface:
     </p>
     <pre class="idl">
-      typedef any NDEFRecordData;
-
       [Exposed=Window]
       interface NDEFRecord {
         constructor(NDEFRecordInit recordInit);
@@ -1325,7 +1323,7 @@ setTimeout(() => controller.abort(), 3000);
         USVString mediaType;
         USVString id;
 
-        NDEFRecordData data;
+        any data;
       };
     </pre>
 
@@ -1362,10 +1360,6 @@ setTimeout(() => controller.abort(), 3000);
         </tr>
       </tbody>
     </table>
-    <p>
-      The <dfn>NDEFRecordData</dfn> is a union type representing data types allowed
-      for <a href="#dom-ndefrecordinit-data">NDEFRecordInit.data</a> property.
-    </p>
     <p>
       The <dfn>mediaType</dfn> property represents the <a>MIME type</a> of
       the <a>NDEF record</a> payload.
--- a/index.html
+++ b/index.html
@@ -918,43 +918,42 @@
 
     <section> <h3>Push a smart poster message</h3>
       <p>
-        Smart poster defined in NFC Forum Smart Poster Record Type Definition
-        specification. It is an NDEF record that contains an <a>NDEF message</a>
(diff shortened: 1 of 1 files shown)


kenchris (approved):
Would like a code example though

kenchris:
Want to merge or are you adding a code example first?

zolkis:
Wait, I will add examples (wrongly discussed in the issue, whereas I meant to comment here).

zolkis:
@kenchris it's better to review commit by commit, as much of the changes are related to making examples sub-sections.
Then, let's merge and do the rest in other PRs.

zolkis:
I am thinking to better specify smart poster for instance: now it's relying on the SP RTD spec (e.g. actions are 0, 1 and 2, with no context explained).

index.html
```diff
@@ -919,6 +919,93 @@ <h4>
   reader.scan({ recordType: "example.com:sp"});
       </pre>
     </aside>
+    <aside title="Push a smart poster message" class="example">
+      <p>
+        Smart poster defined in NFC Forum Smart Poster Record Type Definition
```
kenchris: In this example we write a Smart Poster as defined in...

index.html
```diff
@@ -919,6 +919,93 @@ <h4>
...
+        specification. It is an NDEF record that contains an <a>NDEF message</a>
+        as payload, which in turn may contain several records: a URI record that
```
kenchris: which at least contains the following records (need to check spec, maybe only URL is required)

index.html
```diff
@@ -919,6 +919,93 @@ <h4>
...
+        refers to a content, and additional optional records for that content's
+        title, icon, size, type and possible action (e.g. send SMS, make call,
+        launch browser, or save for later, or open for editing, etc).
+      </p>
+      <pre class="highlight">
+        const writer = new NFCWriter();
+        let sp = new NDEFMessage({
```
kenchris: you are not modifying so use const

index.html
```diff
@@ -919,6 +919,93 @@ <h4>
...
+          records: [
+            {
+              recordType: "url",
+              data: "https://my.org/content/19911"
+            },
+            {
+              recordType: "t", // type, a local type to Sp
```
kenchris: maybe explain what it represents? Maybe we should make a non-nominative section quickly explaining what a smart poster may contain?

kenchris: ah you do explain what it means, `type` just made it confusing.

"smart poster type, a local type to Sp"

index.html
```diff
@@ -919,6 +919,93 @@ <h4>
...
+            {
+              recordType: "act",  // action, a local type to Sp
+              data: 0  // do the action, in this case open in the browser
+            },
+          ];
+        });
+        writer.push({ records: [
+          {
+            recordType: "smart-poster",
+            data: sp.records
```
kenchris: If you are just calling `sp.records` then why do we need to create the NDEFMessage object and not just an array of json records?

index.html
```diff
@@ -919,6 +919,93 @@ <h4>
...
+      </pre>
+    </aside>
+    <aside title="Push an external record with an NDEF message as payload"
+           class="example">
+      <p>
+        External type records can be used to create application defined records
+        that may even contain an <a>NDEF message</a> as payload.
+      </p>
+      <p>
+        This example show how to do the previous Smart Poster use case with
```
kenchris: I find it confusing to refer to Smart Poster here as you are not storing anythign like a smart poster

kenchris:
> I am thinking to better specify smart poster for instance: now it's relying on the SP RTD spec (e.g. actions are 0, 1 and 2, with no context explained).

Not sure what you mean.

index.html
```diff
@@ -624,113 +624,111 @@ <h4>
...
-const writer = new NFCWriter();
-writer.push(
-  "Hello World"
-).then(() => {
-  console.log("Message pushed.");
-}).catch(error => {
-  console.log(`Push failed :-( try again: ${error}.`);
-});
+    <pre class="example">
+      const writer = new NFCWriter();
```
kenchris, line 634: Doesn't this extra indentation add extra indentation to the spec?

kenchris:
> > I am thinking to better specify smart poster for instance: now it's relying on the SP RTD spec (e.g. actions are 0, 1 and 2, with no context explained).
> 
> Not sure what you mean.

I don't want to create a special smart poster object or so, because they can contain other records, like the Signature spec adds a new record to it. But we could specify it but maybe as that doesn't require any implementation it is basically just a non-nominative informative section. That is what I was suggesting

index.html
zolkis: No, I compared it with the original, no difference at the rendered doc.

index.html
zolkis: I thought is was more clear as an example. You could have that Sp message from other source as well (e.g. previously read). But no problem, I can compact the example.

index.html
zolkis: OK, I can skip that remark. What I meant is that you can share the same content in 2 ways: with the standardized Sp, or in an app-specific way using external records and embedded NDEF message.

index.html
zolkis: Yes, only URL is mandatory. We might want to standardize Sp in Web NFC indeed.

kenchris: More add an informal section describing it (as we were standardizing it, more or less)

zolkis: Let's handle the Sp in a next PR. I made the suggested changes, please re-check.

index.html
```diff
@@ -871,54 +868,132 @@ <h4>
...
+        writer.push({ records: [
+          {
+            recordType: "smart-poster",
+            data: { records: [
+              {
+                recordType: "url",
+                data: "https://my.org/content/19911"
+              },
+              {
+                recordType: "t", // smart poster type, a local type to Sp
```
kenchris, line 941: How are these local types exposed when reading the records? `smart-poster:t` ?

Anyway we can look at that in follow ups

kenchris (approved):

index.html
zolkis: According to the Sp RTD it's just the "t" for title and "act" for action, then the regular tnf=1 text, tnf=1 uri, and tnf=2 (mime type).

zolkis:
Merging for now and let's take the Sp informative section, plus other changes in next PRs.

index.html
kenchris: yes but these are local, so that t is restricted to Sp records, but I guess that is fine exposing it like that

PR #360, opened 2019-10-01
Fix #282 - visibility change

### LINKED ISSUE #282: The link to 'focus' is not appropriate

kenchris:
@zolkis

***
<a href="https://pr-preview.s3.amazonaws.com/kenchris/web-nfc/pull/360.html" title="Last updated on Oct 1, 2019, 8:15 AM UTC (5b143bc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/360/b084aa5...kenchris:5b143bc.html" title="Last updated on Oct 1, 2019, 8:15 AM UTC (5b143bc)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1251,21 +1251,9 @@
     <section> <h4>Visible document</h4>
       <p>
         Web NFC functionality is allowed only for the {{Document}} of the
-        <a>top-level browsing context</a>, which must be <a>visible</a>.
+        <a>top-level browsing context</a>, which must be
+        <dfn data-cite="PAGE-VISIBILITY#dom-visibilitystate-visible">visible</dfn>.
       </p>
-      <p>
-        To determine if a given |document:Document| is <dfn>visible</dfn>
-        the user agent MUST run the following steps:
-      </p>
-      <ol class=algorithm>
-        <li>
-          If |document|'s <a href="http://w3c.github.io/page-visibility/#page-visibility#dom-visibilitystate">visibilityState</a>
-          is "hidden", return false.
-        </li>
-        <li>
-          Return true.
-        </li>
-      </ol>
       <p>
         This also means that UAs should block access to the NFC radio if
         the display is off or the device is locked.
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2053,34 +2041,33 @@ <h2>The <dfn>NDEFRecordType</dfn> string</h2>
...
+      responsible document=] of the <a>current settings object</a>.
+      </li>
+      <li>
+        If |document|'s <a>visibility state</a> is `"visible"`,
+        <a>resume NFC</a> and abort these steps.
+      </li>
+      <li>
+        Otherwise, <a>suspend NFC</a> and attempt to <a>abort a pending
+        push operation</a>.
+      </li>
```
zolkis, line 2062: I wonder wouldn't be enough to suspend NFC, meaning that pending push will not fulfill even on taps? Imagine a visibility change for any reason and the user needs to set up another push. Just wondering should we keep the pending pushes and set up readers and only suspend NFC while they are not visible, instead of resetting the NFC related operations.

kenchris: We should not. It would be very weird to have backgrounded an app and 5 min you foreground it again and put the phone down on a tag and it pushes. I think aborting makes sense and also that way the app can handle it

zolkis: When foregrounding, it could list those current ops (that have not timeouted) and user has a choice to tap or not, or to delete them.

But maybe apps can handle these themselves. However, any system notification or anything that causes visibility change will reset the user flow, and that's intrusive.

kenchris: notifications shouldn't change visibility unless you actually click on them and they focus another app. This is fine. @riju

kenchris: We really dont want to implement such a thing, that is very complicated and requires a lot of UX discussions on Android and is basically a no-go

zolkis: OK, let's keep it simple and if any doubts, discuss in an issue.

zolkis (approved):
Thanks!

PR #372, opened 2019-10-14
Don't restrict toArrayBuffer to json and opaque records

### LINKED ISSUE #371: Why returning null in arrayBuffer() from text, URL, and external record type?
Spec currently says for `arrayBuffer()`:
> 1. If the recordType value is equal to "json" or "opaque", then return an ArrayBuffer whose contents are the bytes. Re-throw any exceptions.
> 2. Otherwise, return null.

I'm not sure to understand why we should return `null` for text, url, and external record types. I think `arrayBuffer()`should always return raw bytes and let web developers handle those as they're explicitly asking for it.

beaufortfrancois:
If https://github.com/w3c/web-nfc/issues/371 gets accepted, this PR fixes https://github.com/w3c/web-nfc/issues/371

***
<a href="https://pr-preview.s3.amazonaws.com/beaufortfrancois/web-nfc/pull/372.html" title="Last updated on Oct 14, 2019, 9:45 AM UTC (775f896)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/372/6818b06...beaufortfrancois:775f896.html" title="Last updated on Oct 14, 2019, 9:45 AM UTC (775f896)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1770,12 +1770,8 @@
           <dt>arrayBuffer</dt>
           <ol>
             <li>
-              If the |recordType| value is equal to "`json`" or "`opaque`", then
-              return an {{ArrayBuffer}} whose contents are the
-              |bytes|. Re-[= exception/throw =] any exceptions.
-            </li>
-            <li>
-              Otherwise, return `null`.
+              Return an {{ArrayBuffer}} whose contents are the |bytes|.
+              Re-[= exception/throw =] any exceptions.
             </li>
           </ol>
           <dt>JSON</dt>


index.html
```diff
@@ -1770,12 +1770,8 @@ <h4>
...
           <ol>
             <li>
-              If the |recordType| value is equal to "`json`" or "`opaque`", then
-              return an {{ArrayBuffer}} whose contents are the
-              |bytes|. Re-[= exception/throw =] any exceptions.
-            </li>
-            <li>
-              Otherwise, return `null`.
+              Return an {{ArrayBuffer}} whose contents are the |bytes|.
+              Re-[= exception/throw =] any exceptions.
```
kenchris, line 1774: Can this actually throw?

kenchris (approved):

index.html
beaufortfrancois: Do you mean "Throw any exceptions" instead of "Re-throw any exceptions"?

beaufortfrancois: In https://w3c.github.io/push-api/#widl-PushMessageData-arrayBuffer-ArrayBuffer, it says:
> The arrayBuffer() method, when invoked, MUST return an ArrayBuffer whose contents are bytes. Exceptions thrown during the creation of the ArrayBuffer object are re-thrown.

PR #373, opened 2019-10-14
Improve data mapping

zolkis:
@kenchris @leonhsl review any time

***
<a href="https://pr-preview.s3.amazonaws.com/zolkis/web-nfc/pull/373.html" title="Last updated on Oct 16, 2019, 4:08 PM UTC (fc02015)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/373/8cbf2bf...zolkis:fc02015.html" title="Last updated on Oct 16, 2019, 4:08 PM UTC (fc02015)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2987,7 +2987,8 @@
                   </ul>
                 </dl>
                 <li>
-                  If |message|'s |id:string| is not `null`:
+                  If |message| is of type {{NDEFMessageInit}} and its
+                  |id:string| is not `null`:
                   <ul>
                     <li>
                       Set |identifier| to the result of invoking
--- a/index.html
+++ b/index.html
@@ -2485,7 +2485,7 @@
     <p>
       When the value of the <dfn>overwrite</dfn> property is
       `false`, the <a href="#steps-push">push algorithm</a>
-      will read the <a>NFC tag</a> regardless of the <a>ignoreRead</a> value
+      will read the <a>NFC tag</a> regardless of the `ignoreRead` value
       to determine if it has <a>NDEF</a> records on it, and if yes, it will not
       execute any pending push.
     </p>
@@ -2982,23 +2982,32 @@
                   <dt>"`media`"</dt>
                   <ul>
                     <li>
-                      If |record|'s data is {{NDEFMessageInit}}, then return
-                      the result of running the <a>create NDEF message</a>
-                      given |record|'s data.
-                    </li>
-                    <li>
-                      Otherwise, <a>map binary data to NDEF</a>.
+                      <a>map binary data to NDEF</a>.
                     </li>
                   </ul>
                   <dt><a>external type</a></dt>
                   <ul>
+                    If |record|'s recordType is an <a>external type</a>,
                     <li>
                       If |record|'s data is of type {{NDEFMessageInit}}, then
                       return the result of running the
                       <a>create NDEF message</a> given |record|'s data.
                     </li>
                     <li>
-                      Otherwise, <a>map external data to NDEF</a>
+                      Otherwise, <a>map external data to NDEF</a> given |record|.
+                    </li>
+                  </ul>
+                  <dt><a>local type</a></dt>
+                  <ul>
+                    If |record|'s recordType is a <a>local type</a> and if
+                    |record| is a payload to another <a>NDEF</a> record,
+                    <li>
+                      If |record|'s data is of type {{NDEFMessageInit}}, then
+                      return the result of running the
+                      <a>create NDEF message</a> given |record|'s data.
+                    </li>
+                    <li>
+                      Otherwise, <a>map local type to NDEF</a> given |record|.
                     </li>
                   </ul>
                   <dt>unmatched type</dt>
@@ -3453,6 +3462,56 @@
               </li>
               <li>
                 If |length| > `0`, set the |ndefRecord|'s <a>PAYLOAD field</a> to |data|.
+              </li>
+            </ol>
+          </li>
+          <li>
+            Return |ndefRecord|.
+          </li>
+        </ol>
+      </p>
+      </section>
+
+      <section><h3>Mapping local type to NDEF</h3>
+      <p>
+        To <dfn>map local type to NDEF</dfn> given a |record:NDEFRecordInit|,
+        run these steps:
+        <ol class=algorithm>
+          <li>
+            If the type of a |record|'s data is not an
+            {{ArrayBuffer}}, [= exception/throw =] a {{TypeError}}
+            and abort these steps.
+          </li>
+          <li>
+            Set |arrayBuffer| to |record|'s data.
+          </li>
+          <li>
+            Set |length:unsigned long| to |arrayBuffer|.[[\ArrayBufferByteLength]].
+          </li>
+          <li>
+            Set |data:byte sequence| to |arrayBuffer|.[[\ArrayBufferData]].
+          </li>
+          <li>
+            Let |ndefRecord| be the notation for the <a>NDEF record</a> to
+            be created by the UA.
+            <ol>
+              <li>
+                Set the |ndefRecord|'s <a>TYPE field</a> to |record|'s recordType.
+              </li>
+              <li>
+                If |record|'s recordType is a standard local type to
+                <a>smart poster</a> or a handover record, then
+                set |ndefRecord|'s <a>TNF field</a> to `1`
+                (<a>well-known type</a> record), otherwise
+                set |ndefRecord|'s <a>TNF field</a> to `4`
+                (<a>external type record</a>).
+              </li>
+              <li>
+                Set the |ndefRecord|'s <a>PAYLOAD LENGTH field</a> to |length|.
+              </li>
+              <li>
+                If |length| > `0`, set the |ndefRecord|'s <a>PAYLOAD field</a>
+                to |data|.
               </li>
             </ol>
           </li>
--- a/index.html
+++ b/index.html
@@ -167,7 +167,7 @@
       <li>Card emulation
        <ol>
         <li>
-          With a secure element: payments by holding your phone close to a
+          With a secure element: make payments by holding a phone close to a
           point-of-sales terminal.
         </li>
         <li>
@@ -1061,7 +1061,7 @@
   <section><h3>Push a text string to a peer device</h3>
     <p>
       It is possible to restrict to which devices (tags or peers) data
-      should be pushed. Below we specify to only push to peers,
+      should be pushed. Below push is specified only to peers,
       and thus, no data is pushed when the user taps a tag.
     </p>
     <pre class="example">
@@ -1094,7 +1094,8 @@
 
   <section> <h3>Read data from tag, and write to empty ones</h3>
     <p>
-      Below we read various different kinds of data which can be stored on a tag.
+      This example shows reading various different kinds of data which can be
+       stored on a tag.
       In the case the tag is empty, we write a text message with the value
       "Hello World".
     </p>
@@ -1994,7 +1995,7 @@
   </p>
   <table class="simple" data-link-for="NDEFRecord">
     <tr>
-      <th>NDEF record<br>TNF and TYPE</th>
+      <th>NDEF record <a>TNF field</a><br> and <a>TYPE field</a></th>
       <th>NDEFRecord<br> recordType</th>
       <th>NDEFRecord<br> mediaType</th>
       <th>NDEFRecord<br> data getters</th>
@@ -3868,7 +3869,8 @@
                 Let |messageEnd:boolean| (<a>ME field</a>) be bit 6 of |header|.
               </li>
               <div class=note>
-                As chunked records are not allowed as sub records, we ignore bit 5 (<a>CF field</a>).
+                As chunked records are not allowed as sub records, ignore bit 5
+                (<a>CF field</a>) is ignored.
               </div>
               <li>
                 Let |shortRecord:boolean| (<a>SR field</a>) be bit 4 of |header|.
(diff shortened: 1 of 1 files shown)


zolkis:
Next commit: simplify data mapping as described [here](https://github.com/w3c/web-nfc/issues/357#issuecomment-541565806).

zolkis:
Next commit: update data mapping algorithms, but please review first the [section 9.4 (Data mapping) on this PR](https://pr-preview.s3.amazonaws.com/zolkis/web-nfc/pull/373.html#data-mapping).

zolkis:
One question is how to read `"absolute-url"`. Do we follow the NFC specs by exposing the URL [also] on `mediaType`, or only with data getters? Then, IMHO we should both `text()` and `arrayBuffer()` getters should work.

To stay consistent when pushing obsulte URI, we don't use `mediaType` at all (currently its `"text/plain"`).

Note the addition of local types. They are tricky, will need to add checking for parent conditions to the algorithms (same ID, embedded message, etc).

zolkis:
Also, IMHO we should allow `json()` and `arrayBuffer()` getters on text and URL records.

index.html
```diff
@@ -1884,50 +1909,72 @@ <h2>The <dfn>NDEFRecordType</dfn> string</h2>
...
       <td><i>unused</i></td>
       <td>{{NDEFMessage}}</td>
-      <td><a>Well-known type</a> record with type "`Sp`"</td>
+      <td><a>Well-known type</a> (TNF=1) record with TYPE="`Sp`"</td>
+    </tr>
+    <tr>
+      <td><dfn>"`absolute-url`"</dfn></td>
+      <td><i>unused</i></td>
+      <td>{{DOMString}} <i>url</i></td>
+      <td><a>Absolute-URL record</a> (TNF=2) with TYPE=<i>url</i></td>
```
leonhsl: TNF=3?

index.html
```diff
@@ -1884,50 +1909,72 @@ <h2>The <dfn>NDEFRecordType</dfn> string</h2>
...
+      <td><a>MIME type record</a> (TNF=2) with TYPE= [= JSON type =].
       </td>
     </tr>
     <tr>
       <td><dfn>"`opaque`"</dfn></td>
-      <td><a>MIME type</a></td>
+      <td>[= MIME type =]</a></td>
       <td>{{ArrayBuffer}}</td>
-      <td><a>MIME type record</a></td>
+      <td><a>MIME type record</a> (TNF=3) with TYPE= [= MIME type =]</td>
```
leonhsl: TNF=2?

index.html
```diff
@@ -1884,50 +1909,72 @@ <h2>The <dfn>NDEFRecordType</dfn> string</h2>
...
     </tr>
     <tr>
-      <td><a>external type</a></td>
+      <td><i>[= external type =]</i></td>
+      <td><i>unused</i></td>
+      <td>
+        {{ArrayBuffer}} or any<br>
+        <a>typed array type</a> or<br>
+        {{NDEFMessage}}
```
leonhsl, line 1965: As we are trying to align our APIs with the NFC forum ones, and we are sort of sure that a smart poster may have a ndef message as payload, but I'm wondering whether it's also true for an external record? I'm not familiar with nfc forum regulations about that.

index.html
```diff
@@ -1884,50 +1909,72 @@ <h2>The <dfn>NDEFRecordType</dfn> string</h2>
...
+        <a>typed array type</a>
       </td>
-      <td><a>External type record</a></td>
+      <td><a>External type record</a> (TNF=4) with TYPE= [= local type =] inside
+        the <a>NDEFMessage</a> payload of another record</td>
+    </tr>
+    <tr>
+      <td><dfn>"`unknown`"</dfn></td>
+      <td><i>unused</i></td>
+      <td>{{ArrayBuffer}}</td>
```
leonhsl, line 1982: Ditto. Are we sure an unknown record could have a ndef message as payload?

index.html
```diff
@@ -1884,50 +1909,72 @@ <h2>The <dfn>NDEFRecordType</dfn> string</h2>
...
+        <a>typed array type</a> or<br>
+        {{NDEFMessage}}
+      </td>
+      <td><a>External type record</a> (TNF=4) with TYPE= [= external type =]</td>
+    </tr>
+    <tr>
+      <td><i>[= local type =]</i></td>
       <td><i>unused</i></td>
       <td>
-        {{ArrayBuffer}} or<br>
```
leonhsl, line 1927: Why does a local type record data could only be ArrayBuffer or typed array? Given that this example  https://w3c.github.io/web-nfc/#push-a-smart-poster-message is feeding it with Number or Text.

index.html
```diff
@@ -1884,50 +1909,72 @@ <h2>The <dfn>NDEFRecordType</dfn> string</h2>
...
-        any <a>typed array type</a>
+        {{ArrayBuffer}} or any<br>
+        <a>typed array type</a>
       </td>
-      <td><a>External type record</a></td>
+      <td><a>External type record</a> (TNF=4) with TYPE= [= local type =] inside
```
leonhsl, line 1976: From what I read at https://w3c.github.io/web-nfc/#well-known-type-records-tnf-1, it seems to me that a local type record is always saved as a well-known record (TNF=1)?

index.html
```diff
@@ -2003,23 +2042,45 @@ <h2>The <dfn>NDEFRecordType</dfn> string</h2>
...
-      <td>"`opaque`"</td>
+      <td>Any embedded <a>NDEF record</a> with TYPE=<a>local type</a></td>
+      <td><a>local type</a></td>
       <td>"`application/octet-stream`"</td>
       <td>
         <a>text()</a> or<br>
         <a>json()</a> or<br>
-        <a>arrayBuffer()</a>
+        <a>arrayBuffer()</a> or<br>
+        <a>toRecords()</a>
```
leonhsl: A local type record itself always supports toRecords()? i.e. its payload is a ndef message?

index.html
zolkis: Payload of external type is up to the applications to handle (since they also define the type).
Web NFC may play nice and add this convenience feature for apps to try parsing the payload as NDEFMessage and make it easy to embed an NDEFMessage into an external type record.

The other take would be to only allow ArrayBuffer (which is closer to the NFC specs). One could argue that if developers use external types, they know well enough what they are doing.

@kenchris what's your take?

index.html
zolkis: Local types are defined at TNF=1 like this

> NFC Forum Local Types are available for use within the context of another record.

so nothing prevents apps from using them in embedded message in an external type record. 
This is the liberal interpretation of the NFC spec.

However,  the conservative interpretation would agree with you, but that means apps cannot use local types, because they are only defined for "Sp" in TNF=1.

index.html
zolkis: Those could have anything (binary data), and on the reading side a swiss army knife of trying to make sense out of it with different getters might make sense. But strictly speaking, we could say that only the ArrayBuffer getter is guaranteed to work.

On the creation side, apps can put any payload, including a full message, but I don't see any point in creating "unknown" records from apps. Since we need to deal with "unknown" records on the reading side, for symmetry I included them in the creation side as well, even though it makes little to no practical sense.

index.html
zolkis: I was thinking on this when added toRecords(). My reasoning was that if an app wants to create a tree deeper than 1, do we cut the tree at depth 1, or allow (provide tools) that apps could embed to infinity?

I took the route of allowing what is theoretically possible, but maybe we could limit the embedding at depth 1, saying "from here on, you can do it, but you are on your own, handle ArrayBuffer directly".

index.html
zolkis: Fixed

index.html
zolkis: Fixed

index.html
zolkis: Local types are also used for handover.
I fail to decode the initial intent with these local types, but it seems more and more that in actual use they are under well-known types, indeed.

Let's simplify things and I remove them from both sides of data mapping. When we get use cases, we can add them back later.

index.html
kenchris: Web NFC should definitely add convenience to write and read custom embedded records - very few native APIs do that and I think that has been holding NFC back somewhat. 

It is pretty clear from the NDEF specs that this is an intended use-case and it is standardized how it works. That is also what local types are defined for.

Smart posters use and and probably some of the handover records as well (would have to check).

NDEF specs don't talk about user facing APIs, just how things are stored and read

kenchris: > One could argue that if developers use external types, they know well enough what they are doing.

Such an argument is what limits adoption. There are quite some details to get right for average joe developer and people will just not do it

index.html
kenchris: We were pointed out that all these getters should just throw if they make no sense, and never return `null` (unless it is a valid return type, like if I stored `null` as JSON)

kenchris: In a sense, it is also totally allowed to write an NDEFMessage as opaque data, you just need to invent your own mimetype (say application/ndef) or you will have trouble handling it.

index.html
kenchris: Local type can be anything an external type can be. They are basically external types (with different type names) that are scoped to an external record

index.html
kenchris: Its easy to explain:

 They are basically external types, but it would be foolish to use up space within a record for re-adding the URL that is required in the external type name, so they avoid adding that, but restricting all types to start with lowercase or number. All standardized well known types start with uppercase so there is no conflict

external type = custom type
local type = custom type scoped within a custom type, thus can avoid the namespacing

A custom type of name say "kenny" can be used in multiple external type records but in that case they are not considered the same type (they are local types, not global) and behavior can differ.

index.html
kenchris: There is no depth limit.

I can easily create an external type record and embed another external type record inside that has embedded NDEF - as well as make a local type that has NDEF.

It is not really hard to handle, you just need to measure the size and set it as payload

index.html
leonhsl: The example code has a local type record as

> {
              recordType: "text",
              data: "Funny dance"
            },

Is it an external type record (TNF=4) or a `text` record (TNF=1)?

index.html
leonhsl: I agree we provide more conveniences to developers if NFC forum spec does not strictly forbid us from doing so.

index.html
leonhsl: I meant to ask whether a local type record is saved as TNF=1 in an nfc tag.

Another question, for the NDEFMessage acting as a payload, should all of its records be local type records?  Or the payload NDEFMessage is just to be a normal NDEFMessage?

index.html
kenchris: We need to add some prefix to these types - I have filed an issue about that today.

index.html
kenchris: No, just a normal NDEF message, which may contain local types

index.html
zolkis: No, we should not prefix. Local types should be only used inside a local context, the most well defined one being a parent record. At the moment, only wkt can have local types, but they could be also used by other types that allow embedded records, such as external type.

index.html
zolkis: OK, we stay with the swiss army knife :).

index.html
zolkis: No, only global types start with upper case, and all local types (including the standard ones) start with lowercase or number.

Local types are defined in the context of WKT, with the mention that they can be used also in other context and with other meaning. The only such place is external records. But that doesn't mean they are basically external types, they are just "local types", scoped in WKT or an external type.

index.html
zolkis: Right, those would work, even though one would need to register that MIME type if they want all parsers  handle it.

index.html
kenchris: I dont mean prefix when written to the record, but we need a way to differentiate

zolkis: The spec doesn't need to differentiate, apps do.

index.html
```diff
@@ -1884,89 +1917,108 @@ <h2>The <dfn>NDEFRecordType</dfn> string</h2>
...
+      <td><a>MIME type record</a> (TNF=2) with TYPE= [= JSON type =].
       </td>
     </tr>
     <tr>
       <td><dfn>"`opaque`"</dfn></td>
-      <td><a>MIME type</a></td>
+      <td>[= MIME type =]</a></td>
       <td>{{ArrayBuffer}}</td>
-      <td><a>MIME type record</a></td>
+      <td><a>MIME type record</a> (TNF=2) with TYPE= [= MIME type =]</td>
```
kenchris, line 1951: ```with the <a>TYPE field</a> equal to```

index.html
kenchris: I disagree but I don't want to block this PR on this, we can handle later

index.html
zolkis: would be too long in a table... I could link them, though

index.html
leonhsl: Aside from WKT and external type records, we are also trying to support local types in the context of `unknown` type record?

index.html
leonhsl: It seems to me that you're saying a local type record is definitely an external record, right? But the record type being `text` conflicts with the WKT text record.

> {
recordType: "text",
data: "Funny dance"
},

zolkis: Local type records can be in WKT (like in "Sp" and handover), these are standardized.
They can be reused in other places where an application context exists, even with changed semantics.
In addition, new local type records can be used in application context, of course they should be chosen to not conflict (locally) with other types.
The term "application context" is not well defined in NFC. So far (by the standardized examples) it seems to mean the context is a parent record.

That parent record (context) that applications can actually use is currently only the external type records.

That is why IMHO prefixing is not needed, since the context already guarantees (local) uniqueness.

If the use case is to override a WKT with a local type of same name, then that is application territory and Web NFC should not add special smartness beyond the NFC specs. 
I would not like to instill and algorithm that would say 

> "if you see the 'local:' prefix in a type name, create an embedding external type record and add this record to an NDEFMessage that is set as payload to the embedding record".

I suggest we keep it simple for now.

That "text" example above is a text record, not a local type record. In other words, for "text" one doesn't need to define a local type. If encoding is special, use a MIME type or a plain external record.

index.html
zolkis: I don't see a way how could we and why should we do that. For `unknown` records, we should just expose `ArrayBuffer`. This would encourage using external type records instead of unknown records (which are a niche case and the NFC Forum likes to avoid it).

index.html
zolkis: OK, I agree we should not have depth limitation, so a local type record could indeed contain other records and should be able to be parsed to multiple records.

zolkis:
Next: 
- Align the algorithms with the updated data mapping.
- Change `opaque` to `media`.  
- Could remove the `json` type, but keep the `json()` getter for time being and convenience.
I guess @beaufortfrancois you could (or actually want to) make the removing JSON PR after we land this one?

kenchris:
Can you land this so that I can do the encoding part?

zolkis:
Wait, it would leave the algorithms inconsistent. 
Since the encoding is pretty much an addition, if you really are in a hurry, make a PR now and I will merge, but if you can hold the horses a little longer, I can land this soon. :)

zolkis:
Now this is mergeable (unless there are obvious misses or errors) and we can do further work in later PR's.

zolkis:
@kenchris @beaufortfrancois

kenchris (approved):

Issue #366, opened 2019-10-10
What's the value of `toJSON()`?


beaufortfrancois:
Is `NDEFRecord.toJSON()` doing more than `JSON.parse(NDEFRecord.toText())`? ~~I guess it doesn't raise an error when payload is not a valid JSON message but~~ I'd like to hear from you what is its value.

```diff
  if (record.mediaType === 'application/json') {
    try {
-     const data = record.toJSON();
+     const data = JSON.parse(record.toText());
      if ('someKey' in data) { /* ... */ }
    } catch(e) {
      // Show error
    }
  }
```

At a certain extent, `toArrayBuffer()` is all it's needed to get text/json as well.

```js
const decoder = new TextDecoder();
const text = decoder.decode(record.toArrayBuffer());
const data = JSON.parse(text);
// etc.
```

kenchris:
toJSON rethrows exceptions. It will call utf+8 decode on bytes and then call json parse

Btw the mime type of json is not "json" or "text/json"

beaufortfrancois:
@kenchris I've edited my post with your feedback. Thank you!

kenchris:
It is not really doing more, though it is not doing that using JS calls. It is the same as saying that toText() is just calling toArrayBuffer and then using the TextDecoder... (Though it does a bit more when it's a text record)

jakearchibald:
There's some precedent here with [push bodies](https://developer.mozilla.org/en-US/docs/Web/API/PushMessageData).

If you're going to add this, the naming should align with push (and fetch). This is especially important as [`toJSON` already has meaning on the platform](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#Description).

The benefit of `response.json()` is the platform can parse it in a streaming manner. That isn't really true here or push, since the amount of data is really small & synchronously available.

If it's common to get the data as text, then a helper method seems reasonable (and that should be named `.text()`). Then I guess you may as well add the rest for consistency. Although I don't think I'd be arguing for that if push hadn't set precedent.

kenchris:
It is already implemented. I think it offers a lot of convenience for normal users. Not everyone understand how json and text is encoded and know how to use a TextDecoder etc. And with our API it is very easy to write JSON, so it should definitely be just as easy to read it back

jakearchibald:
Wait, `toJSON()` got through TAG review?

jakearchibald:
[...]

PR #380, opened 2019-10-17
Remove "json" recordType

beaufortfrancois:
This PR removes the "json" recordType as discussed in https://github.com/w3c/web-nfc/issues/366#issuecomment-542104572

It doesn't remove yet the `json()` NDEFRecord method. This will come in a near PR.

This PR is part of a bigger spec change. Don't merge until we all agree.

***
<a href="https://pr-preview.s3.amazonaws.com/beaufortfrancois/web-nfc/pull/380.html" title="Last updated on Oct 17, 2019, 11:16 AM UTC (d19f1a8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/380/a35eb3c...beaufortfrancois:d19f1a8.html" title="Last updated on Oct 17, 2019, 11:16 AM UTC (d19f1a8)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -732,9 +732,6 @@
         short>
       </ndef-record>
     </p>
-    <p class=note>
-      Web NFC has special handling for working with JSON <a>MIME type</a> data.
-    </p>
     </section>
 
     <section>
@@ -1119,11 +1116,12 @@
             case "url":
               console.log(`URL: ${record.text()}`);
               break;
-            case "json":
-              console.log(`JSON: ${record.json().myProperty}`);
-              break;
             case "media":
-              if (record.mediaType.startsWith('image/')) {
+              if (record.mediaType === "application/json") {
+                const decoder = new TextDecoder();
+                console.log(`JSON: ${JSON.parse(decoder.decode(record.data))}`);
+              }
+              else if (record.mediaType.startsWith('image/')) {
                 const blob = new Blob([record.arrayBuffer()], {type: record.mediaType});
 
                 const img = document.createElement("img");
@@ -1131,6 +1129,9 @@
                 img.onload = () => window.URL.revokeObjectURL(this.src);
 
                 document.body.appendChild(img);
+              }
+              else {
+                console.log(`Media not handled`);
               }
               break;
           }
@@ -1158,12 +1159,17 @@
       reader.onreading = async event => {
         console.log(`Game state: ${ JSON.stringify(event.message.records) }`);
 
+        const encoder = new TextEncoder();
         const newMessage = {
           records: [{
             id: "/mypath/mygame/update",
-            recordType: "json",
+            recordType: "media",
             mediaType: "application/json",
-            data: { level: 3, points: 4500, lives: 3 }
+            data: encoder.encode(JSON.stringify({
+              level: 3,
+              points: 4500,
+              lives: 3
+            }))
           }]
         };
         const writer = new NFCWriter();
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2576,8 +2561,8 @@ <h3>The <dfn>NFCScanOptions</dfn> dictionary</h3>
         class="example highlight">
         const options = {
           id: "https://www.w3.org/*",  // any path from the domain is accepted
-          recordType: "json",
-          mediaType: "application/*+json"  // any JSON-based MIME type
+          recordType: "media",
+          mediaType: "application/*json"  // any JSON-based MIME type
```
beaufortfrancois, line 2565: I believe it was a typo. Please tell me if I'm wrong.

index.html
```diff
@@ -3288,95 +3262,6 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
-        </ol>
-      </p>
-      <aside class=note>
-        <p>
-          JavaScript Object Notation aka JSON can be any of the following [[[ECMASCRIPT]]]
-          types: object, array, String, Boolean, Number or null, which covers a lot of
-          the [[[WEBIDL]]] types as listed here: [= JSON type =].
-        </p>
-        <p>
-          Serializing JSON to bytes can result in 0 bytes in some cases, like serializing
```
kenchris, line 3362: I still think we might want such a note about payload might become empty somewhere

index.html
zolkis: Yes, please correct these.

index.html
beaufortfrancois: If the data is empty, it makes sense that payload is empty right?
```js
writer.push({records: [
    {
      recordType: "media",
      mediaType: "application/json",
      data: new ArrayBuffer()
    }
]});
// OR
writer.push(new ArrayBuffer());
```

zolkis: Yes for creation, but it may not be a valid NDEF record in all cases, in which case the script might get an error later (when trying to push the record and the underlying platform rejecting it).

beaufortfrancois: I'm not seeing this restriction on Android: https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/nfc/NdefRecord.java#413
Does NFC spec forbid empty mime type record payloads?

zolkis: Right, the spec should not make a restriction here, either. If something is not right, the underlying platform will tell. In this case, Android won't seem to object :)

zolkis: The NDEF spec actually allows this (I don't know from where do I have the memory of being invalid in some cases).

> A payload length of 0 is allowed in which case the PAYLOAD field is omitted from the NDEF
record

zolkis: Actually that's the only normative prose I found.

> If the PAYLOAD_LENGTH field value is 0, the PAYLOAD field MUST NOT be present.

kenchris: But a developer might be confused that writing JSON can actually result in reading back an empty payload - that is the note that I would like

beaufortfrancois: I'm not sure which JSON would you write that result in an empty payload?

kenchris: The example showed how :)

kenchris: image

beaufortfrancois: Having a `toJSON` method on a class seems very unlikely and not related to Web NFC.

kenchris:
We seem to be in agreement, let's move on

PR #381, opened 2019-10-17
Introduce .encoding and .lang

kenchris:
***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Oct 17, 2019, 12:49 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Frawcdn.githack.com%2Fw3c%2Fweb-nfc%2Fa35eb3cce80d9ea9e06baff9946b1621bef1be52%2Findex.html%3FisPreview%3Dtrue)

```
Navigation Timeout Exceeded: 30000ms exceeded
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/web-nfc%23381.)._
</details>

### DIFF
--- a/index.html
+++ b/index.html
@@ -571,11 +571,11 @@
         in the [[NDEF-TEXT]] specification.
         The <a>TNF field</a> is `1` and the <a>TYPE field</a> is "`T`" (`0x54`).
         The first byte of the <a>PAYLOAD field</a> is a status byte, followed
-        by the ISO/IANA language code in US-ASCII encoding.
+        by the [=language tag=] in US-ASCII encoding.
         The rest of the payload is the actual text, encoded either in UTF-8 or
         UTF-16, as indicated by the status byte as follows:
         <ul>
-          <li>Bits 0 to 5 define the length of the IANA language code.</li>
+          <li>Bits 0 to 5 define the length of the [=language tag=].</li>
           <li>Bit 6 is `0`.</li>
           <li>
             If bit 7 if set, means the payload is encoded in UTF-8, otherwise in
@@ -1597,10 +1597,13 @@
       interface NDEFRecord {
         constructor(NDEFRecordInit recordInit);
 
-        readonly attribute NDEFRecordType recordType;
+        readonly attribute USVString recordType;
         readonly attribute USVString mediaType;
         readonly attribute USVString id;
         readonly attribute DataView? data;
+
+        readonly attribute USVString? encoding;
+        readonly attribute USVString? lang;
 
         USVString? text();
         [NewObject] ArrayBuffer? arrayBuffer();
@@ -1609,9 +1612,12 @@
       };
 
       dictionary NDEFRecordInit {
-        required NDEFRecordType recordType;
+        required USVString recordType;
         USVString mediaType;
         USVString id;
+
+        USVString encoding;
+        USVString lang;
 
         any data;
       };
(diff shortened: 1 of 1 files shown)


kenchris:
Not sure what is wrong with preview bot, but it works fine locally - no errors

index.html
```diff
@@ -1678,6 +1684,30 @@ <h4>
...
+      The <dfn>encoding</dfn> attribute represents the
+      [=encoding/name|encoding name=] used for encoding the payload in the
+      case it is textual data.
+    </p>
+    <p>
+      The <dfn>lang</dfn> attribute represents the [=language tag=]
+      of the payload in the case that was encoded.
+    </p>
+    <p>
+      A <dfn>language tag</dfn> is a <a>string</a> that matches the
```
kenchris, line 1697: This is copy paste from our web app manifest spec

index.html
```diff
@@ -1699,10 +1729,24 @@ <h4>
...
-      <dfn>recordType</dfn>, <dfn>mediaType</dfn>, <dfn>id</dfn>,
-      and <dfn>data</dfn>, respectively. The mapping from data types of an
+      with its <a>record type</a> <dfn>recordType</dfn>, and optional <a>record identifier</a>
+      <dfn>id</dfn> and payload data <dfn>data</dfn>.
+    </p>
+    <p data-dfn-for="NDEFRecordInit">
+      Additionally, there are additional optional fields that are only applicable
+      for certain <a>record types</a>:
+      <ul>
+        <li data-dfn-for="NDEFRecordInit">
```
kenchris, line 1739: Had to readd here (respec bug probably)

index.html
```diff
@@ -2541,7 +2581,7 @@ <h3>The <dfn>NFCScanOptions</dfn> dictionary</h3>
       <pre class="idl">
         dictionary NFCScanOptions {
           USVString id = "";
-          NDEFRecordType recordType;
+          USVString recordType;
           USVString mediaType = "";
```
kenchris, line 2585: HTML spec uses DOMString for media type... guess we should do the same?

Things that include urls should be USVString I have been told

zolkis: In Absolute URL the recordType would contain a URL

zolkis:
Merge when you see it fit.

zolkis (approved):

index.html
kenchris: I thought we just set it to "absolute-url" and put the url in payload?

kenchris: actually we should do that, we set it to "url" now

image

zolkis: IMHO we should not deviate from NDEF semantics here. For TNF=3 let's use the TYPE field for the URL.

zolkis: Maybe we can _also_ set `data` to the same, but `recordType` should contain the URL as well.

kenchris: As we don't expose the TNF that will be confusing. How will people detect it is a url?

kenchris: These are not supposed to contain payload so it's fine

zolkis: Maybe we should expose TNF?

PR #411, opened 2019-10-24
Add a bit more clarification about local type

kenchris:
@beaufortfrancois

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Oct 24, 2019, 10:13 AM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Frawcdn.githack.com%2Fw3c%2Fweb-nfc%2F98e3868486c551f5e68c44153695147252e8cff9%2Findex.html%3FisPreview%3Dtrue)

```
Navigation Timeout Exceeded: 30000ms exceeded
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/web-nfc%23411.)._
</details>

### DIFF
--- a/index.html
+++ b/index.html
@@ -560,11 +560,19 @@
           handover request, "`Hs`" for handover select, etc.
         </li>
         <li>
-          NFC Forum <dfn>local type</dfn> that are defined by the NFC Forum or
-          by an application, within the scope of the local context of another
-          record or application and always start with lowercase letter or a
-          number. See <a>Smart poster</a> for an example on how local types are
-          used.
+          <p>
+            NFC Forum <dfn>local type</dfn> that are defined by the NFC Forum or
+            by an application, within the scope of the local context of another
+            record or application and always start with lowercase letter or a
+            number. See <a>Smart poster</a> for an example on how local types are
+            used.
+          </p>
+          <p class=note>
+            A [=local type=] is thus defined in terms of a containing record type,
+            and thus don't need any namespacing. For this reason the same local type
+            name can be used within another record type with different meaning and
+            different payload type.
+          </p>
         </li>
       </ul>
     </div>


beaufortfrancois:
Can we explain why local types are actually a thing?

> Local Types are used whenever the burden of using a long, domain name-based External Type is too much, and there is no need to define its meaning outside of the local context.

kenchris:
> Can we explain why local types are actually a thing?

They have very little overhead. Two bytes (header plus type name length) plus the size of the type name. As the record knows exactly what they represent, no need to store mime type and the like - just store the data.

Maybe you have better ways of explaining it :)

zolkis (approved):

beaufortfrancois:
The way I understand it (and I may be wrong) is that local types are meant to avoid repeating the base record type. See example below

before 
```js
writer.push({ records: [
  {
    recordType: "example.com:shoppingItem",
    data: { records: [
      {
        recordType: "example.com:shoppingItemUrl",
        data: new TextEncoder('').encoder('https://example.com/123')
      },
      {
        recordType: "example.com:shoppingItemName",
        data: new TextEncoder('').encoder('NFC Tag')
      },
      {
        recordType: "example.com:shoppingItemDescription",
        data: new TextEncoder('').encoder('Pretty cool')
      },
  }
]});
```

after


```js
writer.push({ records: [
  {
    recordType: "example.com:shoppingItem",
    data: { records: [
      {
        recordType: "u",
        data: new TextEncoder('').encoder('https://example.com/123')
      },
      {
        recordType: "n",
        data: new TextEncoder('').encoder('NFC Tag')
      },
      {
        recordType: "d",
        data: new TextEncoder('').encoder('Pretty cool')
      },
  }
]});
```

kenchris:
Yes that is what I mean with not needing namespacing

zolkis:
It's well visible in the NDEF standard that local types are an afterthought, and not event the "context" is well defined, so we _assume_ another record, because that's how NFC Forum used them in Smart Poster. 

Whatever the thing is, there is a mechanism to use local typing and it may be useful for apps (even though I'd think those apps would choose another tech if they wanted to encapsulate their own doings).

kenchris:
It does state
> Local Types are available for use within the context of another record

context == another record

zolkis:
Then it talks (normatively) about application context that is not well defined. 
However, it does state the intention on saving on potentially long domain-name external types.

beaufortfrancois:
Shall we update local type section with the intention then?

kenchris:
Feel free to attempt and I will review :-)

PR #412, opened 2019-10-24
Added precision to local types

beaufortfrancois:
As suggested in https://github.com/w3c/web-nfc/pull/411#issuecomment-545925744, this PR adds a precision to local types.

***
<a href="https://pr-preview.s3.amazonaws.com/beaufortfrancois/web-nfc/pull/412.html" title="Last updated on Oct 25, 2019, 12:26 PM UTC (f662aa4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/412/8970772...beaufortfrancois:f662aa4.html" title="Last updated on Oct 25, 2019, 12:26 PM UTC (f662aa4)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -564,8 +564,9 @@
             NFC Forum <dfn>local type</dfn> that are defined by the NFC Forum or
             by an application, within the scope of the local context of another
             record or application and always start with lowercase letter or a
-            number. See <a>Smart poster</a> for an example on how local types are
-            used.
+            number. Those are usually short to save on potentially long domain-name
+            containing record types. See <a>Smart poster</a> for an example on
+            how local types are used.
           </p>
           <p class=note>
             A [=local type=] is thus defined in terms of a containing record type,
--- a/index.html
+++ b/index.html
@@ -562,11 +562,12 @@
         <li>
           <p>
             NFC Forum <dfn>local type</dfn> that are defined by the NFC Forum or
-            by an application, within the scope of the local context of another
-            record or application and always start with lowercase letter or a
-            number. Those are usually short to save on potentially long domain-name
-            containing record types. See <a>Smart poster</a> for an example on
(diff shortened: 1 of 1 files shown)


kenchris (approved):

index.html
```diff
@@ -564,8 +564,9 @@ <h4>
             NFC Forum <dfn>local type</dfn> that are defined by the NFC Forum or
             by an application, within the scope of the local context of another
             record or application and always start with lowercase letter or a
-            number. See <a>Smart poster</a> for an example on how local types are
-            used.
+            number. Those are usually short to save on potentially long domain-name
```
zolkis: > Those are usually short to save on potentially long domain-name containing record types.

Local types are usually short strings that are unique in the local context of the containing record. They are used when it's not worth defining full external types (that might contain long domain names) and when there is no need to define the meaning of the types outside of the local context of the containing record.

kenchris: Local types are string defined type names that are unique only within the local context of the containing record. As these are contained within another record, there is no need for adding a developer owned domain as namespace as required by external types, which takes up additional storage, as their meaning within the record should be known if the containing record type is known.

zolkis: Francois, feel free to reformulate these comments.

beaufortfrancois: What do you think of the new patch?

zolkis (approved):

index.html
zolkis: Looks good. Next, perhaps we also need to add precision to algorithms wrt the external and local types parsing/validation. We should require FQDN in external types, then we'd need a pointer to an external algorithm on parsing/validating FQDN (I know there are a number of reg-exps around). Also, perhaps we need to factor out an algo for validating local types (check for local context and case).

PR #415, opened 2019-10-28
ignore media type for anything but mime

### LINKED ISSUE #387: NDEFRecord mediaType should be non null for media records
I'd like to make sure mapping is symmetric as much as possible and one bit that seems off is `mediaType`.

NDEFRecordInit mediaType is unused for all records but media.
However NDEFRecord mediaType is set for "text", "url", "external type", "local type", and "unknown".

```js
await writer.push({ records: [
  {
    recordType: 'media',
    mediaType: 'application/octet-stream',
    data: new ArrayBuffer()
  },
  {
    recordType: 'text',
    data: 'hello'
  }
]});

reader.scan();
reader.onreading = event => {
  const decoder = new TextDecoder();
  for (const record of event.message.records) {
   if (record.recordType === 'media') {
    console.assert(record.mediaType === 'application/octet-stream');
  } else if (record.recordType === 'text') {
    // I didn't provide one so it should be empty.
    // And if I had provided a valid mediaType but different from 'text/plain' it would be unexpected.
    console.assert(record.mediaType === '');
    // Same would apply to "url"
  }
};
```

beaufortfrancois:
This PR makes `mediaType` optional in NDEFRecord so that it only applies to mime records.

FIX https://github.com/w3c/web-nfc/issues/387

***
<a href="https://pr-preview.s3.amazonaws.com/beaufortfrancois/web-nfc/pull/415.html" title="Last updated on Oct 28, 2019, 1:08 PM UTC (74dc8d7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/415/05d62b5...beaufortfrancois:74dc8d7.html" title="Last updated on Oct 28, 2019, 1:08 PM UTC (74dc8d7)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1203,7 +1203,6 @@
     <pre class="example">
       const reader = new NDEFReader();
       await reader.scan({
-        recordType: "mime",
         mediaType: "application/*json"
       });
       reader.onreading = event => {
@@ -2528,7 +2527,6 @@
         class="example highlight">
         const options = {
           id: "https://www.w3.org/*",  // any path from the domain is accepted
-          recordType: "mime",
           mediaType: "application/*json"  // any JSON-based MIME type
         }
       </pre>
@@ -2537,7 +2535,6 @@
         class="example highlight">
         const options = {
           id: "https://w3.org/info/restaurant/daily-menu/",
-          recordType: "mime",
           mediaType: "application/octet-stream"
         }
       </pre>
--- a/index.html
+++ b/index.html
@@ -1623,7 +1623,7 @@
         constructor(NDEFRecordInit recordInit);
 
         readonly attribute USVString recordType;
-        readonly attribute USVString mediaType;
+        readonly attribute USVString? mediaType;
         readonly attribute USVString id;
         readonly attribute DataView? data;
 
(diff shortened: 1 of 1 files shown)


zolkis:
Wait, are you saying that once we have `mediaType` specified it means we MUST have a media record, making `recordType` unnecessary, hence omitted?

I think we should keep `recordType` set to `mime` in this case.

beaufortfrancois:
Nope. `recordType` is always mandatory. `mediaType` though is filled and required only for "mime" records.


image

index.html
```diff
@@ -1203,7 +1203,6 @@ <h4>
     <pre class="example">
       const reader = new NDEFReader();
       await reader.scan({
-        recordType: "mime",
```
zolkis, line 1206: `recordType` not needed? are we counting on the implementation inferring it should be "mime" type?

index.html
```diff
@@ -2008,35 +2007,35 @@ <h2>The <dfn>record type</dfn> string</h2>
       <td>0</td>
       <td><i>unused</i></td>
       <td>"`empty`"</td>
-      <td>""</td>
+      <td>`null`</td>
```
zolkis: `null` or `undefined`? I'd prefer the latter - but do you have reasons for `null`?

index.html
```diff
@@ -3259,16 +3241,6 @@ <h3><dfn>Writing or pushing content</dfn></h3>
             {{BufferSource}}, [= exception/throw =] a {{TypeError}}
             and abort these steps.
           </li>
-          <li>
```
zolkis: Isn't this part of "mapping binary data to NDEF"? It is for MIME type records. Didn't you want to remove from external type (mapping external data to NDEF)?

index.html
beaufortfrancois: This is scanning, not writing.

index.html
beaufortfrancois: `undefined` is fine for me as well.

index.html
beaufortfrancois: Good catch indeed. I was too quick. I'll fix this.

beaufortfrancois:
@zolkis I've addressed your feedback.

zolkis (approved):
Thanks! Occam rulez :)

PR #419, opened 2019-10-28
Throw TypeError if mediaType for non mime records

### LINKED ISSUE #387: NDEFRecord mediaType should be non null for media records

beaufortfrancois:
As suggested by https://github.com/w3c/web-nfc/issues/387#issuecomment-543624143, we should throw a TypeError if `mediaType` is not `undefined` for non mime records.

Fix https://github.com/w3c/web-nfc/issues/387

***
<a href="https://pr-preview.s3.amazonaws.com/beaufortfrancois/web-nfc/pull/419.html" title="Last updated on Oct 28, 2019, 1:30 PM UTC (95cfb30)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/419/e845e7d...beaufortfrancois:95cfb30.html" title="Last updated on Oct 28, 2019, 1:30 PM UTC (95cfb30)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2995,6 +2995,10 @@
         these steps:
         <ol class=algorithm>
           <li>
+            If |record|'s |mediaType| is not `undefined`, [= exception/throw =]
+            a {{TypeError}} and abort these steps.
+          </li>
+          <li>
             Let |ndef| be the notation for the <a>NDEF record</a> to
             be created by the UA.
           </li>
@@ -3029,6 +3033,10 @@
           "`text/vcard`".
         </p>
         <ol class=algorithm>
+          <li>
+            If |record|'s |mediaType| is not `undefined`, [= exception/throw =]
+            a {{TypeError}} and abort these steps.
+          </li>
           <li>
             If the type of |record|'s data is not a {{DOMString}} or a {{BufferSource}},
             [= exception/throw =] a {{TypeError}} and abort these steps.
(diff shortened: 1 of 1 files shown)


zolkis:
We could also ignore it. 
Somehow when I see it explicitly in the algorithm, I find myself asking "why is this relevant here?".
Leaving it to your judgement. @kenchris?

zolkis (approved):

beaufortfrancois:
I think* the goal is to make sure we get the same NDEFMessage content between what's pushed and what's read.

```js
await writer.push({
  records: [{
    'recordType': 'text',
    'mediaType': 'text/foo'
  }]
});

reader.scan();
reader.onreading = ({ message }) => {
  const record = message.records[0];
  console.log(record.mediaType == 'text/foo'); //  <- false if TypeError is not thrown
});

zolkis:
Yes. Somehow I mistook this change was for reading. For writing, definitely we should check. Merging.

PR #422, opened 2019-10-29
Add unknown record JS example

beaufortfrancois:
As suggested by @kenchris, this PR adds a JS example featuring unknown records inside an external record.

***
<a href="https://pr-preview.s3.amazonaws.com/beaufortfrancois/web-nfc/pull/422.html" title="Last updated on Oct 29, 2019, 10:37 AM UTC (dc4fdfe)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/422/766310e...beaufortfrancois:dc4fdfe.html" title="Last updated on Oct 29, 2019, 10:37 AM UTC (dc4fdfe)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1415,6 +1415,51 @@
       ]});
     </pre>
   </section>
+
+  <section> <h3>Push and read unknown records inside an external record</h3>
+    <p>
+      Unknown type records may be useful inside external type records as
+      developers know what they represent and therefore can avoid specifying the
+      mime type.
+    </p>
+    <pre class="example">
+      const encoder = new TextEncoder();
+      const writer = new NDEFWriter();
+      writer.push({ records: [
+        {
+          recordType: "example.com:shoppingItem", // External record
+          data: {
+            records: [
+              {
+                recordType: "unknown", // Shopping item name
+                data: encoder.encode("Food")
+              },
+              {
+                recordType: "unknown", // Shopping item description
+                data: encoder.encode("Provide nutritional support for an organism.")
+              }
+            ]
+          }
+        }
+      ]});
+    </pre>
+    <pre class="example">
+      const reader = new NDEFReader();
+      reader.scan({ recordType: "example.com:shoppingItem" });
+      reader.onreading = event => {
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1415,6 +1415,51 @@ <h4>
...
+      mime type.
+    </p>
+    <pre class="example">
+      const encoder = new TextEncoder();
+      const writer = new NDEFWriter();
+      writer.push({ records: [
+        {
+          recordType: "example.com:shoppingItem", // External record
+          data: {
+            records: [
```
zolkis, line 1432: Not sure if the order of records is preserved. The app then would not be 100% if the first unknown record is the one they assume to be. I suggest using only one `unknown` record in the message, or include an id. (not that I would endorse using `unknown` records, as `external` is probably better for all use cases. I think `unknown` is mostly as a backup for reading use cases (if nothing works). But I need to double check these.

beaufortfrancois: According to the spec below, it looks like it depends on the platform.

> Actual NDEF records do not carry an index number; the ordering is implicitly given by the order in which the NDEF records are serialized. For example, if NDEF records are repackaged by an intermediate application, then that application is responsible for ensuring that the order of NDEF records is preserved.

zolkis: Then having two `unknown` records in a message is a hazard :).

kenchris: But when written as a sub record with web nfc we could enforce order... I dont actually see why changing the order would make sense

zolkis: What I meant was that in an (educative) example it would be better either to use `unknown` + id, or not use two `unknown` records in the same message, but the example intro ("developers know what they represent and therefore can avoid specifying the mime type") makes clear the intention. I am not blocking this, just made a remark.

zolkis (approved):

index.html
beaufortfrancois: Does that look good to you as well @kenchris?

kenchris: yes

kenchris (approved):

Issue #391, opened 2019-10-18
Avoid creating custom errors


kenchris:
We are told in other specs to avoid creating custom error types especially if they add nothing special

![image](https://user-images.githubusercontent.com/1101339/67083754-0609ab80-f19c-11e9-8ff1-3a4bd25d1208.png)

kenchris:
@rakuco just fixed wake lock spec in this sense

rakuco:
For more context: my commit fixed w3c/wake-lock#238 after following https://w3ctag.github.io/design-principles/#state-and-subclassing, as our `WakeLockEvent` had a single `lock` member that also happened to be identical to `Event.target` anyway.

I'm not very familiar with the NFC spec though, I guess if you were to drop `NFCErrorEvent` you'd still need to find a way to put that `error` attribute somewhere.

kenchris:
@reillyeon might have ideas

reillyeon:
In the Generic Sensors API the [`SensorErrorEvent`](https://w3c.github.io/sensors/#the-sensor-error-event-interface) is used to indicate that an error occurred while the sensor was active. The intuition is that a sensor can be initialized successfully and begin producing readings but then encounter and error (such as disconnection) and there needs to be a mechanism to report that. As the sensor has already been started successfully the Promise returned by the `start()` method has already settled, so an additional mechanism is needed.

In this specification I don't see `NFCErrorEvent` being thrown in any circumstance other than in the steps to initialize a scan. It is reasonable in that case to remove the event and have `scan()` return a Promise that is rejected instead. If, however, we want to report later error conditions we will need to have an event, just like in Generic Sensors.

I wonder if this concept is reusable to the point that the TAG would recommend standardizing a `DOMExceptionEvent` or `ErrorEvent` that could be reused across specifications for this purpose.

beaufortfrancois:
I agree with @reillyeon.

The funny thing is that push already rejects a promise with an appropriate DOMException, not a NDEFPushErrorEvent.

```js
const reader = new NDEFReader();
try {
  await reader.scan();
  // Scanning is happening...
} catch(e) {
  // NDEF scanning could not start.
}
```

I'll be happy to take action and update spec accordingly.

beaufortfrancois:
@kenchris Please assign this issue to me.

PR #432, opened 2019-11-05
Add onerror NDEFReader event

### LINKED ISSUE #279: Notify NFCReaders of failures when trying to get a message from a tag?
See
https://developer.android.com/reference/android/nfc/tech/Ndef.html#getNdefMessage().
Once a tag is discovered, we call `android.nfc.tech.Ndef.getNdefMessage()` to try to get a message from it, which may
  - return null if the tag is in the `INITIALIZED` state, i.e. formatted but empty
  - throw exceptions: `TagLostException,` `IOException,` or `FormatException`

In such cases do we need to let all active `NFCReaders` be aware of these failures so that they can choose how to proceed next? i.e. dispatching some error events properly to `NFCReaders.`

Currently http://w3c.github.io/web-nfc/#the-nfc-reading-algorithm only describes the case that a ndef message has already been gotten from the tag, so does the implementation in Chromium.

beaufortfrancois:
This PR adds back the `onerror` NDEFReader event. It is only fired when the NFC device in proximity range does not expose NDEF technology for reading or formatting when `scan()` has started.

FIX: https://github.com/w3c/web-nfc/issues/279

***
<a href="https://pr-preview.s3.amazonaws.com/beaufortfrancois/web-nfc/pull/432.html" title="Last updated on Nov 5, 2019, 7:52 AM UTC (fa84e47)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/432/4d4e377...beaufortfrancois:fa84e47.html" title="Last updated on Nov 5, 2019, 7:52 AM UTC (fa84e47)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2137,6 +2137,7 @@
     interface NDEFReader : EventTarget {
       constructor();
 
+      attribute EventHandler onerror;
       attribute EventHandler onreading;
 
       Promise&lt;void&gt; scan(optional NDEFScanOptions options={});
@@ -2272,6 +2273,10 @@
   <p>
     The <dfn data-dfn-for="NDEFReader">onreading</dfn> is an {{EventHandler}} which is called to notify
     that new reading is available.
+  </p>
+  <p>
+    The <dfn data-dfn-for="NDEFReader">onerror</dfn> is an {{EventHandler}}
+    which is called to notify that an error happened when reading.
   </p>
   <section><h3>NFC state associated with the settings object</h3>
   <p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2273,6 +2274,10 @@ <h2>The <dfn>record type</dfn> string</h2>
     The <dfn data-dfn-for="NDEFReader">onreading</dfn> is an {{EventHandler}} which is called to notify
     that new reading is available.
   </p>
+  <p>
+    The <dfn data-dfn-for="NDEFReader">onerror</dfn> is an {{EventHandler}}
+    which is called to notify that an error happened when reading.
```
kenchris: During?

kenchris (approved):

index.html
beaufortfrancois: You mean "during reading" instead of "when reading"?

kenchris:
Yes

beaufortfrancois:
@kenchris Feedback addressed.
@zolkis Feel free to merge if you approve.

zolkis (approved):

reillyeon:
Nack, this is exactly what I warned against in my comment on #391. We should carefully consider before having two mechanisms for reporting errors.

I can see a way to keep this error but we should make sure there are examples demonstrating how we expect developers to catch both kinds of errors.

beaufortfrancois:
`scan` will reject only if scanning cannot be started (invalid options, hardware not available, permission denied, etc.)
`onerror` will be fired only if there are errors happening during an already started scan (not formatted tag, context lost)

reillyeon:
Can you update the examples to include a handler for the error event?

beaufortfrancois:
Good idea! I'll do that tomorrow.

beaufortfrancois:
@reillyeon Here it is: https://github.com/w3c/web-nfc/pull/439

PR #439, opened 2019-11-07
Add onerror example

beaufortfrancois:
This PR addresses comment posted by @reillyeon in https://github.com/w3c/web-nfc/pull/432#issuecomment-550478360

***
<a href="https://pr-preview.s3.amazonaws.com/beaufortfrancois/web-nfc/pull/439.html" title="Last updated on Nov 18, 2019, 1:51 PM UTC (c5109bc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/439/6b0cdb8...beaufortfrancois:c5109bc.html" title="Last updated on Nov 18, 2019, 1:51 PM UTC (c5109bc)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1131,6 +1131,28 @@
     </pre>
   </section>
 
+
+  <section> <h3>Handle scanning errors</h3>
+    <p>
+      This example shows what happens when [=NDEFReader/scan=] promise rejects and
+      [=NDEFReader/onerror=] is fired.
+    </p>
+    <pre class="example">
+      const reader = new NDEFReader();
+      reader.scan().then(() => {
+        console.log("Scan started successfully.");
+        reader.onerror = event => {
+          console.log("Error! Cannot read data from NFC tag.");
+        };
+        reader.onreading = event => {
+          console.log("NDEF message read.");
+        };
+      }).catch(error => {
+        console.log(`Scan failed to start :-( try again: ${error}.`);
+      });
+    </pre>
+  </section>
+
   <section> <h3>Read data from tag, and write to empty ones</h3>
     <p>
       This example shows reading various different kinds of data which can be
--- a/index.html
+++ b/index.html
@@ -1142,13 +1142,13 @@
       reader.scan().then(() => {
         console.log("Scan started successfully.");
         reader.onerror = event => {
-          console.log("Error! Cannot read data from NFC tag.");
+          console.log("Error! Cannot read data from the NFC tag. Try a different one?");
         };
         reader.onreading = event => {
           console.log("NDEF message read.");
         };
       }).catch(error => {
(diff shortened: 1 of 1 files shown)


zolkis (approved):

index.html
```diff
@@ -1131,6 +1131,28 @@ <h4>
     </pre>
   </section>
 
+
+  <section> <h3>Handle scanning errors</h3>
+    <p>
+      This example shows when [=NDEFReader/scan=] promise rejects and
```
kenchris: shows when sounds weird... shows how to handle? shows what happens?

beaufortfrancois:
Let's wait for @reillyeon to review before merging.

index.html
```diff
@@ -1131,6 +1131,28 @@ <h4>
...
+    <p>
+      This example shows what happens when [=NDEFReader/scan=] promise rejects and
+      [=NDEFReader/onerror=] is fired.
+    </p>
+    <pre class="example">
+      const reader = new NDEFReader();
+      reader.scan().then(() => {
+        console.log("Scan started successfully.");
+        reader.onerror = event => {
+          console.log("Error! Cannot read data from NFC tag.");
```
reillyeon: The log message below says "try again" but this case is really the one were retrying makes sense (tell the user to hold their device against the tag for longer) whereas an error below indicates a more fatal issue.

beaufortfrancois: The onerror event is raised only for `IOException` and `FormatException`, not `TagLostException`. I believe* this means holding device longer won't change anything. Both are fatal errors.
Do you want to me remove "try again" to make it clearer?

Throws | Reason
-- | --
TagLostException | if the tag leaves the field
IOException | if there is an I/O failure, or the operation is canceled
FormatException | if the NDEF Message on the tag is malformed

reillyeon: Interesting. Do you think we should fire an event on `TagLostException`? That seems like the most likely and most actionable error that a site could encounter.

Given the current behavior I would word the error as, "Error! Cannot read data from the NFC tag. Try a different one?"

My goal is for this example code to demonstrate to developers what parts of their UI should be responding to different events, and how. This also gives us a good opportunity to test whether or not developers will be able to easily code against these potential errors.

beaufortfrancois: I don't think `TagLostException`provides that much value as users will try again if they see app is not reacting, while `IOException ` and `FormatException` are more "It's over" errors. They're nothing users can do about it as they use wrong formatted tags, and website will only show them something like "Sorry! This tag is not compatible with this application".

I'll update my example with your error message.

beaufortfrancois: See https://github.com/w3c/web-nfc/pull/439/commits/c5109bca16b5790ed3196dc2f11804ab445906e1

reillyeon (approved):

PR #440, opened 2019-11-08
Update Security section

zolkis:
Added attacker model, threats analysis etc from the Security and Privacy document, updated for the current API.

Commented out the sections dealing with relaxing permissions based on the ID field,
but added as mechanisms for content identification.

Signature records are not specified in this PR, only the general policies with them.

***
<a href="https://pr-preview.s3.amazonaws.com/zolkis/web-nfc/pull/440.html" title="Last updated on Nov 21, 2019, 1:55 PM UTC (ce9c576)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/440/28024bb...zolkis:ce9c576.html" title="Last updated on Nov 21, 2019, 1:55 PM UTC (ce9c576)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4293,7 +4293,8 @@
             <li>
               For trusting the integrity of the data exchanged via NFC, user
               agents MAY use encrypted <a>NFC content</a> with key management
-              based on Public Key Infrastructure. Key management is out of the scope of Web NFC.
+              based on Public Key Infrastructure.
+              Key management is out of the scope of Web NFC.
               <p class="note">
                 This is different from application level end to end encryption.
                 Key management is then in application domain.
@@ -4328,30 +4329,29 @@
 
     <section> <h3>Content identification</h3>
       <p>
-        Implementations or applications MAY use an identification mechanism for telling apart “Web NFC content" from generic NFC content. Also, a
-        mechanism for conveying implementation specific metadata which is not exposed to applications MAY be implemented.
+        Implementations or applications MAY use an identification mechanism for
+        telling apart “Web NFC content" from generic NFC content. Also, a
+        mechanism for conveying implementation specific metadata which is not
+        exposed to applications MAY be implemented.
       </p>
+      <section> <h4>Store site URL as record identifier when writing data</h4>
+        <p>
+          When pushing an <a>NDEF message</a>, the [= host/registrable domain =],
+          <a data-cite="dom#concept-host-serializer">serialized</a>,
+          of the <a>current settings object</a> MAY be stored as the <a>record
+          identifier</a> in each top-level <a>NDEF Record</a>.
+          This does not make the payload trusted.
+        </p>
+      </section>
+      <section> <h4>Include URI records with registrable domain</h4>
+        <p>
+          When pushing an <a>NDEF message</a>, the [= host/registrable domain =],
+          <a data-cite="dom#concept-host-serializer">serialized</a>,
+          of the <a>current settings object</a> MAY be stored as a signed
+          <a>URI record</a>. This does not make other records trusted.
+        </p>
+      </section>
     </section>
-
-    <!--
-    <section> <h3>Store site URL as record identifier when writing data</h3>
-      <p>
-        When pushing an <a>NDEF message</a>, the [= host/registrable domain =],
-        <a data-cite="dom#concept-host-serializer">serialized</a>,
-        of the <a>current settings object</a> MAY be stored as the <a>record
-        identifier</a> in each top-level <a>NDEF Record</a>. This does not make
-        the payload trusted.
-      </p>
-    </section>
-    <section> <h3>Include URI records with registrable domain</h3>
-      <p>
-        When pushing an <a>NDEF message</a>, the [= host/registrable domain =],
-        <a data-cite="dom#concept-host-serializer">serialized</a>,
-        of the <a>current settings object</a> MAY be stored as a signed
-        <a>URI record</a>. This does not make other records trusted.
-      </p>
-    </section>
-    -->
   </section>
 
   <section> <h3>Security policies</h3>
@@ -4439,24 +4439,35 @@
         allowing surface to attacks. Version 2.0 of [[NFC-SECURITY]] included
         tag hardware attributes in the signature and allowed for shorter
         certificates.
-        Therefore the following additional policies apply:
+        In this version of the specification the following policies are
+        recommended:
         <ul>
           <li>
-            A <a>smart poster</a> can be trusted only if signed by using
+            A <a>smart poster</a> MAY be trusted only if signed by using
             a single <a>NDEF signature</a> record by the same issuer.
           </li>
           <li>
-            An <a>NDEF message</a> can be trusted only if signed by
+            An <a>NDEF message</a> MAY be trusted only if signed by
             a single <a>NDEF signature</a> record by the same issuer.
+          </li>
+          <li>
+            User agents expose <a>NDEF signature</a> records without verifying
+            them. It is applications' responsibility to verify signatures and
+            to sign <a>NFC content</a>.
+            <p class="note">
+              Indication whether <a>NFC content</a> is trusted may be added in
+              future versions. Also, origin policies MAY be applied in order to
+              relax user prompting when a trust chain could be established.
+            </p>
           </li>
         </ul>
       </div>
     </section>
-      <!--p>
-        For Bluetooth and WiFi handover (supported in later versions),
-        the user should have to grant access to the secondary API and must be
-        able to properly understand what they are granting.
-      </p-->
+    <p class="note">
+      For Bluetooth and WiFi handover (supported in later versions),
+      the user should have to grant access to the secondary API and must be
+      able to properly understand what they are granting.
+    </p>
   </section> <!-- Policies -->
 </section> <!-- Security and Privacy  -->
 
--- a/index.html
+++ b/index.html
@@ -2612,10 +2612,6 @@
         To <dfn>obtain push permission</dfn>, run these steps:
         <ol class=algorithm>
           <li>
-            If there is a <a>prearranged trust relationship</a>,
-            return `true`.
-          </li>
-          <li>
             Run the
             <a>query a permission</a> steps for the
             <a>Web NFC permission name</a> until completion.
@@ -3502,10 +3498,6 @@
         To <dfn>obtain reading permission</dfn>, run these steps:
         <ol>
           <li>
-            If there is a <a>prearranged trust relationship</a>,
-            return `true`.
-          </li>
-          <li>
             Otherwise, if the user has earlier denied permission for the calling
             <a>origin</a> for all future calls of
             <a data-link-for="NDEFReader">scan()</a>
@@ -4095,9 +4087,9 @@
       specification.
     </p>
     <p>
-      The integrity of <a>NFC content</a> SHOULD NOT be trusted by web sites
-      and applications unless it is encrypted by applications, or a
-      <a>prearranged trust relationship</a> exists.
+      The integrity of <a>NFC content</a> MAY be trusted by web sites
+      and applications only if encrypted by applications, or if signed
+      with an <a>NDEF signature</a>.
     </p>
     <p>
       Security considerations for media types in general are discussed in
@@ -4229,18 +4221,15 @@
       In order to support security policies concerning Web NFC, the following
       mechanisms are recommended for the implementations.
     </p>
-    <section> <h3><dfn>Obtain permission</dfn></h3>
-      Implementations SHOULD use one or more of the following mechanisms for
-      <dfn>obtaining permission</dfn>.
+    <section> <h3><dfn>Obtaining permission</dfn></h3>
+      Implementations SHOULD use one or more of the following mechanisms to
+      <dfn>obtain permission</dfn>.
       <ul>
         <li>
           A mechanism for obtaining <a>expressed permission</a> of the user.
         </li>
         <li>
           A mechanism for <a>ask for forgiveness</a> policies.
-        </li>
-        <li>
-          A mechanism for <a>prearranged trust relationship</a>.
         </li>
       </ul>
       <section> <h4><dfn>Expressed permission</dfn></h4>
@@ -4265,47 +4254,22 @@
           handle future occurences of the same operation by the same action.
         </p>
       </section>
-      <section> <h4><dfn>Prearranged trust relationship</dfn></h4>
-        <div>
-          A prearranged trust relationship means the UA has already
-          established a trust relationship for a certain operation using a
-          platform specific mechanism, so that an <a>expressed permission</a>
-          from the user is not any more needed. A few examples are given below.
-          <ul>
-            <li>
-              Web apps installed from a store, or web pages installed to home
-              screen (with [[appmanifest]]) MAY be considered trusted by the user
-              agent. In this case the trust concerns the web page and allowed
-              NFC functionality, but does not necessarily also concern the
-              data exchanged through NFC.
-            </li>
-            <li>
-              For trusting the integrity of the data exchanged via NFC, user
-              agents MAY use encrypted <a>NFC content</a> with key management
-              based on Public Key Infrastructure.
-              Key management is out of the scope of Web NFC.
-              <p class="note">
-                This is different from application level end to end encryption.
-                Key management is then in application domain.
-              </p>
-            </li>
-            <li>
-              User agents MAY consider using <a>NDEF signature</a> records for
-              <a>NFC content</a> integrity. When <a>NDEF record</a>s are signed
-              with a site certificate, then that is considered
-              <a>prearranged trust relationship</a> and browser origin policy
-              MAY be applied to relax the security policies in this document,
-              in the case when the user agent wants to minimize or eliminate
-              user prompts involved in obtaining permissions.
-              The signing tools and Public Key Infrastructure for key management
-              are out of scope of Web NFC.
-            </li>
-          </ul>
-        </div>
-      </section>
     </section>
 
-    <section> <h3>Sign NDEF records</h3>
+    <section> <h3>Encrypting <a>NFC content</a></h3>
+      <p>
+        For trusting the integrity of the data exchanged via NFC, user agents
+        MAY use encrypted <a>NFC content</a> with key management based on
+        Public Key Infrastructure (PKI).
+        Key management is out of the scope of Web NFC.
+      </p>
+      <p class="note">
+        This is different from application level end to end encryption.
+        Key management is then in application domain.
+      </p>
+    </section>
+
+    <section> <h3>Signing NDEF records</h3>
       <p>
         When pushing an <a>NDEF message</a>, one, more, or all records MAY be
         signed using an <a>NDEF signature</a>.
@@ -4313,6 +4277,14 @@
       <p>
         In order to build a trust chain, a Public Key Infrastructure is also
         needed for key management.
+      </p>
+      <p>
+        For tags signed with <a>NDEF signature</a> version 1.0 ([[NFC-SECURITY]]),
+        the signature is applied only to the <a>TYPE field</a>, <a>ID field</a>
+        and <a>PAYLOAD field</a>, leaving out the first byte of the NDEF header,
+        allowing surface to attacks. Version 2.0 of [[NFC-SECURITY]] included
+        tag hardware attributes in the signature and allowed for shorter
+        certificates.
       </p>
     </section>
 
@@ -4403,7 +4375,7 @@
       -->
     </section>
 
-    <section> <h4>Warn risk of physical location leak</h4>
+    <section> <h4>Warn about risk of physical location leak</h4>
       <p>
         When listening for and pushing <a>NFC content</a>,
         the UA MAY warn the user that the given <a>origin</a> may be able to
@@ -4422,12 +4394,6 @@
 
     <section> <h4>Signing <a>NFC content</a></h4>
       <div>
-        For tags signed with <a>NDEF signature</a> version 1.0 ([[NFC-SECURITY]]),
-        the signature is applied only to the <a>TYPE field</a>, <a>ID field</a>
-        and <a>PAYLOAD field</a>, leaving out the first byte of the NDEF header,
-        allowing surface to attacks. Version 2.0 of [[NFC-SECURITY]] included
-        tag hardware attributes in the signature and allowed for shorter
-        certificates.
         In this version of the specification the following policies are
         recommended:
         <ul>
--- a/index.html
+++ b/index.html
@@ -235,12 +235,12 @@
   <div>
     An <dfn>NFC tag</dfn> is a passive NFC device.
     The <a>NFC tag</a> is powered by magnetic induction when an active NFC
-    device is in proximity range. An <a>NFC tag</a> contains a single
-    <a>NDEF message</a>.
+    device is in proximity range. An <a>NFC tag</a> that supports <a>NDEF</a>
+    contains a single <a>NDEF message</a>.
     <p class="note">
       The way of reading the message may happen through proprietary
       technologies, which require the reader and the tag to be of the same
-      manufacturer. Implementations are expected to encapsulate this.
+      manufacturer. They may also expose an <a>NDEF</a> message.
     </p>
   </div>
   <p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -4116,91 +4076,298 @@ <h3>Parsing content</h3>
...
 <section> <h2 id="security">Security and Privacy</h2>
   <p>
-    The trust model, attacker model, threat model and possible mitigation
-    proposals for Web NFC are presented in the
-    <a href="http://w3c.github.io/web-nfc/security-privacy.html">
-    Security and Privacy</a> document. This section presents the chosen
-    security and privacy model through normative requirements to
-    implementations.
+    NFC technology involves multiple levels of security. Payments done with NFC
+    are considered to be at least as secure as with payment cards at hardware
```
kenchris: "as secure" not "to be at least" sounds bad

index.html
```diff
@@ -4116,91 +4076,298 @@ <h3>Parsing content</h3>
...
+    level, but the whole software stack needs to be security hardened.
+    However, data transfers involve more threats. In order to mitigate these,
```
kenchris: more -> additional ?

kenchris: NDEF based data transfers?

index.html
```diff
@@ -4116,91 +4076,298 @@ <h3>Parsing content</h3>
...
-    The integrity of <a>NFC content</a> SHOULD NOT be trusted when
-    used for implementing security policies, for instance the authenticity of
-    <a>record identifier</a>, unless a <a>prearranged trust relationship</a> exists.
-  </p>
-  <p>
-    Security considerations for media types in general are discussed in
-    [[RFC2048]] and [[RFC2046]].
-  </p>
+    <p>
+      Web pages using Web NFC are not trusted.
```
kenchris: The users (web site and apps) of Web NFC are not considered trusted

index.html
```diff
@@ -4116,91 +4076,298 @@ <h3>Parsing content</h3>
...
+      This means that the user needs to be aware of exactly what a web page is
```
kenchris: the end user

index.html
zolkis: User = person. Web pages are neither.

index.html
zolkis: It's confusing that you mean web pages by user and persons by "end user".
I think it is better to use web pages and users.

index.html
```diff
@@ -4116,91 +4076,298 @@ <h3>Parsing content</h3>
...
+      intending to do with NFC at any given moment. Implementations need to
```
kenchris: This sounds like something most users can never live up to. Maybe write it softer, "A end user should be aware that the web site might misuse the data... "

index.html
```diff
@@ -4116,91 +4076,298 @@ <h3>Parsing content</h3>
...
+      make sure that when the user authorizes a method of this API, then only
+      that action is run, without side effects, and exactly in the context and
+      the number of times the user allows the execution of NFC related
+      operations, according to the algorithmic steps detailed in this
+      specification.
+    </p>
+    <p>
+      The integrity of <a>NFC content</a> SHOULD NOT be trusted by web pages
```
kenchris: mix of web pages and applications (here meaning the same)

index.html
```diff
@@ -4116,91 +4076,298 @@ <h3>Parsing content</h3>
...
+      unless it is encrypted by applications, or a
+      <a>prearranged trust relationship</a> exists.
```
kenchris: I was told before in wake lock to get rid of this "prearranged trust relationship"

index.html
```diff
@@ -4116,91 +4076,298 @@ <h3>Parsing content</h3>
...
+    </p>
+    <p>
+      Security considerations for media types in general are discussed in
```
kenchris: MIME types (link)

index.html
```diff
@@ -4116,91 +4076,298 @@ <h3>Parsing content</h3>
...
-    </ul>
-  </div>
+  <section class="informative"> <h2>Attacker model</h2>
+    <p>
+      The following attacker patterns have been considered:
+      <ul>
+        <li>
+          malicious web page: phishing user data or location, destroying tags,
+          modifying an NFC tag to cause indirect damage through fake identities
+          and attack vectors;
```
kenchris: Why semi colon?

index.html
```diff
@@ -4116,91 +4076,298 @@ <h3>Parsing content</h3>
...
+        <td>
+          Malicious tag may be involuntarily or voluntarily read by devices
+          and the data read may constitute an attack vector on the user agent.
+        </td>
+        <td>
+          This is a generic problem with all existing NFC tags.
+          The data is considered application specific.
+          Implementations need security hardening.
+          Involuntary touch is low probability due to short range and critical
+          angle for reading.
```
kenchris: plus our additional focus requirements

index.html
```diff
@@ -4116,91 +4076,298 @@ <h3>Parsing content</h3>
...
+          It’s easier to attack elsewhere (e.g. WiFi and Bluetooth).
```
kenchris: Not sure we want to mention that

index.html
```diff
@@ -4116,91 +4076,298 @@ <h3>Parsing content</h3>
...
+        <td>
+          Support for smart posters is not among the main use cases of Web
+          NFC; if and when supported, no automatic actions should be allowed,
+          the user must be made aware and given the ability to control what is
+          happening during the NFC communication.
+        </td>
+      </tr>
+      <tr>
+        <td>
+          Malicious Web NFC tag could send an arbitrary payload to a website
```
kenchris: I don't like send here... the web page itself reads and needs to be aware that data might need sanitation, especially if made non-inert

index.html
```diff
@@ -4116,91 +4076,298 @@ <h3>Parsing content</h3>
...
+          tag so the contents could later be verified.
+        </td>
+      </tr>
+      <tr>
+        <td>
+          A Web NFC tag could be used for leaking the user’s physical location,
+          <ul>
+            <li>
+              if the reading triggers a user’s device to navigate to a website
+              that can identify the tag and knows the tag's physical location.
```
kenchris: (if they know where that tag is location, ie serial number or encoded data)

index.html
```diff
@@ -4116,91 +4076,298 @@ <h3>Parsing content</h3>
...
+            </li>
+            <li>
+              by sending <a>NFC tag</a>s, either by directly sharing the user
```
kenchris: sending NFC tags??

index.html
```diff
@@ -4116,91 +4076,298 @@ <h3>Parsing content</h3>
...
+              location (when the <a>browsing context</a> has access to that),
+              or by sending information that can be used for inferring the
+              location of the user.
+            </li>
+          </ul>
+        </td>
+        <td>
+          A Web NFC tag SHOULD NOT ever trigger a user’s device to navigate
```
kenchris: remove Web

index.html
```diff
@@ -4116,91 +4076,298 @@ <h3>Parsing content</h3>
...
+              user prompts involved in obtaining permissions.
+              The signing tools and Public Key Infrastructure for key management
+              are out of scope of Web NFC.
+            </li>
+          </ul>
+        </div>
+      </section>
+    </section>
+
+    <section> <h3>Sign NDEF records</h3>
```
kenchris: Signed?

index.html
```diff
@@ -4116,91 +4076,298 @@ <h3>Parsing content</h3>
...
+      <p>
+        When pushing an <a>NDEF message</a>, one, more, or all records MAY be
+        signed using <a>NDEF signature</a>.
```
kenchris: using a

index.html
```diff
@@ -4116,91 +4076,298 @@ <h3>Parsing content</h3>
...
+      </p>
+      <p>
+        In order to build a trust chain, a Public Key Infrastructure is also
+        needed for key mananegent.
```
kenchris: spelling

index.html
zolkis: Hmm, okay... do we split it up as it currently may mean 3 things?
I guess the store/manifest part goes away and the end-to-end encryption and signing go elsewhere.

index.html

index.html

index.html
zolkis: Will remove this threat as it doesn't add specific enough information.

index.html
zolkis: Try again that sentence, I don't get the point.

index.html
zolkis: I remove that whole sentence because doesn't add to the previous point.

index.html
kenchris: You don't explain how they can identify the tag and use that to detect your physical location.

They can because tags can carry serial numbers or unique data that can be associated with the physical placement of the tag that was read

index.html

index.html
zolkis: Verb. Many mechanism titles use a verb. If we change, should be consistent.

index.html

index.html

index.html

index.html

zolkis: NFC data transfers would be better term in a security section. Used that.

index.html

index.html
zolkis: It's about authorization, not about misusing data. Reformulated, nevertheless.

index.html
zolkis: Reformulated

index.html
zolkis: This is a short formulation of _what_ the threat is; it would be besides the point to maintain _how_ a website can identify the tag - what is important is if it _can_ identify the tag and knows where the tag was left off (again, we don't need to explain in what ways can you maintain a tag's physical location, as there are quite many options).

index.html
```diff
@@ -4214,24 +4542,26 @@ <h3>Parsing content</h3>
...
         otherwise fail.
       </p>
       <p>
-        Setting up listeners for reading <a>NFC content</a> should
+        Setting up listeners for reading <a>NFC content</a> SHOULD
         <a>obtain permission</a>.
       </p>
       <p>
-        The process of reading an <a>NDEF message</a> does not need to
+        Pushing <a>NFC content</a> to an <a>NFC peer</a> SHOULD
```
ereshetova: Any reason why it is SHOULD and not MUST? If an NFC content can be pushed from peer to peer without user approval, this can be used for malware spread.

index.html
```diff
@@ -4242,39 +4572,57 @@ <h3>Parsing content</h3>
...
-        able to properly understand what they are granting.
-      </p-->
+
+    <section> <h4>Signing <a>NFC content</a></h4>
+      <div>
+        In this version of the specification the following policies are
+        recommended:
+        <ul>
+          <li>
+            A <a>smart poster</a> MAY be trusted only if signed by using
```
ereshetova, line 4494: Should we define explicitly that signature should cover whole NFC tag/poster in this case or only signed data should be considered trusted.

index.html
```diff
@@ -4116,91 +4060,475 @@ <h3>Parsing content</h3>
...
+        <dt><strong>Actors</strong></dt>
+        <dd>
+          Malicious MitM user.
+        </dd>
+        <dt><strong>Mitigation, comments</strong></dt>
+        <dd>
+          Use <a>NDEF signature</a>s to protect the NFC tag payload
+          and metadata. Additionally, harden the platform stack and
+          communication between the web NFC frontend and backend.
+          This thhreat is out of scope for Web NFC implementations.
```
ereshetova: typo "thhreat"

ereshetova:
Apart from comments above, I think everything is ok

index.html
zolkis: Peer push requires user gesture of putting 2 devices in range, i.e. one or two persons is always involved, within an arms reach, though not necessarily knowing of each other (metro malware spread scenario). If the pushing user is malicious, will just okay the permission anyway. The other end should accept the push, though. If the page is malicious, obtaining permission before pushing is a MUST indeed.

index.html
zolkis: The text says it should be signed with a single NDEF signature.
The question here is whether can we trust the smart poster if other records are also in the message. In that case we can trust only if it's the first record in the message or it's preceded by another signature record.

index.html
```diff
@@ -4116,91 +4060,475 @@ <h3>Parsing content</h3>
...
+          tags locally.
+          As a result can compromise integrity of user device, cause data
+          injection, redirect to malicious web page, phishing user location,
+          causing side actions such as installing applications, trigger
+          automated dispatching or other actions.
+        </li>
+        <li>
+          <strong>Malicious MitM user</strong>: any MitM (man in the middle)
+          style attack between Web NFC implementation and an <a>NFC adapter</a>
+          in a user device. A
```
beaufortfrancois: Typo: "A Also"

index.html
```diff
@@ -4116,91 +4060,475 @@ <h3>Parsing content</h3>
...
+    </p>
+    <section> <h3><dfn>Obtaining permission</dfn></h3>
+      Implementations SHOULD use one or more of the following mechanisms to
+      <dfn>obtain permission</dfn>.
+      <ul>
+        <li>
+          A mechanism for obtaining <a>expressed permission</a> of the user.
+        </li>
+        <li>
+          A mechanism for <a>ask for forgiveness</a> policies.
```
beaufortfrancois: Shall we be more conservative and add "ask for forgiveness" to the spec when we actually witness user agents that feel the need of having this kind of mechanism?

index.html
```diff
@@ -4116,91 +4060,475 @@ <h3>Parsing content</h3>
...
+    <section> <h3>Encrypting <a>NFC content</a></h3>
+      <p>
+        For trusting the confidentiality of the data exchanged via NFC, user agents
+        MAY use encrypted <a>NFC content</a> with key management based on
+        Public Key Infrastructure (PKI).
+        Key management is out of the scope of Web NFC.
+      </p>
+      <p class="note">
+        This is different from application level end to end encryption.
+        Key management is then in application domain.
```
beaufortfrancois: Can we be explicit and define what/who is application domain?

index.html
```diff
@@ -4116,91 +4060,475 @@ <h3>Parsing content</h3>
...
+      </section>
+      <section> <h4>Include URI records with registrable domain</h4>
+        <p>
+          When pushing an <a>NDEF message</a>, the [= host/registrable domain =],
+          <a data-cite="dom#concept-host-serializer">serialized</a>,
+          of the <a>current settings object</a> MAY be stored as a signed
+          <a>URI record</a>. This does not make other records trusted.
+        </p>
+      </section>
+    </section>
```
beaufortfrancois, line 4422: It seems to me that a "real" example for content identification would be more about storing a public key in a tag and checking its authenticity with a private key stored in the cloud?

beaufortfrancois (changes requested):
Thank you for all this work in polishing the security section of Web NFC!
See comments below.

index.html
zolkis: Or we change "recommended" mechanism to "possible"?

Or should I plainly remove (comment out) that mechanism?

index.html
zolkis: Actually the note does not need the sentence "Key management is then in application domain."

zolkis: This case applies to the implementation. I think it's not realistic to Web NFC to promise implementing content encryption with key management, so we could as well remove it.

However, the security section is a general informal section, and the question is should we _exclude_ the possibility any implementation goes with this?

index.html
zolkis: Yes, we could add that as a separate mechanism. 

This one is the remains of an old idea whose context has changed. Currently we don't do this, but does it mean we have to remove it from the Security section (again, this is a general informal section). The normative part is in the Security policies section.

index.html
beaufortfrancois: I would simply remove it that mechanism. What do you think @kenchris?

index.html
beaufortfrancois: I would not exclude the possibility but rather unclutter the security section?
Maybe add that to the security questionnaire for TAG.

index.html
beaufortfrancois: I'd prefer leaving it out as this would give (bad) ideas to people.

zolkis:
@ereshetova @beaufortfrancois new text is up.

zolkis:
In short:
- "ask for forgiveness" removed
- there is a single "obtain permission" algorithm
- security mechanisms that are not going to be used by Web NFC have been removed
- separate security mechanisms for implementations and applications
- separate policies for implementations (normative) and applications (recommended).

index.html
```diff
@@ -4116,91 +4036,332 @@ <h3>Parsing content</h3>
...
+
+  <section> <h3>Security policies</h3>
+    <p>
+      This section lists the normative security policies for implementations.
+    </p>
+
     <section> <h4>Secure Context</h4>
       <p>
         Only <a>secure contexts</a> are allowed to access <a>NFC content</a>.
         Browsers may ignore this rule for development purposes only.
```
beaufortfrancois: Nit: `s/may/MAY/`

index.html
```diff
@@ -4214,67 +4375,76 @@ <h3>Parsing content</h3>
...
+              Indication whether <a>NFC content</a> is trusted may be added in
+              future versions. Also, origin policies MAY be applied in order to
+              relax user prompting when a trust chain could be established.
+            </p>
+          </li>
+        </ul>
+      </div>
+    </section>
+    <p class="note">
+      For Bluetooth and WiFi handover (supported in later versions),
```
beaufortfrancois, line 4522: Shall we add this only when Bluetooth and WiFi handover are actually a thing?

index.html
```diff
@@ -4214,67 +4375,76 @@ <h3>Parsing content</h3>
...
+          </li>
+          <li>
+            An <a>NDEF message</a> MAY be trusted only if signed by
+            a single <a>NDEF signature</a> record by the same issuer.
+          </li>
+          <li>
+            User agents expose <a>NDEF signature</a> records without verifying
+            them. It is applications' responsibility to verify signatures and
+            to sign <a>NFC content</a>.
+            <p class="note">
```
beaufortfrancois: Do we have plans already to expose this?

index.html
```diff
@@ -4419,57 +4294,31 @@ <h3>Parsing content</h3>
...
-      </section>
-      <section> <h4><dfn>Ask for forgiveness</dfn></h4>
-        <p>
-          Denotes an unobtrusive way to notify the user about an ongoing operation
-          with a possible action to cancel or undo the operation, and to
-          handle future occurences of the same operation by the same action.
-        </p>
-      </section>
+      <p>
+        Implementations SHOULD use one or more of the following mechanisms to
```
beaufortfrancois: "one or more" does't make sens anymore since there's only one right?

index.html
```diff
@@ -4419,57 +4294,31 @@ <h3>Parsing content</h3>
     </section>
   </section>
 
-  <section class="informative"> <h2>Security mechanisms</h2>
+  <section class="informative"> <h2>Security mechanisms for implementations</h2>
     <p>
       In order to support security policies concerning Web NFC, the following
-      mechanisms are recommended for the implementations.
+      mechanisms are recommended for implementations.
```
beaufortfrancois: There's only one mechanism.

beaufortfrancois (commented):
Thank you @zolkis!

index.html
zolkis: OK, I comment it out.

index.html
zolkis: Yes

index.html
zolkis: Thanks for noticing. Fixed.

index.html
zolkis: Right. Actually I removed those introductory sentences, they don't add any value, just make it longer.

index.html
```diff
@@ -924,6 +888,9 @@ <h4>
       The user opens a web page which can write an <a>NFC tag</a>. The write
       operations may be one of the following:
       <ol>
+        <li>
+          Writing to a non-formatted <a>NFC tag</a>.
+        </li>
         <li>
           Writing to an empty <a>NFC tag</a>.
```
kenchris: empty, but formatted

index.html
```diff
@@ -4116,91 +4036,328 @@ <h3>Parsing content</h3>
...
+          against data disclosure and data modification.
+          This also includes Denial of Service attacks against a solution
+          deployed with NFC tags (e.g. a malicious actor destroying tags linked
+          to a solution).
+        </li>
+        <li>
+          <strong>User identity or other privacy sensitive attributes</strong>
+          that can be directly or indirectly determined by using Web NFC, by the
+          <a>NFC content</a> creator by a web site using Web NFC.
+          This data could be used directly or leaked forward to third parties. Examples are user location, device identifiers and user identifiers.
```
kenchris: wrap line

index.html
```diff
@@ -4116,91 +4036,328 @@ <h3>Parsing content</h3>
...
+        </li>
+        <li>
+          <strong>User data</strong> exposed to a web page using Web NFC.
+          While web page might collect user data using other means than
```
kenchris: while *a*

index.html
```diff
@@ -4116,91 +4036,328 @@ <h3>Parsing content</h3>
...
+          <strong>Malicious NFC tag creator</strong>: same as malicious web page
+          owner, but it has a possibility to create, delete or modify the NFC
+          tags locally.
+          As a result can compromise integrity of user device, cause data
+          injection, redirect to malicious web page, phishing user location,
+          causing side actions such as installing applications, trigger
+          automated dispatching or other actions.
+        </li>
+        <li>
+          <strong>Malicious MitM user</strong>: any MitM (man in the middle)
```
kenchris: Dont people write this all uppercase MITM?

index.html
```diff
@@ -4116,91 +4036,328 @@ <h3>Parsing content</h3>
...
+      </dl>
+    </section>
+
+    <section> <h4>NFC tag modification</h4>
+      <dl>
+        <dt><strong>Threat description</strong></dt>
+        <dd>
+          An NFC tag is being modified without user consent.
+          This might enable further attacks using a malicious tag
+          or can be a Denial of Service attack to make one or more tags
```
kenchris, line 4263: You write MITM as abbreviation but not DOS which is a much more common abbreviation

index.html
```diff
@@ -4116,91 +4036,328 @@ <h3>Parsing content</h3>
...
+          on a NFC tag) or in-transfer between Web NFC and the
+          <a>NFC adapter</a> are read by unauthorized parties.
+        </dd>
+        <dt><strong>Affected assets</strong></dt>
+        <dd>
+          Confidential NDEF message payload in-transfer and in-storage.
+        </dd>
+        <dt><strong>Actors</strong></dt>
+        <dd>
+          Malicious MitM user, malicious web page creator.
```
kenchris: I think I would just write these out

index.html
zolkis: Right. I replaced `MitM` with `man-in-the-middle` and (`MITM` in one occurrence), according to the spelling in https://en.wikipedia.org/wiki/Man-in-the-middle_attack

index.html
zolkis: Actually it's DoS for that one, DOS is the ancient OS fossil :) 
https://en.wikipedia.org/wiki/Denial-of-service_attack

index.html
zolkis: Yes, I did that.

kenchris (approved):

index.html
kenchris: Well I used it a lot, so am I ancient too? ;)

zolkis: Me too, I was self-sarcastic :)

kenchris:
Maybe you want to add (DoS) some places... It is almost more known than the spelled out version

zolkis:
> Maybe you want to add (DoS) some places... It is almost more known than the spelled out version

It's only in one place, but ok, I amended.

kenchris:
Wanna merge?

zolkis:
No, since @beaufortfrancois requested changes and needs to re-review.

beaufortfrancois (approved):

beaufortfrancois:
LGTM with @kenchris suggestions.

zolkis:
> LGTM with @kenchris suggestions.

Thanks. Already taken care of, please check.

index.html

index.html

index.html

PR #441, opened 2019-11-10
non-Mime record should have a null mediaType

Honry:
***
<a href="https://pr-preview.s3.amazonaws.com/Honry/web-nfc/pull/441.html" title="Last updated on Nov 10, 2019, 12:25 PM UTC (edd1020)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/441/6b0cdb8...Honry:edd1020.html" title="Last updated on Nov 10, 2019, 12:25 PM UTC (edd1020)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1914,35 +1914,35 @@
       <td>0</td>
       <td><i>unused</i></td>
       <td>"`empty`"</td>
-      <td>`undefined`</td>
+      <td>`null`</td>
     </tr>
     <tr>
       <td>[=Well-known type record=]</td>
       <td>1</td>
       <td>"`T`"</td>
       <td>"`text`"</td>
-      <td>`undefined`</td>
+      <td>`null`</td>
     </tr>
     <tr>
       <td>[=Well-known type record=]</td>
       <td>1</td>
       <td>"`U`"</td>
       <td>"`url`"</td>
-      <td>`undefined`</td>
+      <td>`null`</td>
     </tr>
     <tr>
       <td>[=Well-known type record=]</td>
       <td>1</td>
       <td>"`Sp`"</td>
       <td nowrap>"`smart-poster`"</td>
-      <td>`undefined`</td>
+      <td>`null`</td>
     </tr>
     <tr>
       <td>[=Local type=] record*</td>
       <td>1</td>
       <td>[=local type name=]</td>
       <td>[=local type name=]</td>
-      <td>`undefined`</td>
+      <td>`null`</td>
     </tr>
     <tr>
       <td>[=MIME type record=]</td>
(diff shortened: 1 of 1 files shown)


Honry:
@leonhsl, @beaufortfrancois, PTAL, thanks!

leonhsl:
lgtm, Thanks!

zolkis:
Looks good.

Out of curiosity, is there a particular reason you used `null` and not `undefined`?

I thought that when a variable or property is declared in WebIDL but not assigned in any algorithm in the spec, it should be `undefined` and not `null`.

At least on JS side, I find `undefined` more exact (in code), but both are falsy values and likely no one is going to be mislead to test the type of `mediaType` (which would be `object` if it was assigned  `null` and its `undefined` when `undefined`). 

I guess you use `null` because `mediaType` is nullable in WebIDL and you can use `nullptr` in C++.
However, I wonder why exactly are these (`mediaType`, `encoding` and `lang`) nullable in WebIDL, when they are the properties that are used only in certain cases depending on `recordType`, so they should be `undefined` when their use was not meant for.

Nevertheless, these are all strings and it makes sense to use `nullptr` for them in C++, so I guess it's easier for implementations if these are nullable in WebIDL and initialized to `null` in the algorithms?

No strong opinion here but I am interested if there is any other particular implementation reason for explicitly using `null`.

Honry:
@reillyeon, could you help clarify it from implementation perspective?

leonhsl:
I don't think it's an implementation issue.

I understand the attributes of an interface are always **defined**, they're accessed via some internal getter/setter functions of the interface.

It's different from the case for the properties of a `dictionary`, which you can define/undefine arbitrarily at any time.

And I also feel it's unnecessary to explicitly give a `null` to the `mediaType` attribute anywhere, whose initial value is just `null` due to the `?` suffixed to the type.

zolkis:
That depends whether the property is nullable in WebIDL or not. [AFAIK](https://www.chromium.org/blink/webidl#TOC-undefined-and-null) if a property is `undefined` in JS, you need to throw exception, there is no getter or setter. 

(Again, I am not blocking this change, I am just curious.)

leonhsl:
Nullable or not is not related to the discussion here around the `undefined` thing.

From what I understand, if you try to access an attribute that is not defined in the interface (of course there is neither getter nor setter), you'll just get an exception thrown. (actually in implementation we do not worry about this because V8 binding layer takes the care)
But you'll never get `undefined` from an attribute already defined in the WebIDL.

And, if you try to get a property that is not yet set in a dictionary, you just get `undefined`.
(JS interface and dictionary have much difference)

zolkis:
What I question is why do we use nullable properties in WebIDL for `mediaType`, `encoding` and `lang`. Since they are used only with certain `recordType` values, they should be `undefined` in invalid cases, not just `null`.

If the WebIDL was with non-nullable properties for `mediaType`, `encoding` and `lang`, then in the spec `undefined` should be used in algorithms instead of `null`, and in C++ the getter should throw.

I wondered was `null` preferred because the way the bindings work, or implementation.
Based on what you say looks like because of the bindings.

Anyway, if `null` works for everyone, let's move on with that. And yes, I agree with Leon that algorithms don't need to have explicit statements for `null` initialization of nullable properties. But it adds to the clarity in this case, so we can keep them from my part.

leonhsl:
Can we merge this?

beaufortfrancois:
My 2 cents as I don't feel strongly about it.
If we go with `null`, I'd like to make sure our algorithms explicitly set properties to `null` as it is not the default value. This would make the spec much verbose without adding that much value in my opinion though.

gentle ping @reillyeon

kenchris:
Maybe worth filing an issue here as well: https://github.com/w3ctag/design-principles/issues

reillyeon (approved):

reillyeon:
For interface types I believe that `null` is a more correct falsey value than `undefined`, which is used for dictionaries.

zolkis (approved):

PR #444, opened 2019-11-18
Fix problems in "Create NDEF message"

### LINKED ISSUE #443: Problems in '9.10.1 Creating NDEF message'
https://w3c.github.io/web-nfc/#creating-ndef-message
  1.  lacks descriptions about how to wrap a `DOMString` or `BufferSource` to create an `NDEFMessage`.
  2. 4.3 description is wrong:

> If message is of type NDEFMessageInit and its id is not null:
  
  3. It seems to me the spec wants to say `NDEFRecord#id` is null when `NDEFRecordInit#id` is not provided, if so, please change `NDEFRecord` interface definition to reflect it, i.e. add '?' suffix.

beaufortfrancois:
As raised by @leonhsl, this PR fixes issues in "Create NDEF message":
1. Handle `DOMString` and `BufferSource` to create an NDEFMessage.
2. Use NDEFRecord instead of NDEFMessage for `id`.

FIX #443

***
<a href="https://pr-preview.s3.amazonaws.com/beaufortfrancois/web-nfc/pull/444.html" title="Last updated on Nov 19, 2019, 9:52 AM UTC (1bcd82b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/444/6b0cdb8...beaufortfrancois:1bcd82b.html" title="Last updated on Nov 19, 2019, 9:52 AM UTC (1bcd82b)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2706,8 +2706,10 @@
                   |recordType| set to "`text`" and |data| set to |message|.
                 </li>
                 <li>
-                  Set |message|'s records to a list of records that contain only
-                  |textRecord|.
+                  Let |records| be the list « |textRecord| ».
+                </li>
+                <li>
+                  Set |message|'s records to |records|.
                 </li>
               </ul>
               <dt>{{BufferSource}}</dt>
@@ -2718,8 +2720,10 @@
                     |mediaType| set to "`application/octet-stream`".
                   </li>
                   <li>
-                    Set |message|'s records to a list of records that contain only
-                    |mimeRecord|.
+                    Let |records| be the list « |mimeRecord| ».
+                  </li>
+                  <li>
+                    Set |message|'s records to |records|.
                   </li>
               </ul>
               <dt>{{NDEFMessageInit}}</dt>
--- a/index.html
+++ b/index.html
@@ -2703,7 +2703,7 @@
               <ul>
                 <li>
                   Let |textRecord| be an <a>NDEFrecord</a> initialized with its
-                  |recordType| set to "`text/plain`" and |data| set to |message|.
+                  |recordType| set to "`text`" and |data| set to |message|.
                 </li>
                 <li>
                   Set |message|'s records to a list of records that contain only
@@ -2826,7 +2826,7 @@
                 </dl>
               </li>
               <li>
-                If |record|'s |id| is not `null`:
+                If |record|'s |id| is not `undefined`:
                 <ul>
                   <li>
                     Set |identifier| to the result of invoking
--- a/index.html
+++ b/index.html
@@ -1530,7 +1530,7 @@
 
         readonly attribute USVString recordType;
         readonly attribute USVString? mediaType;
-        readonly attribute USVString id;
+        readonly attribute USVString? id;
         readonly attribute DataView? data;
 
         readonly attribute USVString? encoding;
@@ -2697,15 +2697,46 @@
           |message:NDEFMessageSource| run these steps:
         </p>
         <ol class=algorithm id="create-web-nfc-message">
-          <li>
-            If the |message:NDEFMessageSource| parameter is not of type defined
-            by the <a>NDEFMessageSource</a> union, throw a {{TypeError}} and
-            abort these steps.
-          </li>
-          <li>
-            If the |message:NDEFMessageSource| parameter is of
-            <a>NDEFMessageInit</a> type, and |message|'s records
-            [= list/is empty =], throw a {{TypeError}} and abort these steps.
+          <li>Switch on |message:NDEFMessageSource|'s type:
+            <dl>
+              <dt>{{DOMString}}</dt>
+              <ul>
+                <li>
+                  Let |textRecord| be an <a>NDEFrecord</a> initialized with its
+                  |recordType| set to "`text/plain`" and |data| set to |message|.
+                </li>
+                <li>
+                  Set |message|'s records to a list of records that contain only
+                  |textRecord|.
+                </li>
+              </ul>
+              <dt>{{BufferSource}}</dt>
+              <ul>
+                  <li>
+                    Let |mimeRecord| be an <a>NDEFrecord</a> initialized with its
+                    |recordType| set to "`mime`", |data| set to |message|, and
+                    |mediaType| set to "`application/octet-stream`".
+                  </li>
+                  <li>
+                    Set |message|'s records to a list of records that contain only
+                    |mimeRecord|.
+                  </li>
+              </ul>
+              <dt>{{NDEFMessageInit}}</dt>
+              <ul>
+                <li>
+                  If |message|'s records [= list/is empty =], throw a
+                  {{TypeError}} and abort these steps.
+                </li>
+              </ul>
+              <dt>unmatched type</dt>
+              <ul>
+                <li>
+                  [= exception/throw =] a {{TypeError}} and abort these steps.
+                </li>
+              </ul>
+            </dl>
+            </dl>
           </li>
           <li>
             Let |output| be the notation for the <a>NDEF message</a>
@@ -2795,8 +2826,7 @@
                 </dl>
               </li>
               <li>
-                If |message| is of type {{NDEFMessageInit}} and its
-                |id:string| is not `null`:
+                If |record|'s |id| is not `null`:
                 <ul>
                   <li>
                     Set |identifier| to the result of invoking


zolkis (approved):

index.html
```diff
@@ -2697,15 +2697,46 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
-            If the |message:NDEFMessageSource| parameter is of
-            <a>NDEFMessageInit</a> type, and |message|'s records
-            [= list/is empty =], throw a {{TypeError}} and abort these steps.
+          <li>Switch on |message:NDEFMessageSource|'s type:
+            <dl>
+              <dt>{{DOMString}}</dt>
+              <ul>
+                <li>
+                  Let |textRecord| be an <a>NDEFrecord</a> initialized with its
+                  |recordType| set to "`text/plain`" and |data| set to |message|.
```
leonhsl: `text/plain`  --> `text`

index.html
```diff
@@ -2795,8 +2826,7 @@ <h3><dfn>Writing or pushing content</dfn></h3>
                 </dl>
               </li>
               <li>
-                If |message| is of type {{NDEFMessageInit}} and its
-                |id:string| is not `null`:
+                If |record|'s |id| is not `null`:
```
leonhsl: We'd better say _If |record|'s |id| is not `undefined`:_? Given that `NDEFRecordInit` is a dictionary and `id` is non-nullable.

leonhsl: Also, for the `undefined` case, we might want to set `NDEFRecord#id` to the current origin (like https://example.com/) rather than doing nothing? Then later when we try to filter records with `NDEFScanOptions#id` being https://example.com/, this record can be found out.

leonhsl (commented):
Thank you very much for doing this PR!

index.html
beaufortfrancois: 🤦‍♂ thank you!

index.html
beaufortfrancois: You're absolutely right.

beaufortfrancois: I'm actually not sure we want to do this.
@kenchris and @zolkis What do you think?
If web app doesn't specify any `id`, should it write `https://example.com/` in the record id?

My main concerns is that I'm not sure web developers would expect this behavior when they simply call `writer.push('foo')` as this would make it harder to guess the size of the NDEFMessage to push.

zolkis: Currently we don't any more use origin policy (Web NFC record and id), so I don't see any point in setting `id` to origin. It should be application's decision.

leonhsl: By reading the example at https://w3c.github.io/web-nfc/#save-and-restore-game-progress-with-another-device, it implies that if `id` is specified as `/mypath/mygame/update`, we write it as `https://mygame.com/mypath/mygame`. But if `id` is not specified we just skip writing it.
The logic seems not that tight.

index.html
```diff
@@ -2697,15 +2697,46 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
+          <li>Switch on |message:NDEFMessageSource|'s type:
+            <dl>
+              <dt>{{DOMString}}</dt>
+              <ul>
+                <li>
+                  Let |textRecord| be an <a>NDEFrecord</a> initialized with its
+                  |recordType| set to "`text`" and |data| set to |message|.
+                </li>
+                <li>
+                  Set |message|'s records to a list of records that contain only
```
kenchris: Why can't there be multiple records? This way of writing this also sounds weird, took me a while to understand what you were trying to do

index.html
```diff
@@ -2697,15 +2697,46 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
+              </ul>
+              <dt>{{BufferSource}}</dt>
+              <ul>
+                  <li>
+                    Let |mimeRecord| be an <a>NDEFrecord</a> initialized with its
+                    |recordType| set to "`mime`", |data| set to |message|, and
+                    |mediaType| set to "`application/octet-stream`".
+                  </li>
+                  <li>
+                    Set |message|'s records to a list of records that contain only
```
kenchris: im sure the infra spec has algos (you can refer to) on how to append to a list

index.html
leonhsl: And, the `urlRecord` naming in https://w3c.github.io/web-nfc/#creating-a-record-identifier looks like some leftover from the old era. I'm not sure, are we already done refining this algorithm and it's complying with the example code mentioned above?

index.html
kenchris: https://infra.spec.whatwg.org/#list-append

index.html
zolkis: Yes, we should clean them out. 
The example can still contain an id, actually it is nice to have example with id, since now it is in application domain.

beaufortfrancois: @zolkis Can you share what the example should look like?

index.html
beaufortfrancois: This prose is only for `writer.push("foo")` case. So only one text record if I'm not mistaken.

index.html
beaufortfrancois: Would this be more appropriate?

```
Let |records| be the list « |mimeRecord| ». 
Set |message|'s records to |records|.
```

kenchris: I think that is more clear, yes

beaufortfrancois: Fixed in https://github.com/w3c/web-nfc/pull/444/commits/1bcd82bcc77639f1775a5b2e6d9baafb018e7c98

kenchris (approved):

index.html
beaufortfrancois: @kenchris I guess we're good then?

beaufortfrancois:
@kenchris https://github.com/w3c/web-nfc/pull/444#discussion_r347818532 is still not resolved though

index.html
kenchris: id: document.location.origin ?

I am fine with us not setting this by default

zolkis:
[Example 7](https://w3c.github.io/web-nfc/#example-7) is fine. My comment meant that it makes sense to keep the `id` example even if we remove the idea of storing origin in `id`.

index.html
leonhsl: My point is that we should give a meaningful value to `NDEFRecord#id` if it's not specified by NDEFRecordInit, whether the value should be the origin or not does not matter.

beaufortfrancois: I'm not sure why `NDEFRecord#id` should have a meaningful value if web dev didn't specify one. @leonhsl Can you explain? Sorry for being slow.

leonhsl: If we leave `NDEFRecord#id` undefined, it means we will not write any id information into a nfc tag for this record.
Then, later `scan({id: 'https://example.com'})` will NOT match it, which I think is **incorrect** behavior, because callers expect to see all records written by pages in https://example.com. 
While the record with `/mypath/myid` gets its id transformed to `https://example.com/mypath/myid` and will hit the match.

leonhsl: Actually our id generation mechanism is already implying that the record id written by WebNFC API is bound inherently with the current web page, so I suggest to push through this rule everywhere.

We've already been doing some internal transformation over the ids (binding them to the current host), so I don't think generating `https://example.com` for an `undefined `id will take users by surprise.

zolkis: Right, there is a mismatch.
We have canceled the idea to write Web NFC id (first as separate id record, then as record id), but we did not clean that up completely. Now we need to decide what we want: do we keep it completely, or do we remove it completely, but now it's hanging between two ideas, indeed.

If we take removing this "additional smartness" consistently, then we should fix (or remove) the id generation algorithm. It should be something the application explicitly wants. We could include some convenience (like a boolean flag whether to include it), but apps can initialize `id` themselves, for instance it could be a public key or a registrable domain etc.

beaufortfrancois: Shall we open a separate issue for the Web NFC id?
I'd be in favor of removing this "id" constraint algorithm and let web developers do what ever they want with it.

```js
await writer.push({ records: [{ recordType: 'text', data: 'hi', id: '123' }] });
...
await reader.scan({id: '123' });
reader.onreading = event => {};
```

zolkis: > I'd be in favor of removing this "id" constraint algorithm and let web developers do what ever they want with it.

I agree. Let's track this in an issue.

leonhsl: Let's move forward at the issue https://github.com/w3c/web-nfc/issues/445.

PR #446, opened 2019-11-25
Remove URL constraint for record id

### LINKED ISSUE #445: Keep (how?) or remove the record id?
This issue comes from the discussion at https://github.com/w3c/web-nfc/pull/444#discussion_r349416140

beaufortfrancois:
As discussed in https://github.com/w3c/web-nfc/pull/444#discussion_r349416140, this PR removes binding record ids with the hostname of the current web page.

FIX #445

***
<a href="https://pr-preview.s3.amazonaws.com/beaufortfrancois/web-nfc/pull/446.html" title="Last updated on Nov 25, 2019, 9:03 AM UTC (a78d5fb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/446/f28bf10...beaufortfrancois:a78d5fb.html" title="Last updated on Nov 25, 2019, 9:03 AM UTC (a78d5fb)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -896,13 +896,7 @@
         </li>
         <li>
           Writing to an <a>NFC tag</a> which already contains a
-          <a>NDEF message</a> with a different <a>record identifier</a>
-          (i.e. overwriting a web-specific tag).
-        </li>
-        <li>
-          Writing to an <a>NFC tag</a> which already contains a
-          <a>NDEF message</a> with the same <a>record identifier</a>
-          (i.e. updating own tag).
+          <a>NDEF message</a>.
         </li>
         <li>
           Writing to other, writable <a>NFC tag</a>s (i.e. overwriting a
@@ -1162,10 +1156,9 @@
   <section> <h3>Save and restore game progress with another device</h3>
     <p>
       Filtering of relevant data sources can be done by the use of
-      the <a>NDEFScanOptions</a>. Below we accept the record identifier URL with
-      "`/mypath/mygame/`" in its path from "`mygame.com`"
-      domain and its subdomains. When we read the data, we immediately update
-      the game progress by issuing a push with a custom NDEF data layout.
+      the <a>NDEFScanOptions</a>. Below we use a custom record identifier so
+      that when we read the data, we immediately update the game progress by
+      issuing a push with a custom NDEF data layout.
     </p>
     <p>
       The example allows reading and pushing to both peers and tags,
@@ -1173,14 +1166,14 @@
     </p>
     <pre class="example">
       const reader = new NDEFReader();
-      await reader.scan({ id: "https://mygame.com/mypath/mygame" });
+      await reader.scan({ id: "my-game-progress" });
       reader.onreading = async event => {
         console.log(`Game state: ${ JSON.stringify(event.message.records) }`);
 
         const encoder = new TextEncoder();
         const newMessage = {
           records: [{
-            id: "/mypath/mygame/update",
+            id: "my-game-progress",
             recordType: "mime",
             mediaType: "application/json",
             data: encoder.encode(JSON.stringify({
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1162,25 +1156,24 @@ <h4>
...
+      issuing a push with a custom NDEF data layout.
     </p>
     <p>
       The example allows reading and pushing to both peers and tags,
       whichever one is tapped first.
     </p>
     <pre class="example">
       const reader = new NDEFReader();
-      await reader.scan({ id: "https://mygame.com/mypath/mygame" });
+      await reader.scan({ id: "my-game-progress" });
```
zolkis, line 1170: It is not excluded someone could use a URL as id, though.

kenchris, line 1170: according to NDEF spec ids are supposed to be urls

index.html
```diff
@@ -3327,111 +3249,6 @@ <h3><dfn>Writing or pushing content</dfn></h3>
         "`*/*json`", on the other hand, matches both.
       </div>
     </section>
-    <section> <h3>URL patterns</h3>
```
zolkis, line 3330: Hmm. A lot of work is gone... :) 
Not sure it couldn't be useful at some point, but for now yes, it's to be removed.

index.html
kenchris: Screenshot 2019-11-25 at 09 31 24

index.html
```diff
@@ -3640,8 +3451,7 @@ <h3><dfn>Writing or pushing content</dfn></h3>
         <ol>
           <li>
             If |reader|.<a>[[\Id]]</a> is [= dictionary member/present =]
-            and doesn't <a href="#dfn-match-record-identifier-with-url-pattern">match</a>
-            any |record:NDEFRecord|'s id where |record|
+            and it is not equal to any |record|'s id where |record|
```
zolkis, line 3455: Fair enough, an id should be compared for equality, not for partial match. Good.

zolkis (approved):
Thanks!

index.html
kenchris: I am not so certain. For people using URLs being able to use * is quite handy

index.html
kenchris: semanticly it is an url (can be relative so you can say the examples are) but we should at least note that in a note or something

zolkis: Right, I forgot about that.

beaufortfrancois: I've edited example description at https://github.com/w3c/web-nfc/pull/446/commits/7c2d2c10fb5cfa6e4f2606d56544ecd91bc68373

kenchris: Where we introduce the id field, are we making clear that it is a url?

kenchris: updated example looks good

beaufortfrancois: I've edited example description at 7c2d2c1

index.html
beaufortfrancois: How about waiting for web developers feedback in origin trials?
We can easily add it if people need it.

index.html
beaufortfrancois: > Where we introduce the id field, are we making clear that it is a url?

I've updated `id` introduction at https://github.com/w3c/web-nfc/pull/446/commits/a78d5fbea5713d7521c7ba15f3560af124468575

kenchris (approved):

index.html
zolkis: Sounds good. But how will people know there would be a possibility to use URL patterns and what are they? A note perhaps?

index.html
zolkis: If we include relative URLs, what they are relative to, and how does the implementation add the base URL when the record is written?

leonhsl: The implementation does not add any base URL but just writes the relative URL as is.

leonhsl (commented):
lgtm

PR #463, opened 2019-12-12
s/undefined/null/g for |data| when empty payload

### LINKED ISSUE #462: NDEFRecord.data should return null rather than undefined
In following algorithms,
> 9.11.9.6 Parsing NDEF MIME type records
9.11.9.8 Parsing NDEF external type records
9.11.9.9 Parsing NDEF unknown type records

Spec says:
>  Let buffer be the byte sequence of ndefRecords's PAYLOAD field if that exists, or otherwise undefined.
Set record's data to buffer.

But per discussion in https://github.com/w3c/web-nfc/pull/441,  `null` is a more correct falsey value than `undefined` for an interface type.

Besides, setter of record's data is missing in empty record type's parser algorithm.

beaufortfrancois:
FIX #462

***
<a href="https://pr-preview.s3.amazonaws.com/beaufortfrancois/web-nfc/pull/463.html" title="Last updated on Dec 12, 2019, 8:28 AM UTC (e89b33a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/463/27f97a9...beaufortfrancois:e89b33a.html" title="Last updated on Dec 12, 2019, 8:28 AM UTC (e89b33a)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3958,7 +3958,7 @@
         <li>
           Let |buffer:byte sequence| be the <a>byte sequence</a> of
           |ndefRecords|'s <a>PAYLOAD field</a> if that exists, or otherwise
-          `undefined`.
+          `null`.
         </li>
         <li>
           Set |record|'s <a>data</a> to |buffer|.
@@ -4010,7 +4010,7 @@
         <li>
           Let |buffer:byte sequence| be the <a>byte sequence</a> of
           |ndefRecords|'s <a>PAYLOAD field</a> if that exists, or otherwise
-          `undefined`.
+          `null`.
         </li>
         <li>
           Set |record|'s <a>data</a> to |buffer|.
@@ -4036,7 +4036,7 @@
         <li>
           Let |buffer:byte sequence| be the <a>byte sequence</a> of
           |ndefRecords|'s <a>PAYLOAD field</a> if that exists, or otherwise
-          `undefined`.
+          `null`.
         </li>
         <li>
           Set |record|'s <a>data</a> to |buffer|.
--- a/index.html
+++ b/index.html
@@ -3717,6 +3717,9 @@
           <li>
             Set |record|'s <a>mediaType</a> to `null`.
           </li>
+          <li>
+            Set |record|'s <a>data</a> to `null`.
+          </li>
         </ol>
       </li>
       <li>


Honry:
Setter of record's data is missing for empty record type's parser algorithm.

beaufortfrancois:
> Setter of record's data is missing for empty record type's parser algorithm.

Good catch!

Honry:
@beaufortfrancois, one more question, is the empty record type's id an empty string or null?

zolkis (approved):

beaufortfrancois:
According to the current spec text below, it is not clear to be honest.

> To parse an NDEF record given ndef into a record, run these steps:
> 1. **Set record's id to ndef's id.**
> 2. Set record's lang to null.
> 3. Set record's encoding to null.
> 4. If ndef's typeNameField (TNF field) is 0 (empty record):
> - 1. Set record's recordType to "empty".
> - 2. Set record's mediaType to null.

@zolkis @kenchris Do you have opinions?
I'll open a separate PR for this if we want to update spec if that's ok with you.

@Honry Does the current PR looks good to you?

Honry:
Per `9.11.9.1 Parsing records from bytes`: 
> If idLength > 0, let ndef's id be result of running UTF-8 decode on the next idLength (ID field) bytes, or else let ndef's id be the empty string.

It should be empty string.

> @Honry Does the current PR looks good to you?

Thanks for your quick fix, LGTM. :)

Honry (approved):

beaufortfrancois:
> Parsing records from bytes



> Per `9.11.9.1 Parsing records from bytes`:
> 
> > If idLength > 0, let ndef's id be result of running UTF-8 decode on the next idLength (ID field) bytes, or else let ndef's id be the empty string.
> 
> It should be empty string.

NFC spec says `The value 0x00 (Empty) indicates that there is no type or payload associated with this NDEF record. When used, the TYPE_LENGTH, ID_LENGTH and PAYLOAD_LENGTH fields are zero and the TYPE, ID and PAYLOAD fields are omitted from the NDEF record.`

So I guess I would say `null` instead of empty string.
Let 's hear what @zolkis and @kenchris think as well.

Honry:
@beaufortfrancois, thanks for the clarification, I will create a specific issue to discuss it further.

Honry:
Open https://github.com/w3c/web-nfc/issues/465.

zolkis:
When an NFC field is not present, I would map it to `undefined`, or when that is not desired or possible, then `null`.

PR #467, opened 2019-12-12
Update smart-poster example

beaufortfrancois:
@zolkis What do you think of this updated smart-poster example?
Main differences are:
- use of `find` instead of `for loop + continue` to reduce indentation
- use record.encoding to demonstrate UTF-16 potential encoding
- use `:smart-poster` instead of `:sp` as it's not clear what it means for newcomers.

***
<a href="https://pr-preview.s3.amazonaws.com/beaufortfrancois/web-nfc/pull/467.html" title="Last updated on Dec 12, 2019, 3:22 PM UTC (d7385c5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/467/ce9fd84...beaufortfrancois:d7385c5.html" title="Last updated on Dec 12, 2019, 3:22 PM UTC (d7385c5)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1369,25 +1369,20 @@
     </p>
     <pre class="example">
       const reader = new NDEFReader();
-      await reader.scan({ recordType: "example.com:sp" });
+      await reader.scan({ recordType: "example.com:smart-poster" });
       reader.onreading = event => {
-        const socialPost = event.message.records[0];
-        if (!socialPost) {
-          return;
-        }
+        const smartPosterRecord = event.message.records.find(
+          record => record.type == "example.com:smart-poster"
+        );
 
-        let action;
-        let text = "";
+        let action, text;
 
-        const decoder = new TextDecoder();
-        for (let record of socialPost.toRecords()) {
-          switch (record.recordType) {
-            case "text":
-              text = decoder.decode(record.data);
-              break;
-            case "act":
-              action = record.data.getUint8(0);
-              break;
+        for (const record of smartPosterRecord.toRecords()) {
+          if (record.recordType == "text") {
+            const decoder = new TextDecoder(record.encoding);
+            text = decoder.decode(record.data);
+          } else if (record.recordType == "act") {
+            action = record.data.getUint8(0);
           }
         }
 


index.html
```diff
@@ -1369,25 +1369,20 @@ <h4>
     </p>
     <pre class="example">
       const reader = new NDEFReader();
-      await reader.scan({ recordType: "example.com:sp" });
+      await reader.scan({ recordType: "example.com:smart-poster" });
```
zolkis, line 1372: Just as a note. Here we also have the choice to say `":smart-poster"`. According to the spec, `example.com` needs to match the the effective origin (though the algorithm is not there yet). If the app omits it, then the implementation will add it.

It would be important we get this right, so this policy needs discussion. If we want this, we need to add the origin-matching algorithm.

zolkis (approved):
Thanks, this looks nice!
This is an example for both external type and smart poster (embedded in the external type).
Please note the policy needs discussion, and depending on that, the external type validation needs to be updated.

kenchris (approved):

index.html
beaufortfrancois: Thinking about it again, I don't think Web NFC should be smart and use the effective domain of the origin as domain name in external type names when creating/reading external type records. 

I would argue web developers should always set the full external type record. Think about staging environments, where `staging.example.com:smart-poster` would be written automatically if `:smart-poster` were to be allowed. It would not match when read in a prod environment.
Does this make sense?

kenchris: I dont like the smartness either...

zolkis: Yes, the spec would be simpler that way.
So developers could give any 
a) free string we don't validate, only for the last part `:` + `<type-name>`
b) string that we validate to a domain name (we need an algorithm for it, not trivial)
c) a string that needs to validate to domain name that is part of the effective domain (algorithm needed).

Which option should we choose? (add more if you have ideas)

kenchris: Rely on URL spec?

kenchris: image

zolkis: So that means b + a. Fine for me.

beaufortfrancois: @zolkis Can you file an issue to track this?

PR #474, opened 2019-12-13
Fix #470: remove adding effective domain

### LINKED ISSUE #470: Simplify creating external records
Following the discussion [here](https://github.com/w3c/web-nfc/pull/467#discussion_r357220894).

@beaufortfrancois: 
> Thinking about it again, I don't think Web NFC should be smart and use the effective domain of the origin as domain name in external type names when creating/reading external type records.
I would argue web developers should always set the full external type record. Think about staging environments, where `staging.example.com:smart-poster` would be written automatically if `:smart-poster` were to be allowed. It would not match when read in a prod environment.

So the proposal is to validate input for external type as follows:
- find the last occurrence of `:`; 
- the part after it should validate as an external type
- the part before it has to be a domain name that should be validated by the [valid domain](https://url.spec.whatwg.org/#valid-domain) steps from the URL Standard (thanks @kenchris for the pointer).

zolkis:

***
<a href="https://pr-preview.s3.amazonaws.com/zolkis/web-nfc/pull/474.html" title="Last updated on Dec 16, 2019, 1:50 PM UTC (89b3905)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/474/4438fc8...zolkis:89b3905.html" title="Last updated on Dec 16, 2019, 1:50 PM UTC (89b3905)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2848,12 +2848,17 @@
           length exceeds 255 bytes, return `false`.
         </li>
         <li>
-          Let  |colonIndex| be the last occurrence of `U+003A` (`:`) in |input|.
-          If a `U+003A` (`:`) is not found, return `false`.
+          Let |domain| be the |input| from the start of |input| up to but
+          excluding the last occurrence of U+003A (`:`), or `null` if that is
+          not found.
         </li>
         <li>
-          Let |domain| be the part of |input| preceding |colonIndex|,
-          and |type| be the part following |colonIndex|.
+          Let |type| be the |input| after the last occurrence, if any, of
+          `U+003A` (`:`) up to the end of |input|, or `null` if that is
+          not found.
+        </li>
+        <li>
+          If |domain| or |type| is `null`, return `false`.
         </li>
         <li>
           Run the following sub-steps of
--- a/index.html
+++ b/index.html
@@ -1696,10 +1696,9 @@
         typically used in a record of an <a>NDEFMessage</a> that is the payload
         of a parent <a>NDEFRecord</a>, for instance in a <a>smart poster</a>.
         The context of the <a>local type</a> is the parent record whose payload
-        is the <a>NDEFMessage</a> to which this record belongs. The value
-        MUST NOT be equal to any other <a>record type</a>s defined
-        in this API, including other [=local type names=] and
-        [=external type names=].
+        is the <a>NDEFMessage</a> to which this record belongs and the
+        <a>local type name</a> SHOULD NOT conflict with any other type names
+        used in that context.
       </p>
       <p>
         Any implementation of Web NFC MUST transparently expose chunked records
@@ -2909,6 +2908,12 @@
         <li>
           If |input| does not start with a lowercase character or a number,
           return `false`.
+        </li>
+        <li>
+          If |input| is equal to any other <a>record type</a>s defined
+          in this API, including other [=local type names=] or
+          [=external type names=], return `false`.
+        </li>
         <li>
           Return `true`.
         </li>
--- a/index.html
+++ b/index.html
@@ -768,9 +768,9 @@
     </p>
     <p>
       The <dfn>external type</dfn> is a URN with the prefix `"urn:nfc:ext:"` followed by
-      the name of the owner domain, adding a colon, then a non-zero type name,
-      for instance `"urn:nfc:ext:w3.org:atype"`, stored as `"w3.org:atype"`
-      in the <a>TYPE field</a>.
+      the name of the owner domain, adding a `U+003A` (`:`), then a non-zero
+      type name, for instance `"urn:nfc:ext:w3.org:atype"`, stored as
+      `"w3.org:atype"` in the <a>TYPE field</a>.
     </p>
     <div>
       <ndef-record
@@ -1629,7 +1629,8 @@
         <a>recordType</a> attribute.
       </li>
       <li>
-        If the |recordType| value is "`smart-poster`", or an [=external type name=],
+        If the |recordType| value is "`smart-poster`", or if running
+        <a>validate external type</a> on |recordType| returns `true`,
         then return the result of running <a>parse records from bytes</a> on |bytes|.
         Re-[= exception/throw =] any exceptions.
       </li>
@@ -1648,8 +1649,8 @@
         <a>NDEF record</a> types.
       </p>
       <p>
-        A set of known standardized values exists, but it is also possible
-        for organizations to create their own custom [=external type names=].
+        A standardized <dfn>well known type name</dfn> can be one of the
+        following:
       </p>
       <dl>
         <dt>The "<dfn>empty</dfn>" string</dt>
@@ -1681,45 +1682,25 @@
         <dd>
           The value representing an <a>unknown record</a>.
         </dd>
-        <dt>An <dfn>external type name</dfn></dt>
-        <dd>
-          A {{DOMString}} representing a custom type for the
-          <a>external type record</a>.
-          The type must follow the [=external type name=] ABNF.
-          <pre class="abnf">
-            ext-type             = reg-name ":" custom-type
-            custom-type          = 1*(ALPHA / DIGIT / other)
-
-            DIGIT                = %x30-39
-            ALPHA                = %x41-5A / %x61-7A   ; A-Z / a-z
-            other                = "(" / ")" / "+" / "," / "-" / ":" / "=" /
-                                   "@" / ";" / "$" / "_" / "!" / "*" / "'" / "."
-          </pre>
-          The `reg-name` value is a [=host/registrable domain=] owned by the issuing
-          organization, a "`:`" and a type, e.g. "`w3.org:member`".
-          And additional ABNF exists for [=well-known type records=]:
-          <pre class="abnf">
-            wkt-type             = (ALPHA / DIGIT) *(ALPHA / DIGIT / other)
-          </pre>
-          <p class=note>
-            The [[[NFC-RTD]]] defines every type in the [=well-known type records=]
-            and <a>external type records</a> in terms of URNs, but only a subset of
-            the URN is actually stored in the <a>NDEF record</a>'s <a>TYPE field</a>,
-            which corresponds to the above two ABNFs.
-          </p>
-        </dd>
-        <dt>A <dfn>local type name</dfn></dt>
-        <dd>
-          A {{DOMString}} that MUST start with lowercase character or a number,
-          representing a type for a NFC Forum [=local type=],
-          typically used in a record of an <a>NDEFMessage</a> that is the payload
-          of a parent <a>NDEFRecord</a>, for instance in a <a>smart poster</a>.
-          The context of the <a>local type</a> is the parent record whose payload
-          is the <a>NDEFMessage</a> to which this record belongs. The value
-          MUST NOT be equal to any other <a>record type</a>s defined
-          in this API.
-        </dd>
       </dl>
+      <p>
+        In addition to [=well known type names=] it is also possible for
+        organizations to create a custom <dfn>external type name</dfn>
+        which is a string consisting of a domain name and a custom type name,
+        separated by a colon `U+003A` (`:`).
+      </p>
+      <p>
+        Applications MAY also use a <dfn>local type name</dfn>, which is a
+        string that MUST start with lowercase character or a number,
+        representing a type for a NFC Forum [=local type=]. It is
+        typically used in a record of an <a>NDEFMessage</a> that is the payload
+        of a parent <a>NDEFRecord</a>, for instance in a <a>smart poster</a>.
+        The context of the <a>local type</a> is the parent record whose payload
+        is the <a>NDEFMessage</a> to which this record belongs. The value
+        MUST NOT be equal to any other <a>record type</a>s defined
+        in this API, including other [=local type names=] and
+        [=external type names=].
+      </p>
       <p>
         Any implementation of Web NFC MUST transparently expose chunked records
         as single logical records, therefore <a>unchanged record</a>s are not
@@ -2793,13 +2774,9 @@
             </dl>
           </li>
           <li>
-            Let |name:string| be the result of running the
-            <a>validate external type</a> steps on |record|'s
-            <a>recordType</a>. If |name| is not `false`,
+            If running <a>validate external type</a> on |record|'s
+            <a>recordType</a> returns `true`,
             <ul>
-              <li>
-                Let |record|'s <a>recordType</a> be |name:string|.
-              </li>
               <li>
                 If |record|'s <a>data</a> is of type
                 {{NDEFMessageInit}}, then return the result of running
@@ -2838,24 +2815,17 @@
     </section>
 
     <section><h3>Validating external type</h3>
-      <p>
+      <div>
         The [[NFC-RTD]] specifies that external types MUST contain
-        the domain name of the issuing organization, a colon and a type
-        name that is at least one character long.
-      </p>
+        the domain name of the issuing organization, a colon `U+003A` (`:`)
+        and a type name that is at least one character long, for instance
+        "`w3.org:member`".
+      </div>
       <p>
-        The [[NFC-RTD]] specifies to omit the URN prefix “`urn:nfc:ext:`”
-        in the binary encoding of external type records, therefore
-        applications SHOULD NOT specify that when creating
-        <a>external type records</a>.
-      </p>
-      <p>
-        When the domain name is not provided, by default Web NFC uses the
-        <a data-cite="html">effective domain</a> of the {{Document/origin}} as
-        domain name in external type names when
-        <a href="#mapping-external-data-to-ndef">
-        creating external type records</a>. Therefore, applications MAY specify
-        only the colon and the external type name.
+        The [[NFC-RTD]] specifies the URN prefix “`urn:nfc:ext:`” as well, but
+        it is not written, i.e. it is not used in the binary encoding of
+        <a>external type records</a>, therefore Web NFC applications SHOULD NOT
+        specify the URN prefix when creating <a>external type records</a>.
       </p>
       <p class="note">
         The [[NFC-RTD]] requires that external type names are represented
@@ -2866,32 +2836,58 @@
         <a href="https://www.w3.org/2001/tag/doc/URNsAndRegistries-50.html">
         W3C TAG recommendations</a> to avoid using URNs in the Web.
         Therefore, Web NFC does not use the URN prefix neither when reading
-        or writing <a>NDEF messages</a>, but requires the leading colon ('`:`')
-        before the type name.
+        or writing <a>NDEF messages</a>.
       </p>
       <p>
-        To <dfn>validate external type</dfn> given a |input:USVString| run
+        To <dfn>validate external type</dfn> given |input:USVString|, run
         these steps:
       </p>
       <ol class=algorithm id="validate-external-type">
         <li>
           If |input:string| is not a {{USVString}} or it is empty, or its
-          length exceeds 255 bytes, or it contains white spaces,
-          return `false`.
+          length exceeds 255 bytes, return `false`.
         </li>
         <li>
-          If |input| does not start with colon ('`:`'), return |input|.
-          <p class="issue">
-            Validating domain name needs an algorithm.
-          </p>
+          Let  |colonIndex| be the last occurrence of `U+003A` (`:`) in |input|.
+          If a `U+003A` (`:`) is not found, return `false`.
         </li>
         <li>
-          If |input| starts with colon ('`:`'), let |domain:string| be the
-          <a data-cite="url#concept-url-serializer">serialization</a> of the
-          <a data-cite="html">effective domain</a> of the {{Document/origin}}.
+          Let |domain| be the part of |input| preceding |colonIndex|,
+          and |type| be the part following |colonIndex|.
         </li>
         <li>
-          Return |domain| concatenated with |input|.
+          Run the following sub-steps of
+          <a href="https://url.spec.whatwg.org/#host-parsing">host parsing</a>
+          on |domain|:
+          <ul>
+            <li>
+              Let |asciiDomain| be the result of running
+              <a href="https://url.spec.whatwg.org/#concept-domain-to-ascii">
+              domain to ASCII</a> on |domain|.
+            </li>
+            <li>
+              If |asciiDomain| is failure, return `false`.
+            </li>
+            <li>
+              If |asciiDomain| contains a
+              <a href="https://url.spec.whatwg.org/#forbidden-host-code-point">
+              forbidden host code point</a>, return `false`.
+            </li>
+          </ul>
+        </li>
+        <li>
+          If |type| is empty, return `false`.
+        </li>
+        <li>
+          If |type| contains [=code points=] that are not
+          [=ASCII alphanumeric=], or `U+0024` (`$`), `U+0027` (`'`),
+          `U+0028` `LEFT PARENTHESIS` (`(`), `U+0029` `RIGHT PARENTHESIS` (`)`),
+          `U+002A` (`*`), `U+002B` (`+`), `U+002C` (`,`), `U+002D` (`-`),
+          `U+002E` (`.`), `U+003B` (`;`), `U+003D` (`=`), `U+0040` (`@`),
+          `U+005F` (`_`), return `false`.
+        </li>
+        <li>
+          Return `true`.
         </li>
       </ol>
     </section>
@@ -3816,7 +3812,7 @@
     <p>
       The Unicode standard defines multiple encodings like UTF-8, UTF-16 and UTF-32.
       UTF-8 is the preferred encoding on the web, and it has the advantage
-      that it is endianness agnostic, as code points are represented in single bytes.
+      that it is endianness agnostic, as [=code points=] are represented in single bytes.
     </p>
     <p>
       The NDEF text records allow the text to be encoded as either UTF-8 or UTF-16.
@@ -3830,7 +3826,7 @@
     </p>
     <p>
       For UTF-16, byte order matters as it might differ between reading and writing
-      due to each code point spanning two bytes. For this reason UTF-16 encoded text
+      due to each [=code point=] spanning two bytes. For this reason UTF-16 encoded text
       usually contains a byte order mark (also known as BOM) which is written as
       `0xFEFF`. This means that if the byte order differs between host machine and
       NDEF, then the value will be read back as `0xFFFE`, indicating that the byte
--- a/index.html
+++ b/index.html
@@ -2910,9 +2910,8 @@
           return `false`.
         </li>
         <li>
-          If |input| is equal to any other <a>record type</a>s defined
-          in this API, including other [=local type names=] or
-          [=external type names=], return `false`.
+          If |input| is equal to the <a>record type</a> of any <a>NDEF record</a>
+          defined in its containing <a>NDEF message</a>, return `false`.
         </li>
         <li>
           Return `true`.
--- a/index.html
+++ b/index.html
@@ -2823,9 +2823,9 @@
       </div>
       <p>
         The [[NFC-RTD]] specifies the URN prefix “`urn:nfc:ext:`” as well, but
-        it is not written, i.e. it is not used in the binary encoding of
-        <a>external type records</a>, therefore Web NFC applications SHOULD NOT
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2866,32 +2850,52 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
         these steps:
       </p>
       <ol class=algorithm id="validate-external-type">
         <li>
           If |input:string| is not a {{USVString}} or it is empty, or its
-          length exceeds 255 bytes, or it contains white spaces,
+          length exceeds 255 bytes, return `false`.
+        </li>
+        <li>
+          Find the last occurrence of colon (`':'`) in |input|. If not found,
```
kenchris: Should probably wrote the Unicode number like Url spec does

index.html
```diff
@@ -1684,29 +1686,8 @@ <h2>The <dfn>record type</dfn> string</h2>
...
-            wkt-type             = (ALPHA / DIGIT) *(ALPHA / DIGIT / other)
-          </pre>
-          <p class=note>
-            The [[[NFC-RTD]]] defines every type in the [=well-known type records=]
-            and <a>external type records</a> in terms of URNs, but only a subset of
-            the URN is actually stored in the <a>NDEF record</a>'s <a>TYPE field</a>,
-            which corresponds to the above two ABNFs.
-          </p>
+          <a>external type record</a> that satisfies the
+          <a>validate external type</a> steps.
```
zolkis: "satisfies" doesn't really cut it, right?

index.html
```diff
@@ -2866,32 +2850,52 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
+        <li>
+          Find the last occurrence of colon (`':'`) in |input|. If not found,
           return `false`.
         </li>
         <li>
-          If |input| does not start with colon ('`:`'), return |input|.
-          <p class="issue">
-            Validating domain name needs an algorithm.
-          </p>
+          Otherwise let |domain| be the part of |input| preceding the last colon,
```
kenchris: Comma after otherwise

index.html
```diff
@@ -1629,7 +1629,8 @@ <h4>
         <a>recordType</a> attribute.
       </li>
       <li>
-        If the |recordType| value is "`smart-poster`", or an [=external type name=],
+        If the |recordType| value is "`smart-poster`", or if running
+        <a>validate external type</a> on |recordType| returns other than `false`,
```
zolkis: I hope you hate this sentence, too :)

index.html
```diff
@@ -2866,32 +2850,52 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
         <li>
-          If |input| does not start with colon ('`:`'), return |input|.
-          <p class="issue">
-            Validating domain name needs an algorithm.
-          </p>
+          Otherwise let |domain| be the part of |input| preceding the last colon,
+          and |type| be the part following the last colon.
+        </li>
+        <li>
+          Run the following sub-steps of <a href="">host parsing</a> on |domain|:
```
kenchris: Does that link work?

index.html
```diff
@@ -2793,12 +2774,11 @@ <h3><dfn>Writing or pushing content</dfn></h3>
             </dl>
           </li>
           <li>
-            Let |name:string| be the result of running the
-            <a>validate external type</a> steps on |record|'s
-            <a>recordType</a>. If |name| is not `false`,
+            If running <a>validate external type</a> on |record|'s
+            <a>recordType</a> returns |type| that is not `false`,
```
zolkis: the question here is, is the validation algo transforming the input (the steps from parse-host do, but do we use it, since it comes from apps)? If the steps don't change the record type, the algorithm can just succeed or fail, and no need to return `false` or a transformed string.

index.html
```diff
@@ -2838,24 +2818,28 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
-        <a>external type records</a>.
-      </p>
+        the domain name of the issuing organization, a colon (`:`) and a type
+        name that is at least one character long and follows the
+        <dfn>external type name ABNF</dfn>:
+        <pre class="abnf">
+          ext-type             = reg-name ":" custom-type
+          custom-type          = 1*(ALPHA / DIGIT / other)
+
+          DIGIT                = %x30-39
```
kenchris: Seems like NFC supports ip numbers :/ so maybe we need to use more from url spec (also that probably deserves a note)

index.html
```diff
@@ -2793,12 +2774,11 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
           <li>
-            Let |name:string| be the result of running the
-            <a>validate external type</a> steps on |record|'s
-            <a>recordType</a>. If |name| is not `false`,
+            If running <a>validate external type</a> on |record|'s
+            <a>recordType</a> returns |type| that is not `false`,
             <ul>
               <li>
-                Let |record|'s <a>recordType</a> be |name:string|.
+                Let |record|'s <a>recordType</a> be |type|.
```
zolkis: If there is no transformation (or not interesting), we should omit this step.

index.html
```diff
@@ -2866,32 +2850,52 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
+        </li>
+        <li>
+          If |type| is empty, return `false`.
         </li>
         <li>
-          If |input| starts with colon ('`:`'), let |domain:string| be the
-          <a data-cite="url#concept-url-serializer">serialization</a> of the
-          <a data-cite="html">effective domain</a> of the {{Document/origin}}.
+          If |type| does not match the `custom-type` definition from the
+          <a>external type name ABNF</a>, return `false`.
```
zolkis: could we improve this text? "match ABNF" stinks.

index.html
```diff
@@ -2866,32 +2850,52 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
         </li>
         <li>
-          Return |domain| concatenated with |input|.
+          Return |asciiDomain| concatenated with `':'` and |type|.
```
zolkis: Are we interested in this, or just return `true` and let recordType be the same that was provided by the app?

kenchris (changes requested):

index.html
zolkis: argh, omitted the URL

index.html
zolkis: `U+003A` I guess

index.html

index.html
zolkis: This definition is used only in the custom-type, so not.

zolkis: `reg-name` is not ABNF'd.

zolkis: but we have algorithm for checking it, so no problem IMHO

index.html
```diff
@@ -2866,32 +2850,54 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
         <li>
-          If |input| does not start with colon ('`:`'), return |input|.
-          <p class="issue">
-            Validating domain name needs an algorithm.
-          </p>
+          Find the last occurrence of colon (`':'`, `U+003A`) in |input|.
+          If not found, return `false`.
+        </li>
+        <li>
+          Otherwise, let |domain| be the part of |input| preceding the last colon,
```
beaufortfrancois: Please remove "Otherwise, ".

index.html
```diff
@@ -2866,32 +2850,54 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
         <li>
-          If |input| starts with colon ('`:`'), let |domain:string| be the
-          <a data-cite="url#concept-url-serializer">serialization</a> of the
-          <a data-cite="html">effective domain</a> of the {{Document/origin}}.
+          If |type| does not match the `custom-type` definition from the
+          <a>external type name ABNF</a>, return `false`.
         </li>
         <li>
-          Return |domain| concatenated with |input|.
+          Return |asciiDomain| concatenated with `':'` and |type|.
```
beaufortfrancois: It is not clear to me how/where the result returned by this algorithm is used in the spec.

zolkis: I made a related comment [here](https://github.com/w3c/web-nfc/pull/474#discussion_r357703634). It is used to replace the `recordType` (input to this validation) with the result of this algorithm (as a transformation).
However, validations steps usually don't transform. Then, since it's app input, I am not sure we want to use transformation (by the steps borrowed from the URL Standard host parsing steps).
We could just return `true` or `false` and use the original user input. That would be simpler (and I'd like it more).

beaufortfrancois: > We could just return `true` or `false` and use the original user input. That would be simpler (and I'd like it more).

+1

index.html
kenchris: Other that false is ugly... just make the algo return true/false

index.html
```diff
@@ -1648,8 +1649,9 @@ <h2>The <dfn>record type</dfn> string</h2>
         <a>NDEF record</a> types.
       </p>
       <p>
-        A set of known standardized values exists, but it is also possible
-        for organizations to create their own custom [=external type names=].
+        A set of known standardized values exists (well-known-types), but it is
```
kenchris: well-known-type names ? you refer to names below

index.html
kenchris: I don't think so.

Maybe you can write the ABNF in words instead. 

> which is a non whitespace string consisting of a domain name and a custom name, separated by U+003A (:). 
> 
> The custom name code points are ASCII alphanumeric, U+0024 ($), U+0027 ('), U+0028 LEFT PARENTHESIS, U+0029 RIGHT PARENTHESIS, U+002A (*), U+002B (+), U+002C (,), U+002D (-), U+002E (.), U+003B (;), U+003D (=), U+0040 (@), U+005F (_).

index.html
```diff
@@ -2866,32 +2850,54 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
+        </li>
+        <li>
+          Run the following sub-steps of
+          <a href="https://url.spec.whatwg.org/#host-parsing">host parsing</a>
+          on |domain|:
+          <ul>
+            <li>
+              Let |asciiDomain| be the result of running
+              <a href="https://url.spec.whatwg.org/#concept-domain-to-ascii">
+              domain to ASCII</a> on |domain|.
```
kenchris, line 2870: I wonder if this handles leading and trailing spaces - we should probably disallow...

index.html
kenchris: Yeah, its basically ASCII without whitespaces and a few extra special chars. You can write an algorithm.

- Is ASCII
- has whitespaces
- Contains other non-valid characters

index.html
kenchris: Nah, we can readd then

index.html
kenchris: And no control characters

kenchris: Actually they accept a very weird selection!

0x24 $
0x27 '
0x28 (
0x29 )
0x2A *
0x2B +
0x2C ,
0x2D -
0x2E .
0x3B ;
0x3D =
0x40 @ 
0x5F _

kenchris: URL spec uses wording like 

> Otherwise, if byte is 0x25 (%) and the next two bytes after byte in input are not in the ranges 0x30 (0) to 0x39 (9), 0x41 (A) to 0x46 (F), and 0x61 (a) to 0x66 (f), all inclusive, append byte to output.

kenchris: and

> The URL code points are ASCII alphanumeric, U+0021 (!), U+0024 ($), U+0026 (&), U+0027 ('), U+0028 LEFT PARENTHESIS, U+0029 RIGHT PARENTHESIS, U+002A (*), U+002B (+), U+002C (,), U+002D (-), U+002E (.), U+002F (/), U+003A (:), U+003B (;), U+003D (=), U+003F (?), U+0040 (@), U+005F (_), U+007E (~), and code points in the range U+00A0 to U+10FFFD, inclusive, excluding surrogates and noncharacters.

That can be used for description :)

kenchris: The type code points are ASCII alphanumeric, U+0024 ($), U+0027 ('), U+0028 LEFT PARENTHESIS, U+0029 RIGHT PARENTHESIS, U+002A (*), U+002B (+), U+002C (,), U+002D (-), U+002E (.), U+003B (;), U+003D (=), U+0040 (@), U+005F (_).

kenchris: image

index.html
zolkis: Thanks, great!

index.html
zolkis: Later step handles forbidden code points.
https://url.spec.whatwg.org/#forbidden-host-code-point

index.html
zolkis: I have updated (amended) the steps, please check again.

index.html
```diff
@@ -1681,45 +1682,26 @@ <h2>The <dfn>record type</dfn> string</h2>
...
-          is the <a>NDEFMessage</a> to which this record belongs. The value
-          MUST NOT be equal to any other <a>record type</a>s defined
-          in this API.
-        </dd>
       </dl>
+      <p>
+        In addition to [=well known type names=] it is also possible for
+        organizations to create a own custom <dfn>external type name</dfn> that
+        which is a string consisting of a domain name and a custom ASCII name,
+        separated by a colon (`':'`). The string may not contain any whitespace
```
kenchris: I think my other comment is probably better

index.html
```diff
@@ -2866,32 +2849,57 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
+          </ul>
+        </li>
+        <li>
+          If |type| is empty, return `false`.
         </li>
         <li>
-          If |input| starts with colon ('`:`'), let |domain:string| be the
-          <a data-cite="url#concept-url-serializer">serialization</a> of the
-          <a data-cite="html">effective domain</a> of the {{Document/origin}}.
+          If |type| contains code points that are not ASCII alphanumeric, or
```
kenchris: link to ASCII alphanumeric

index.html
zolkis: This is not the place we specify the algorithm, and we should avoid repeating that complex thing. So either we say "satisfies the validate external type name steps", or we give a short summary in words.

index.html
```diff
@@ -2866,32 +2849,57 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
-          length exceeds 255 bytes, or it contains white spaces,
-          return `false`.
+          length exceeds 255 bytes, return `false`.
         </li>
         <li>
-          If |input| does not start with colon ('`:`'), return |input|.
-          <p class="issue">
-            Validating domain name needs an algorithm.
-          </p>
+          Find the last occurrence of colon (`':'`, `U+003A`) in |input|.
```
kenchris: last occurrence of U+003A (`:`)

index.html
```diff
@@ -1681,45 +1682,26 @@ <h2>The <dfn>record type</dfn> string</h2>
...
-          of a parent <a>NDEFRecord</a>, for instance in a <a>smart poster</a>.
-          The context of the <a>local type</a> is the parent record whose payload
-          is the <a>NDEFMessage</a> to which this record belongs. The value
-          MUST NOT be equal to any other <a>record type</a>s defined
-          in this API.
-        </dd>
       </dl>
+      <p>
+        In addition to [=well known type names=] it is also possible for
+        organizations to create a own custom <dfn>external type name</dfn> that
```
kenchris: that which <- bad

index.html
kenchris: In addition to [=well known type names=] it is also possible for organizations to create a own custom <dfn>external type name</dfn> which is a string consisting of a domain name and a developer created type.

The developer created type code points are ASCII alphanumeric, U+0024 ($), U+0027 ('), U+0028 LEFT PARENTHESIS, U+0029 RIGHT PARENTHESIS, U+002A (*), U+002B (+), U+002C (,), U+002D (-), U+002E (.), U+003B (;), U+003D (=), U+0040 (@), U+005F (_).

kenchris: That seems fine and aligned with other specs.

The NFC spec made it complex

zolkis: Can we keep that enumeration in the validate external type steps only, and refer to them from here?

kenchris: URL specs uses U+00** when describing and 0x** when used in algos. I thikn this is fine. It is a spec, not documentation :)

index.html
```diff
@@ -2838,24 +2816,29 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
+        the domain name of the issuing organization, a colon (`:`) and a type
+        name that is at least one character long and follows the
+        <dfn>external type name ABNF</dfn>:
+        <pre class="abnf">
+          ext-type             = reg-name ":" custom-type
+          custom-type          = 1*(ALPHA / DIGIT / other)
+
+          DIGIT                = %x30-39
+          ALPHA                = %x41-5A / %x61-7A   ; A-Z / a-z
+          other                = "(" / ")" / "+" / "," / "-" / "=" /
```
kenchris: I think we should probably remove all of this.

kenchris (changes requested):

index.html
```diff
@@ -2866,32 +2836,57 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
+        </li>
+        <li>
+          If |type| is empty, return `false`.
         </li>
         <li>
-          If |input| starts with colon ('`:`'), let |domain:string| be the
-          <a data-cite="url#concept-url-serializer">serialization</a> of the
-          <a data-cite="html">effective domain</a> of the {{Document/origin}}.
+          If |type| contains code points that are not [=ASCII alphanumeric=], or
+          `U+0024` (`$`), `U+0027` (`'`), `U+0028` (`(`), `U+0029` (`)`),
```
kenchris: interestingly URL specs writes out (() as LEFT PARENTESIS etc

index.html
zolkis: OK, I took another attempt on it (amend).

index.html
kenchris: image

kenchris: We should probably follow for consistency

zolkis: Hmm. Strange. Should we use the same? (outsider reader might be confused why)

index.html
```diff
@@ -2866,32 +2836,57 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
         <li>
-          If |input| does not start with colon ('`:`'), return |input|.
-          <p class="issue">
-            Validating domain name needs an algorithm.
-          </p>
+          Find the last occurrence of `U+003A` (`:`) in |input|.
+          If not found, return `false`.
+        </li>
+        <li>
+          Let |domain| be the part of |input| preceding the last colon,
```
kenchris: cannot refer to colon here. Need to store the index above.

kenchris: Let colonIndex be the last ...

index.html
```diff
@@ -1681,45 +1682,26 @@ <h2>The <dfn>record type</dfn> string</h2>
...
-          MUST NOT be equal to any other <a>record type</a>s defined
-          in this API.
-        </dd>
       </dl>
+      <p>
+        In addition to [=well known type names=] it is also possible for
+        organizations to create a own custom <dfn>external type name</dfn>
+        which is a string consisting of a domain name and a custom ASCII name,
+        separated by a colon (`':'`) that satisfies the
+        <a>validate external type</a> steps.
```
kenchris: I still think it makes more sense to just write it out here instead of the satisfied wording

index.html
```diff
@@ -1681,45 +1682,26 @@ <h2>The <dfn>record type</dfn> string</h2>
...
-          is the <a>NDEFMessage</a> to which this record belongs. The value
-          MUST NOT be equal to any other <a>record type</a>s defined
-          in this API.
-        </dd>
       </dl>
+      <p>
+        In addition to [=well known type names=] it is also possible for
+        organizations to create a own custom <dfn>external type name</dfn>
+        which is a string consisting of a domain name and a custom ASCII name,
+        separated by a colon (`':'`) that satisfies the
```
kenchris: should we write the U+00 number here?

index.html
```diff
@@ -2866,32 +2836,57 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
+        </li>
+        <li>
+          Run the following sub-steps of
+          <a href="https://url.spec.whatwg.org/#host-parsing">host parsing</a>
+          on |domain|:
+          <ul>
+            <li>
+              Let |asciiDomain| be the result of running
+              <a href="https://url.spec.whatwg.org/#concept-domain-to-ascii">
+              domain to ASCII</a> on |domain|.
```
kenchris, line 2870: What are we actually writing? Are we writing the UTF-8 version of urls always? Maybe this deserves a note or so somewhere

index.html
kenchris: You can say `U+0029 RIGHT PARENTHESIS ())` - that is allowed

kenchris: but I agree that it is ugly and that `U+0029 RIGHT PARENTHESIS` seems fine. We will link code points to INFRA spec anyway so we are good

kenchris: We definitely need to follow INFRA spec... everyone is moving to do that

index.html
zolkis: We are not writing anything, just checking if app (user?) provided string qualifies for ext type name.

index.html
zolkis: Yeah, it's still pseudocode, and it's not more clear than before (when we just want to say use the last colon), but fine.

index.html
beaufortfrancois: +1

index.html
```diff
@@ -1681,45 +1682,25 @@ <h2>The <dfn>record type</dfn> string</h2>
...
-          of a parent <a>NDEFRecord</a>, for instance in a <a>smart poster</a>.
-          The context of the <a>local type</a> is the parent record whose payload
-          is the <a>NDEFMessage</a> to which this record belongs. The value
-          MUST NOT be equal to any other <a>record type</a>s defined
-          in this API.
-        </dd>
       </dl>
+      <p>
+        In addition to [=well known type names=] it is also possible for
+        organizations to create a own custom <dfn>external type name</dfn>
```
beaufortfrancois: Nit: remove `own`

index.html
zolkis: Solved in the Gordius way :)

index.html
```diff
@@ -2866,32 +2836,57 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
-          return `false`.
+          length exceeds 255 bytes, return `false`.
         </li>
         <li>
-          If |input| does not start with colon ('`:`'), return |input|.
-          <p class="issue">
-            Validating domain name needs an algorithm.
-          </p>
+          Let  |colonIndex| be the last occurrence of `U+003A` (`:`) in |input|.
+          If a `U+003A` (`:`) is not found, return `false`.
```
beaufortfrancois: Nit: Shall we swap the order of those two sentences?

index.html
beaufortfrancois: +1

index.html
zolkis: Removed the validation specific part.

index.html
zolkis: Fixed

index.html

index.html
zolkis: Fixed everywhere.

index.html
zolkis: I was thinking about that, but the way you write code, it's an error/exception handling. If I put it to the front, it's double pass if someone takes the steps literally. Maybe there is a better formulation?

index.html
zolkis: Right! Actually I removed it recently, strange.

index.html
```diff
@@ -2866,32 +2836,58 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
-          length exceeds 255 bytes, or it contains white spaces,
-          return `false`.
+          length exceeds 255 bytes, return `false`.
         </li>
         <li>
-          If |input| does not start with colon ('`:`'), return |input|.
-          <p class="issue">
-            Validating domain name needs an algorithm.
-          </p>
+          Let  |colonIndex| be the last occurrence of `U+003A` (`:`) in |input|.
```
kenchris: one space too many

index.html
kenchris: Not here, but when we are writing the type to the NDEF we are writing something :) - I assume as UTF-8

index.html
```diff
@@ -2866,32 +2836,58 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
+              Let |asciiDomain| be the result of running
+              <a href="https://url.spec.whatwg.org/#concept-domain-to-ascii">
+              domain to ASCII</a> on |domain|.
+            </li>
+            <li>
+              If |asciiDomain| is failure, return `false`.
+            </li>
+            <li>
+              If |asciiDomain| contains a
+              <a href="https://url.spec.whatwg.org/#forbidden-host-code-point">
```
kenchris: no better respec.org/xref way?

index.html
```diff
@@ -2838,24 +2815,17 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
+      </div>
       <p>
-        When the domain name is not provided, by default Web NFC uses the
-        <a data-cite="html">effective domain</a> of the {{Document/origin}} as
-        domain name in external type names when
-        <a href="#mapping-external-data-to-ndef">
-        creating external type records</a>. Therefore, applications MAY specify
-        only the colon and the external type name.
+        The [[NFC-RTD]] specifies the URN prefix “`urn:nfc:ext:`” as well, but
+        it is not written, i.e. it is not used in the binary encoding of
```
kenchris: is not stored in the NDEF record,

index.html
kenchris: Or we could do like url spec:

> If bytes contains a 0x3D (=), then let name be the bytes from the start of bytes up to but excluding its first 0x3D (=), and let value be the bytes, if any, after the first 0x3D (=) up to the end of bytes. If 0x3D (=) is the first byte, then name will be the empty byte sequence. If it is the last, then value will be the empty byte sequence.

If |input| contains a `U+003A` (`:`), then let |domain| be the |input| from the start of |input| up to but excluding its first `U+003A` (`:`), and let |type| be the |input|, if any, after the first `U+003A` (`:`) up to the end of |input|.

kenchris: If |input| does not contains a U+003A (:), return false

Let |domain| be the |input| from the start of |input| up to but excluding its first U+003A (:), and let |type| be the |input|, if any, after the first U+003A (:) up to the end of |input|.

kenchris:
Sorry for all the comments, this is getting there

index.html
zolkis: should we keep the other part, i.e. use

> but is not stored in the NDEF record, i.e. it is not used in the binary encoding of  
<a>external type records</a>,

or should we just say

> but is not stored in the NDEF record, therefore Web NFC applications ...

index.html
zolkis: except that it's the last occurrence of `U+003A`.

zolkis: And then need to check if |domain| and |type| are defined, i.e. in the steps above we need to include "or otherwise `null`

kenchris: Is it last though? We allow URLs but the scheme and port etc is not written down.

so we accept people to write "https://intel.com:8080:coolio" or actually "intel.com:coolio" - in the latter one there is only one : ever

kenchris: If |input| does not contains a U+003A (:), return false

Let |domain| be the |input| from the start of |input| up to but excluding its last U+003A (:), and let |type| be the |input|, if any, after the last U+003A (:) up to the end of |input|.

index.html
zolkis: Let's track that in separate issue. This is for removing effective domain.

index.html
```diff
@@ -1681,45 +1682,25 @@ <h2>The <dfn>record type</dfn> string</h2>
...
-          of a parent <a>NDEFRecord</a>, for instance in a <a>smart poster</a>.
-          The context of the <a>local type</a> is the parent record whose payload
-          is the <a>NDEFMessage</a> to which this record belongs. The value
-          MUST NOT be equal to any other <a>record type</a>s defined
-          in this API.
-        </dd>
       </dl>
+      <p>
+        In addition to [=well known type names=] it is also possible for
+        organizations to create a custom <dfn>external type name</dfn>
```
kenchris: comma <dfn> before which

index.html
zolkis: I chose the second.

index.html
zolkis: This needs another round now. The sentence above doesn't handle the case when a colon is not found.

index.html
kenchris: Sure, please file

kenchris (approved):

Issue #437, opened 2019-11-06
scan() should validate NDEFScanOptions#{recordType,mediaType}


leonhsl:
For example, `mediaType` should always be `undefined` for record types other than `mime`, otherwise we just throw a TypeError like what we do in constructing an `NDEFRecord`.

beaufortfrancois:
I understand where this is coming from, but I'm not sure we should throw a TypeError for scan options as this is technically NOT an invalid filter. It's just that scan won't trigger any onreading event.

I'm afraid we'd make it more complicated to cover all cases.
For instance, should we throw a TypeError when NDEFScanOptions `recordType` is set to `FOO`? 

I'll let other chime as I don't feel strongly about it.

zolkis:
>  I'm not sure we should throw a TypeError for scan options as this is technically NOT an invalid filter. It's just that scan won't trigger any onreading event.

I agree with this. The scan algorithm is sufficiently tight IMHO.

leonhsl:
I also prefer simplicity.

Take this scan options `{recordType: 'url', mediaType: 'text/plain'}` as an example, we have 2 choices for matching 'url' records (whose media type is always null):
  1. ignore the careless user input (mediaType) and match all 'url' records.
  2. do strictly, i.e. do not match any 'url' records.

I'd prefer 1 because we are already aware that mediaType does not make any sense for 'url' records at all, if we do not notify users of the invalid input then we'd better not fail silently when matching.

Either way, the spec needs to make a rule/note explicitly in the section of NDEFScanOptions definition.

zolkis:
What happens now is that properties defined in `option` are passed to the internal slots. No checks here. I guess this is what you'd like to get improved. 

Then, when a tag is read, the [Dispathcing NFC content](https://w3c.github.io/web-nfc/#dispatching-nfc-content) is run, which will first test for `id` equality for _any_ records of the message, then for `recordType` equality over _any_ records of the message, then for `mediaType` equality over _any_ records of the message. All that are specified in the internal slots must be equal to the values from the tags. Only then an event is fired.

This means that if `options` specifies `Foo` for `mediaType`, then the tag will have to have a record with `mediaType` being `Foo`, otherwise it's not a match. 

[EDIT] The main point is that an app CAN specify both `recordType = "url"` and a `mediaType` and still get a match, since this will compare ALL records and pick messages that have either a URL record, or have a matching `mediaType`, or an `id` that matches the filter.

zolkis:
[...]

PR #481, opened 2019-12-19
Fix #479: dispatching NFC content steps

### LINKED ISSUE #479: Filters from multiple readers should be independent
Quote [comment](https://github.com/w3c/web-nfc/issues/351#issuecomment-536523664) from @kenchris:
> In theory, as we can have multiple readers, they filtering should really be independent. I guess we could create another issue for that and treat it as an enhancement

zolkis:

***
<a href="https://pr-preview.s3.amazonaws.com/zolkis/web-nfc/pull/481.html" title="Last updated on Dec 24, 2019, 9:13 AM UTC (f382f13)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/481/576799b...zolkis:f382f13.html" title="Last updated on Dec 24, 2019, 9:13 AM UTC (f382f13)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3551,35 +3551,40 @@
             [= list/For each =] |record:NDEFRecord| in |message:NDEFMessage|,
             <ol>
               <li>
-                Let |matched| be `undefined`.
+                Let |matched:boolean| be `false`.
               </li>
               <li>
                 If |reader|.<a>[[\Id]]</a> is not `""` and is not equal to
                 |record|'s <a href="#dom-ndefrecord-id">id</a>,
-                [= iteration/continue =] with the next record.
+                [= iteration/continue =].
                 If it is equal, set |matched| to `true`.
               </li>
               <li>
                 If |reader|.<a>[[\RecordType]]</a> is
                 [= dictionary member/present =] and it is not equal to
                 |record|'s <a href="#dom-ndefrecord-recordtype">recordType</a>,
-                [= iteration/continue =] with the next record.
-                If it is equal and if |matched| is `undefined`, set |matched|
-                to `true`.
+                [= iteration/continue =].
+                If it is equal, set |matched| to `true`.
               </li>
               <li>
                 If |reader|.<a>[[\MediaType]]</a> is not `""` and is not equal
                 to |record|'s <a href="#dom-ndefrecord-mediatype">mediaType</a>,
-                [= iteration/continue =] with the next record.
-                If it is equal and if |matched| is `undefined`, set |matched|
-                to `true`.
+                [= iteration/continue =].
+                If it is equal, set |matched| to `true`.
               </li>
               <li>
-                If |matched| is `true`, <a>fire an event</a> named "`reading`"
-                at |reader| using <a>NDEFReadingEvent</a> with its
-                <a>serialNumber</a> attribute initialized to |serialNumber| and
-                <a>message</a> attribute initialized to |message| and
-                [= iteration/continue =] with the next reader.
+                If |matched| is `true`,
+                <ol>
+                  <li>
+                    <a>fire an event</a> named "`reading`"
+                    at |reader| using <a>NDEFReadingEvent</a> with its
+                    <a>serialNumber</a> attribute initialized to |serialNumber|
+                    and <a>message</a> attribute initialized to |message|.
+                  </li>
+                  <li>
+                    [= iteration/Break =].
+                  </li>
+                </ol>
               </li>
             </ol>
           </li>
--- a/index.html
+++ b/index.html
@@ -2042,8 +2042,7 @@
       <td><dfn>[[\MediaType]]</dfn></td>
       <td>An empty <a>string</a>.</td>
       <td>
-        The {{NDEFScanOptions}}.<a href="dom-ndefrecord-mediatype">mediaType</a>
-        value.
+        The {{NDEFScanOptions}}.<a>mediaType</a> value.
       </td>
       </tr>
       <tr>
@@ -3549,28 +3548,37 @@
         the <a>activated reader objects</a>, run the following sub-steps:
         <ol>
           <li>
-            If |reader|.<a>[[\Id]]</a> is [= dictionary member/present =]
-            and it is not equal to any |record|'s
-            <a href="#dom-ndefrecord-id">id</a> where |record|
-            is an element of |message|, [= iteration/continue =].
-          </li>
-          <li>
-            If |reader|.<a>[[\RecordType]]</a> is [= dictionary member/present =]
-            and it is not equal to any |record|'s
-            <a href="#dom-ndefrecord-recordtype">recordType</a> where |record|
-            is an element of |message|, [= iteration/continue =].
-          </li>
-          <li>
-            If |reader|.<a>[[\MediaType]]</a> is not `""` and
-            it is not equal to any |record|'s
-            <a href="#dom-ndefrecord-mediatype">mediaType</a> where |record| is
-            an element of |message|, [= iteration/continue =].
-          </li>
-          <li>
-            <a>Fire an event</a> named "`reading`" at |reader|
-            using <a>NDEFReadingEvent</a> with its <a>serialNumber</a> attribute
-            initialized to |serialNumber| and <a>message</a> attribute
-            initialized to |message|.
+            For each |record:NDEFRecord| in |message:NDEFMessage|,
+            <ol>
+              <li>
+                Let |matched| be `undefined`.
+              </li>
+              <li>
+                If |reader|.<a>[[\Id]]</a> is not `""` and is not equal to
+                |record|'s <a href="#dom-ndefrecord-id">id</a>, set |matched|
+                to `false`. If it is equal, set |matched| to `true`.
+              </li>
+              <li>
+                If |reader|.<a>[[\RecordType]]</a> is
+                [= dictionary member/present =] and it is not equal to
+                |record|'s <a href="#dom-ndefrecord-recordtype">recordType</a>,
+                set |matched| to `false`. If it is equal and if
+                |matched| is `undefined`, set |matched| to `true`.
+              </li>
+              <li>
+                If |reader|.<a>[[\MediaType]]</a> is not `""` and is not equal
+                to |record|'s <a href="#dom-ndefrecord-mediatype">mediaType</a>,
+                set |matched| to `false`. If it is equal and if |matched| is
+                `undefined`, set |matched| to `true`.
+              </li>
+              <li>
+                If |matched| is `true`, <a>fire an event</a> named "`reading`"
+                at |reader| using <a>NDEFReadingEvent</a> with its
+                <a>serialNumber</a> attribute initialized to |serialNumber| and
+                <a>message</a> attribute initialized to |message| and
+                abort these steps.
+              </li>
+            </ol>
           </li>
         </ol>
       </li>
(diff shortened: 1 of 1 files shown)


zolkis:
@Honry @leonhsl PTAL

zolkis:
Nah, stupid omissions in the algorithm. Amended twice, apologies. Now you should find the other stupid omissions :). This is a relatively simple algorithm IMHO, certainly simpler than my first attempt using [sets](https://infra.spec.whatwg.org/#sets). But if you have better ideas, please share.

index.html
```diff
@@ -3549,28 +3548,37 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
-            it is not equal to any |record|'s
-            <a href="#dom-ndefrecord-mediatype">mediaType</a> where |record| is
-            an element of |message|, [= iteration/continue =].
-          </li>
-          <li>
-            <a>Fire an event</a> named "`reading`" at |reader|
-            using <a>NDEFReadingEvent</a> with its <a>serialNumber</a> attribute
-            initialized to |serialNumber| and <a>message</a> attribute
-            initialized to |message|.
+            For each |record:NDEFRecord| in |message:NDEFMessage|,
```
Honry: Missing link for "For each".

index.html
```diff
@@ -3549,28 +3548,37 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
-            If |reader|.<a>[[\RecordType]]</a> is [= dictionary member/present =]
-            and it is not equal to any |record|'s
-            <a href="#dom-ndefrecord-recordtype">recordType</a> where |record|
-            is an element of |message|, [= iteration/continue =].
-          </li>
-          <li>
-            If |reader|.<a>[[\MediaType]]</a> is not `""` and
-            it is not equal to any |record|'s
-            <a href="#dom-ndefrecord-mediatype">mediaType</a> where |record| is
-            an element of |message|, [= iteration/continue =].
```
Honry, line 3567: One concern, why not use "continue"?

index.html
```diff
@@ -3549,28 +3548,37 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
+                |matched| is `undefined`, set |matched| to `true`.
+              </li>
+              <li>
+                If |reader|.<a>[[\MediaType]]</a> is not `""` and is not equal
+                to |record|'s <a href="#dom-ndefrecord-mediatype">mediaType</a>,
+                set |matched| to `false`. If it is equal and if |matched| is
+                `undefined`, set |matched| to `true`.
+              </li>
+              <li>
+                If |matched| is `true`, <a>fire an event</a> named "`reading`"
```
Honry: Include this in `for each` will lead to multiple times triggering of `reading` event if multiple records in the message matched. Because we are always dispatching the matched message rather than matched record, this maybe another problem.

index.html
```diff
@@ -3549,28 +3548,37 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
-            using <a>NDEFReadingEvent</a> with its <a>serialNumber</a> attribute
-            initialized to |serialNumber| and <a>message</a> attribute
-            initialized to |message|.
+            For each |record:NDEFRecord| in |message:NDEFMessage|,
+            <ol>
+              <li>
+                Let |matched| be `undefined`.
+              </li>
+              <li>
+                If |reader|.<a>[[\Id]]</a> is not `""` and is not equal to
```
leonhsl: I suppose we should always match any record if `Id` is not present?

index.html
```diff
@@ -3549,28 +3548,37 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
+                |record|'s <a href="#dom-ndefrecord-id">id</a>, set |matched|
+                to `false`. If it is equal, set |matched| to `true`.
```
leonhsl: If we found `Id` does not match, we can just skip comparing `RecordType` and `MediaType` and continue with the next record. Like:
For each reader,
  For each record in the message,
    1. If `Id` does not match, continue with the next record;
    2. If `RecordType` does not match, continue with the next record;
    3. If `MediaType` does not match, continue with the next record;
    4. Found a matched record, means the message matches, fire the event to the current reader and continue with the next reader.

index.html
zolkis: Depends. Any of the 3 fields might be missing and produce a match elsewhere, so if there is no `Id`, we can still have a match on `recordType` or `mediaType`.

index.html
zolkis: That was the old algorithm. 
This new one looks for any record in a message that satisfies _all_ defined elements of the filter.

zolkis: But you are right that in this algo, if a filter field is defined but it's not a match, we should continue, not just set `matched` to `false`. I will correct that.

index.html
zolkis: The step says to abort the algorithm if a match is found.

zolkis: Which was bad as well, as it should continue with the next reader instead of aborting.

Honry: Per [spec](https://infra.spec.whatwg.org/#iteration-break):
> [Break ]((https://infra.spec.whatwg.org/#iteration-break))will skip over any remaining steps in an iteration, and skip over any remaining items as well, stopping the iteration.

"break" would be more appropriate.

index.html
```diff
@@ -3549,28 +3548,40 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
-            initialized to |message|.
+            [= list/For each =] |record:NDEFRecord| in |message:NDEFMessage|,
+            <ol>
+              <li>
+                Let |matched| be `undefined`.
+              </li>
+              <li>
+                If |reader|.<a>[[\Id]]</a> is not `""` and is not equal to
+                |record|'s <a href="#dom-ndefrecord-id">id</a>,
+                [= iteration/continue =] with the next record.
```
kenchris: not sure we need to say "with the next record" - it is implied

index.html
```diff
@@ -3549,28 +3548,40 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
+                If it is equal, set |matched| to `true`.
+              </li>
+              <li>
+                If |reader|.<a>[[\RecordType]]</a> is
+                [= dictionary member/present =] and it is not equal to
+                |record|'s <a href="#dom-ndefrecord-recordtype">recordType</a>,
+                [= iteration/continue =] with the next record.
+                If it is equal and if |matched| is `undefined`, set |matched|
```
kenchris: Why the undefined check? I assume that if it is true, setting it to true makes no difference, but you don't want to set it if it was false?

index.html
```diff
@@ -3549,28 +3548,40 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
+                to `true`.
+              </li>
+              <li>
+                If |matched| is `true`, <a>fire an event</a> named "`reading`"
+                at |reader| using <a>NDEFReadingEvent</a> with its
+                <a>serialNumber</a> attribute initialized to |serialNumber| and
+                <a>message</a> attribute initialized to |message| and
+                [= iteration/continue =] with the next reader.
```
kenchris: It is weird that you do continue with another loop (reader?) here..

index.html
zolkis: I needed the `undefined` check since once `matched` was set to `false`, it was not supposed to be set to `true`. However, since the algo changed (we stop at no match), indeed we can simplify this.

index.html
zolkis: We should use "break" as @Honry suggested since we want to continue with the outer loop (next reader).

index.html
zolkis: Yes. Changed.

index.html
```diff
@@ -3549,28 +3548,45 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
+            [= list/For each =] |record:NDEFRecord| in |message:NDEFMessage|,
+            <ol>
+              <li>
+                Let |matched:boolean| be `false`.
+              </li>
+              <li>
+                If |reader|.<a>[[\Id]]</a> is not `""` and is not equal to
+                |record|'s <a href="#dom-ndefrecord-id">id</a>,
+                [= iteration/continue =].
+                If it is equal, set |matched| to `true`.
```
kenchris, line 3560: Why day if equal, you continue above if not

kenchris: Maybe

If "" continue
If not equal continue
Set matched to true

zolkis: If empty, continue is not good because you want to check for the next fields.

zolkis: I have played a bit with other formulations (in the spirit of your comment above) but I find this more clear. If a filter slot is not specified, match depends on other slots. If defined, and not matched, we can continue (otherwise depends on other slots again, but so far is a match).

index.html
```diff
@@ -3549,28 +3548,45 @@ <h3><dfn>Writing or pushing content</dfn></h3>
         the <a>activated reader objects</a>, run the following sub-steps:
```
beaufortfrancois, line 3548: I would add "and a" above: 

> To dispatch NFC content given a serialNumber of type serialNumber **and a** message of type NDEFMessage, run these steps:

index.html
```diff
@@ -3549,28 +3548,45 @@ <h3><dfn>Writing or pushing content</dfn></h3>
...
-            using <a>NDEFReadingEvent</a> with its <a>serialNumber</a> attribute
-            initialized to |serialNumber| and <a>message</a> attribute
-            initialized to |message|.
+            [= list/For each =] |record:NDEFRecord| in |message:NDEFMessage|,
+            <ol>
+              <li>
+                Let |matched:boolean| be `false`.
+              </li>
+              <li>
+                If |reader|.<a>[[\Id]]</a> is not `""` and is not equal to
```
beaufortfrancois: Shall we use "empty string" instead of double quotes as in initial value?
image

index.html
beaufortfrancois: How about this pattern below?

> - If reader.id is not empty string and is equal to records' id, set matched to true.
> - If matched is false, continue.

beaufortfrancois (approved):
LGTM with nits

index.html
leonhsl: I noticed there's some inconsistency in defining `Id`.

First we need to decide `Id` should initially be `undefined` or `An empty string`, then we need to update the following 3 places to be consistent:

1) https://w3c.github.io/web-nfc/#the-ndefscanoptions-dictionary
> dictionary NDEFScanOptions {
  USVString id;
  USVString recordType;
  USVString mediaType = "";
  AbortSignal? signal;
};

2) https://w3c.github.io/web-nfc/#dom-ndefscanoptions-id
> If the dictionary member is not present, then it will be ignored by the NFC listen algorithm.

3)  **here, this matching algorithm**.

index.html
zolkis: That doesn't work since if reader.id was empty, it's will depend on the others, so we don't continue, but fall through.

beaufortfrancois: Let's move on. We may want to ask @leonhsl POV when it gets implemented.

beaufortfrancois (approved):

index.html
zolkis: Let's track this in a separate issue.

index.html
leonhsl: FYI, [the current implementation](https://source.chromium.org/chromium/chromium/src/+/master:services/device/nfc/android/java/src/org/chromium/device/nfc/NfcImpl.java;l=597) is based on the consensus we got on the spec issue #437.

index.html
zolkis: See #488

Issue #478, opened 2019-12-17
NDEFReader.scan should say when it resolves and returns the promise p


dbaron:
So reading through the examples in [the explainer](https://github.com/w3c/web-nfc/blob/gh-pages/EXPLAINER.md), one thing I wondered was that it seems a little odd to do the asynchronous `await reader.scan()` and *then* set the `reader.onreading` property to an event handler.  It made me wonder whether the `reading` events could happen *during* the `reader.scan()` call.  So I decided to look at the [spec text for `reader.scan()`](https://w3c.github.io/web-nfc/#dom-ndefreader-scan) to understand whether this is the case -- and I ran into a problem.  As far as I can tell, nothing in that algorithm ever resolves and returns `p` in the success case -- which seems to imply that the function never returns at all!

I think there should be something in that spec text that says to resolve `p` and return it in the success case.  (At least I think that's how promise-using specifications are supposed to be written, although I suppose there actually isn't anything in the [promises guide](https://www.w3.org/2001/tag/doc/promises-guide) saying so.)

(I got here from w3ctag/design-reviews#461.)

beaufortfrancois:
Thanks for raising this issue @dbaron!
I've fixed spec at https://github.com/w3c/web-nfc/pull/480

Regarding the `onreading` event listener added **after** `await reader.scan()`, we thought it makes sample code easier to read as resolved promise means both that access was granted and NFC is enabled.

dbaron:
This still doesn't seem right, at least based on something that would make the examples work.  Having "resolve p" as step 7 makes it sound like the promise doesn't get resolved until after the browser *stops* trying to detect NFC technology.  I suspect the intent is to resolve the promise after "4.  Add reader to the activated reader objects.", although it's possible that it could be right before (4) as well (which would be detectably different).

dbaron:
(I'd also note that (2) and (3) don't really seem like separate steps.  But it also seems like, implicitly, between them, there's something like "wait for this request to ???", so that they could be combined as something like: " If this is the first listener being set up, then make a request to all NFC adapters to listen to NDEF messages.  Wait for this request to ???.  If the request fails, then the UA MAY reject p with a "NotSupportedError" DOMException and abort these steps."  (Though this also seems to forbid a UA from trying again if it fails once, which might be undesirable.)

Likewise, it's not clear to me that step 5, currently:

[...]

PR #484, opened 2019-12-23
Remove request to adapters to listen to NDEF messages

beaufortfrancois:
I think* we can remove the part where we "make a request to all NFC adapters to listen to NDEF messages" in scan promise as we're already saying that "If there are any NDEFReader instances in activated reader objects then the UA MUST listen to NDEF messages." See https://w3c.github.io/web-nfc/#listening-for-content

Note that we've already checked NFC adapter existence and if NFC connection can be established at that point.

Context: https://github.com/w3c/web-nfc/issues/478#issuecomment-568401815

***
<a href="https://pr-preview.s3.amazonaws.com/beaufortfrancois/web-nfc/pull/484.html" title="Last updated on Dec 24, 2019, 6:32 AM UTC (d54bf27)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/484/1ea511c...beaufortfrancois:d54bf27.html" title="Last updated on Dec 24, 2019, 6:32 AM UTC (d54bf27)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3430,14 +3430,6 @@
                 abort these steps.
               </li>
               <li>
-                If this is the first listener being set up, then make a request
-                to all <a>NFC adapter</a>s to listen to <a>NDEF message</a>s.
-              </li>
-              <li>
-                If the request fails, then the UA MAY reject |p| with a
-                {{"NotSupportedError"}} {{DOMException}} and abort these steps.
-              </li>
-              <li>
                 Add |reader| to the <a>activated reader objects</a>.
               </li>
               <li>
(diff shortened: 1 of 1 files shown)


kenchris (approved):

zolkis:
The explicit step was outlined to make clear that if there are multiple NFC adapters, all of them should be used in parallel. The UA should not make a decision which one to use.

I think that removing this would be fair but may need to clarify in the premise.

> If there are any NDEFReader instances in activated reader objects then the UA MUST listen to NDEF messages on all connected NFC adapters.

(or something like this).

beaufortfrancois:
@zolkis I've updated "listening for content" section based on your feedback at https://github.com/w3c/web-nfc/pull/484/commits/d54bf27456e000782122f806bc5589309766ebb4

zolkis (approved):

zolkis:
Thank you!

PR #489, opened 2019-12-24
Make NDEFRecord.toRecords() nullable

### LINKED ISSUE #487: toRecords() should be nullable or not?
I think `toRecords`() is better to be nullable.
It returning null indicates this `Sp/external` record's payload is just a plain buffer but not an NDEF message at all, and returning an empty sequence indicates the payload is an NDEF message that does not have any record.

Currently the spec https://w3c.github.io/web-nfc/#dfn-parse-records-from-bytes makes `toRecords`() return an empty sequence for both the cases mentioned above.

WDYT? The background is https://chromium-review.googlesource.com/c/chromium/src/+/1980586, we need to change either spec or impl to keep them consistent.
 @kenchris  @zolkis  @beaufortfrancois @riju

beaufortfrancois:
FIX https://github.com/w3c/web-nfc/issues/487

***
<a href="https://pr-preview.s3.amazonaws.com/beaufortfrancois/web-nfc/pull/489.html" title="Last updated on Dec 24, 2019, 9:51 AM UTC (a05ba0e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/489/019d78f...beaufortfrancois:a05ba0e.html" title="Last updated on Dec 24, 2019, 9:51 AM UTC (a05ba0e)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1529,7 +1529,7 @@
         readonly attribute USVString? encoding;
         readonly attribute USVString? lang;
 
-        sequence&lt;NDEFRecord&gt; toRecords();
+        sequence&lt;NDEFRecord&gt;? toRecords();
       };
 
       dictionary NDEFRecordInit {
@@ -3587,6 +3587,9 @@
     To <dfn>parse records from bytes</dfn> given |bytes:byte sequence|,
     run these steps:
     <ol class=algorithm>
+      <li>
+        If the length of |bytes| is `0`, return `null` and abort these steps.
+      </li>
       <li>
         Let |records| be the empty list.
       </li>


beaufortfrancois:
@leonhsl Does that look good to you as well?

zolkis (approved):

zolkis:
In addition, @leonhsl please check if `toRecords()` should return `null` also instead of re-throwing errors from the invoked "parse records from bytes" algorithm. Please file an issue if that is needed.

index.html
```diff
@@ -3587,6 +3587,9 @@ <h3>Parsing content</h3>
     To <dfn>parse records from bytes</dfn> given |bytes:byte sequence|,
     run these steps:
     <ol class=algorithm>
+      <li>
+        If the length of |bytes| is `0`, return `null` and abort these steps.
```
leonhsl, line 3591: Even if the length is >0, |bytes| may be a plain buffer that can not be parsed to get an NDEF message. From what I read, the current algo does not seem to be checking whether |bytes| is in the format of NDEF message or not...

leonhsl:
Yes I think we'd better also change https://w3c.github.io/web-nfc/#dfn-convert-ndefrecord-data-bytes to make toRecords() return `null` instead of re-throwing exceptions.

beaufortfrancois:
Noted. I'll address this shortly.

Issue #375, opened 2019-10-15
local types need some prefix of a kind


kenchris:
![image](https://user-images.githubusercontent.com/1101339/66819081-a06cb380-ef3e-11e9-88b9-4407e2dc06a5.png)

Look at the above. I can easily make a local type called "text" and then it will conflict with existing "text". NDEF binary gets around this by using uppercase for standardized well-known types and initial lowercase (or number) for local types, but that is very un-web-like.

A simple fix would be just add "local:" in front which kind of fits with external types "android.com:" etc

zolkis:
Local types have the local context (e.g. that of a parent record). 
Any "prefix" is application decision (e.g. to avoid conflicts).

Actually the NFC spec just talks about apps can "reuse" the local types (already defined in well-known types). It doesn't regulate anything else, except that they start with lowercase or number.

IMHO Web NFC should not impose any further restrictions here and we should not prefix local types.

kenchris:
We don't! There is a difference between how to describe types and how they are written in binary - you seem to confuse the two.

We use the name "text" for "T" as the web likes descriptive names/enums that are lowercase.

You can very well use a T well-known inside an embedded record, which we write using "recordType: text". Now what if I want to create a well-known local type "text" - I cannot write it as "text" because that would become "T" and not "text" - this is just for us to describe that we want a local type.

Without any prefix we are actually adding restrictions on what people can do, unlike what the NDEF spec does. You are disallowing "text", "media" etc.

We could just say that local types are written as `recordType: "local:text"` and that would return in a well-known type called "text" thus starting with lowercase

zolkis:
It's the apps that should care about prefixing, not the spec. 

Not much related (since apps decide), but the example seems a bit extreme, why would apps want to define a local type "text" instead of reusing a text record? They just want confusion and more work?

beaufortfrancois:
@kenchris @zolkis Shall we add a note saying that Web NFC does NOT support local type names such as "empty", "text", "url", "smart-poster", "mime", "absolute-url", "unknown" to make it clear there's an actual restriction?

beaufortfrancois:
For info, the spec already says "a local type name value MUST NOT be equal to any other record types defined in this API."

![image](https://user-images.githubusercontent.com/634478/70907793-0db5c700-200a-11ea-8f38-d057bea16d98.png)

kenchris:
[...]

PR #491, opened 2019-12-24
Add isLocal to NDEFRecord and NDEFRecordInit

beaufortfrancois:
This PR adds a boolean attribute `isLocal` to `NDEFRecord` and `NDEFRecordInit` in order to differentiate local type records "text" and TEXT records as raised in https://github.com/w3c/web-nfc/issues/375

Sample has been updated.

@leonhsl Please have a look as well.

***
<a href="https://pr-preview.s3.amazonaws.com/beaufortfrancois/web-nfc/pull/491.html" title="Last updated on Dec 24, 2019, 1:16 PM UTC (53e66a2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/491/910b21d...beaufortfrancois:53e66a2.html" title="Last updated on Dec 24, 2019, 1:16 PM UTC (53e66a2)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1385,7 +1385,7 @@
           if (record.recordType == "text" && !record.isLocal) {
             const decoder = new TextDecoder(record.encoding);
             text = decoder.decode(record.data);
-          } else if (record.recordType == "act" && record.isLocal) {
+          } else if (record.recordType == "act") {
             action = record.data.getUint8(0);
           }
         }
--- a/index.html
+++ b/index.html
@@ -1319,14 +1319,17 @@
             },
             {
               recordType: "t",  // type record, a local type to Sp
+              isLocal: true,
               data: encoder.encode("image/gif") // MIME type of the Sp content
             },
             {
               recordType: "s",  // size record, a local type to Sp
+              isLocal: true,
               data: new Uint32Array([4096]) // byte size of Sp content
             },
             {
               recordType: "act",  // action record, a local type to Sp
+              isLocal: true,
               // do the action, in this case open in the browser
               data: new Uint8Array([0])
             },
@@ -1379,10 +1382,10 @@
         let action, text;
 
         for (const record of externalRecord.toRecords()) {
-          if (record.recordType == "text") {
+          if (record.recordType == "text" && !record.isLocal) {
             const decoder = new TextDecoder(record.encoding);
             text = decoder.decode(record.data);
-          } else if (record.recordType == "act") {
+          } else if (record.recordType == "act" && record.isLocal) {
             action = record.data.getUint8(0);
           }
         }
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1379,10 +1382,10 @@ <h4>
         let action, text;
 
         for (const record of externalRecord.toRecords()) {
-          if (record.recordType == "text") {
+          if (record.recordType == "text" && !record.isLocal) {
             const decoder = new TextDecoder(record.encoding);
             text = decoder.decode(record.data);
-          } else if (record.recordType == "act") {
+          } else if (record.recordType == "act" && record.isLocal) {
```
zolkis: I think the isLocal test is not needed here. Only with those local types that clash with one of the web'ish type names we invented.

zolkis: "act" is an NFC Forum standardized local type.

beaufortfrancois: You're right!

index.html
```diff
@@ -3697,6 +3721,10 @@ <h3>Parsing content</h3>
       <li>
         Set |record|'s <a>id</a> to |ndef|'s |id:string|.
       </li>
+      <li>
+        Set |record|'s <a>isLocal</a> to `true` if |ndef| is a payload
+        to another <a>NDEF record</a>, or otherwise `false`.
+      </li>
```
zolkis, line 3727: Just note to myself, I think this needs one more check. No action for now, I need to think it through.

zolkis (approved):
Thanks! Just a nit comment for the example (your example is not wrong, though).

index.html
beaufortfrancois: Fixed at https://github.com/w3c/web-nfc/pull/491/commits/53e66a2ece7c67bbbfeea9ae324a1cdbcbd9aff4

kenchris (approved):

beaufortfrancois:
I'll merge for now.
@leonhsl Please let us know in the mean time if that looks good to you.

leonhsl:
https://w3c.github.io/web-nfc/#creating-ndef-record needs refactoring, from step 4.

My suggestion:

> If (isLocal) {
    if (not a payload to another record) reject and abort;
    if (is NOT valid local type) reject and abort;
    create and return the record (the current step 7)
    abort;
}
continue to handle Standardized types and external type.

leonhsl:
https://w3c.github.io/web-nfc/#parsing-ndef-records is wrongly setting `isLocal` to `true` for all embedded records, no matter what type they're.
We actually want to set `isLocal=true` only for those local type records (TNF=1 with a local type).

beaufortfrancois:
Thank you @leonhsl.
I address your feedback at https://github.com/w3c/web-nfc/pull/493

PR #492, opened 2019-12-24
Set empty record id to null

### LINKED ISSUE #465: Is an empty record type's id null or empty string?
Per 9.11.9.1 Parsing records from bytes:

> If idLength > 0, let ndef's id be result of running UTF-8 decode on the next idLength (ID field) bytes, or else let ndef's id be the empty string.

It should be empty string.

But spec says 
> The value 0x00 (Empty) indicates that there is no type or payload associated with this NDEF record. When used, the TYPE_LENGTH, ID_LENGTH and PAYLOAD_LENGTH fields are zero and the TYPE, ID and PAYLOAD fields are omitted from the NDEF record.

@beaufortfrancois would prefer null instead of empty string.

@zolkis, @kenchris, what's your opinion?

beaufortfrancois:
This PR sets empty record `id` to `null` as NFC Forum spec says the ID field is omitted from the NDEF record.

FIX https://github.com/w3c/web-nfc/issues/465

***
<a href="https://pr-preview.s3.amazonaws.com/beaufortfrancois/web-nfc/pull/492.html" title="Last updated on Dec 24, 2019, 1:08 PM UTC (989d616)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/492/910b21d...beaufortfrancois:989d616.html" title="Last updated on Dec 24, 2019, 1:08 PM UTC (989d616)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3708,6 +3708,9 @@
         (<a>empty record</a>):
         <ol>
           <li>
+            Set |record|'s <a>id</a> to `null`.
+          </li>
+          <li>
             Set |record|'s <a>recordType</a> to "`empty`".
           </li>
           <li>


kenchris (approved):

zolkis (approved):

zolkis:
Note to myself: so we voted for `null` instead of `undefined`.

PR #493, opened 2019-12-25
Fix isLocal algorithm

beaufortfrancois:
As suggested by @leonhsl at  https://github.com/w3c/web-nfc/pull/491#issuecomment-568853627 and https://github.com/w3c/web-nfc/pull/491#issuecomment-568854522, this PR refactors "Creating NDEF record" algorithm for `isLocal` and makes sure we set `isLocal=true` only for TNF=1 records.

***
<a href="https://pr-preview.s3.amazonaws.com/beaufortfrancois/web-nfc/pull/493.html" title="Last updated on Dec 25, 2019, 9:00 AM UTC (3834de0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/493/f445c52...beaufortfrancois:3834de0.html" title="Last updated on Dec 25, 2019, 9:00 AM UTC (3834de0)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2757,10 +2757,28 @@
             </ul>
           </li>
           <li>
-            If |record|'s <a>isLocal</a> is `true` and |record| is not a payload
-            to another <a>NDEF record</a>, reject |promise| with a {{TypeError}}
-            and abort these steps.
-          </li>
+            If |record|'s <a>isLocal</a> is `true`:
+            <ul>
+              <li>
+                If |record| is not a payload to another <a>NDEF record</a>,
+                reject |promise| with a {{TypeError}} and abort these steps.
+              </li>
+              <li>
+                If running the <a>validate local type</a> steps on |record|'s
+                <a>recordType</a> returns `false`, reject |promise| with a
+                {{TypeError}} and abort these steps.
+              </li>
+              <li>
+                If |record|'s <a>data</a> is of type
+                {{NDEFMessageInit}}, then return the result of running
+                the <a>create NDEF message</a> given |record|'s
+                <a>data</a>.
+              </li>
+              <li>
+                Otherwise, return the result of running
+                <a>map local type to NDEF</a> given |record| and |ndef|.
+              </li>
+            </ul>
           <li>
             Switching on |record|'s <a>recordType</a>, pass |record| and |ndef|
             to one of the following algorithms and return the result.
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -3752,6 +3752,10 @@ <h3>Parsing content</h3>
       <li>
         If |ndef|'s |typeNameField| is `1` ([=well-known type record=]):
         <ol>
+          <li>
+            Set |record|'s <a>isLocal</a> to `true` if |ndef| is a payload
```
leonhsl, line 3756: I suppose {`U`, `T`, `Sp`} types here should not be set `isLocal=true`.

index.html
```diff
@@ -2757,10 +2757,28 @@ <h3><dfn>Writing or pushing content</dfn></h3>
             </ul>
           </li>
           <li>
-            If |record|'s <a>isLocal</a> is `true` and |record| is not a payload
-            to another <a>NDEF record</a>, reject |promise| with a {{TypeError}}
-            and abort these steps.
-          </li>
+            If |record|'s <a>isLocal</a> is `true`:
```
zolkis, line 2760: Looks like you moved this to the front, but the validate local type steps were written with the old algorithm in mind. If we check local types up-front, we need to update/change the validation algorithm as well.

index.html
zolkis: Not all sub-records must be local types.

index.html
zolkis: Hmm, on a second look, the current validation algorithm should fare pretty well here as well.

index.html
zolkis: We need to decide what `isLocal` means: to denote that a record is in the payload of another record, of if the type is a local type.

zolkis: From a previous comment of @leonhsl it seems we want that `isLocal` means only a local type, not sub-recordness. However, when creating records, a local type may only exist in a containing record. But other than local types may be part of a message that is payload to a record.

leonhsl: Yeah, that's what I read from the discussion you had in the issue, `isLocal` is to denote the record is a local type, to differentiate a "text" local type record (sub-record of course) from a TEXT sub-record.

I think we allow users to create local type records at all times, but we reject the `push()` promise when we found the top-level message contains a local type record.

zolkis: I was also thinking if we could consider an implicit context (no conflicts within a message), but  the NFC Forum RTD clearly says "NFC Forum Local Types are available for use within the context of another record". So local type may only occur in the payload of a record, but we should set `isLocal` only for those sub-records that actually have a local type. Done :).

zolkis:
I think I will merge this and continue from there. Thanks!

PR #499, opened 2020-01-02
Add Page Visibility example

beaufortfrancois:
This PR adds to the explainer the Page Visibility API to the list of APIs that interact well with Web NFC.

### DIFF
--- a/EXPLAINER.md
+++ b/EXPLAINER.md
@@ -151,9 +151,10 @@ document.querySelector("#abortButton").onclick = event => {
 
 ## Integration with Page Visibility
 
-Web NFC functionality is allowed only for the Document of the top-level browsing
-context, which must be visible. Thanks to the Page Visibility API, web
-developers can track when it document visibility changes.
+Web NFC functionality is allowed only for the document of the top-level browsing
+context, which must be visible. Thanks to the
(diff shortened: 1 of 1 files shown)


EXPLAINER.md
```diff
@@ -149,6 +149,22 @@ document.querySelector("#abortButton").onclick = event => {
 };
 ```
 
+## Integration with Page Visibility
+
+Web NFC functionality is allowed only for the Document of the top-level browsing
```
kenchris: Since we are not linking to specs, I think it is fine to write document (lowercase)

EXPLAINER.md
```diff
@@ -149,6 +149,22 @@ document.querySelector("#abortButton").onclick = event => {
...
+context, which must be visible. Thanks to the Page Visibility API, web
```
kenchris: Should probably link to spec

kenchris (approved):

EXPLAINER.md
beaufortfrancois: Done

EXPLAINER.md
beaufortfrancois: Good catch. Done.

zolkis (approved):
Thanks!

Issue #501, opened 2020-01-02
Prefixing local records with ":" would be simpler than isLocal


kenchris:
I don't like isLocal because its optional so easy to get wrong, which is made clear with the last commits "sprinking it" over existing code :)

":" is not a valid first character for local record types, so it is fine.

I think it makes sense because we have local records in namespaced (external records) or standardized records (smart poster). It looks like a namespace where you just didn't explicitly write it out.

It will make make it very explicit where a local record type is used.

```js
const writer = new NDEFWriter();
writer.push({ records: [
  {
    recordType: "imdb.com:movies",
    data: { records: [
      {
        recordType: ":rating",
        data: 10
      },
      {
        recordType: ":title", 
        data: "Rise of Skywalker"
      }]
   }
}]);
```

kenchris:
a `recordType: "text"` would thus be a normal T record, and `recordType: ":text"` would be a local record of type "`text"`

beaufortfrancois:
FYI @kenchris https://github.com/w3c/web-nfc/issues/375#issuecomment-568164890 explains why the prefix idea was not picked.
> My take: I am not convinced the arbitrary complexity introduced by Web NFC in 2b) is worth doing, for the sake of saving complexity of defining both (standardized, i.e. already known) TNF and TYPE by applications.

kenchris:
>  A local type written as "_text" or "local:text" would be specified as "__text" or "local:local:text", respectively.

`_` and `:` are not allowed as first letter in local types, has to be number or lowercase letter, so you would never run into the above issues

beaufortfrancois:
For the exercise, I've updated spec to see how it would look: https://github.com/w3c/web-nfc/pull/502

zolkis:
Right, I have overlooked that `:` is illegal starting character for local types.

However, developers might be confused with this new Web NFC specific convention. It is also confusing because `:` is a separator in external types. 

On the other hand, in a way it's a logical choice (local is like external, just that instead of the explicit domain-context we have the implicit local context).

I'd prefer using the NFC Forum standardized type names, but we could assume this as a working hypothesis.

We should make Notes about the Web NFC introduced conventions about type names and ask for developer feedback between this (the current way) or using the plain NFC Forum standardized type names (`T`, `U`, `Sp`, `act`, `etc`).

beaufortfrancois:
https://github.com/w3c/web-nfc/pull/502 has been merged.
[...]

PR #502, opened 2020-01-02
Replace isLocal by colon prefix in recordType

beaufortfrancois:
As proposed in https://github.com/w3c/web-nfc/issues/501, this PR gets rid of the new `isLocal` and replaces it with a colon prefix in `recordType`.

***
<a href="https://pr-preview.s3.amazonaws.com/beaufortfrancois/web-nfc/pull/502.html" title="Last updated on Jan 2, 2020, 2:14 PM UTC (cca312d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/502/7b4431b...beaufortfrancois:cca312d.html" title="Last updated on Jan 2, 2020, 2:14 PM UTC (cca312d)">Diff</a>

### DIFF
--- a/EXPLAINER.md
+++ b/EXPLAINER.md
@@ -112,7 +112,7 @@ reader.onreading = event => {
   let action, text;
 
   for (const record of externalRecord.toRecords()) {
-    if (record.recordType == "text" && !record.isLocal) {
+    if (record.recordType == "text") {
       const decoder = new TextDecoder(record.encoding);
       text = decoder.decode(record.data);
     } else if (record.recordType == "act") {
--- a/index.html
+++ b/index.html
@@ -1318,18 +1318,15 @@
               data: "Funny dance"
             },
             {
-              recordType: "t",  // type record, a local type to Sp
-              isLocal: true,
+              recordType: ":t",  // type record, a local type to Sp
               data: encoder.encode("image/gif") // MIME type of the Sp content
             },
             {
-              recordType: "s",  // size record, a local type to Sp
-              isLocal: true,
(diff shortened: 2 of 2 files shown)


zolkis:
As [mentioned](https://github.com/w3c/web-nfc/issues/501#issuecomment-570210351), I don't quite like conventions added by Web NFC that change the NFC Forum standardized type names.

However, we could have it as a work hypothesis, and in this PR I suggest we add more clarity about what conventions are introduced by Web NFC. 

A Note (ed note or issue) should be added about the need of developer feedback on this.

kenchris:
I see the NFC as low level ways to store the data and not really user friendly. 

I think what we have is much nicer to work with and easier to understand

kenchris (approved):

zolkis (approved):
I suggest adding a Note or other clarificatio n (at your discretion) that summarizes what new conventions Web NFC is making about type names vs the NFC Forum standardized type names.

beaufortfrancois:
I've filed https://github.com/w3c/web-nfc/issues/504

kenchris (approved):

beaufortfrancois:
I'll merge now.
We'll use https://github.com/w3c/web-nfc/issues/504 to discuss differences we want to highlight about Web NFC vs NFC Forum spec.

leonhsl:
The writing direction is OK, i.e. `:act` will be written as `act` into an nfc tag.

But the reading direction misses adding `:` prefix to the local type read from the tag as the final type exposed to Web.

PR #513, opened 2020-01-03
Remove use-cases not supported by Web NFC

beaufortfrancois:
This PR removes all use-cases that are not supported by Web NFC for now in order to convey better what Web NFC is (NDEF P2P focused) , and what it is not (ISO-DEP, HCE).

Please let me know what you think.

***
<a href="https://pr-preview.s3.amazonaws.com/beaufortfrancois/web-nfc/pull/513.html" title="Last updated on Jan 6, 2020, 10:22 AM UTC (8273216)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/513/b1a2853...beaufortfrancois:8273216.html" title="Last updated on Jan 6, 2020, 10:22 AM UTC (8273216)">Diff</a>


cyberphone:
>Please let me know what you think.

It is (IMO) still a pity that Web NFC can't replace QR code for Desktop/Web-to-Mobile/Native applications, that nowadays are just about everywhere.

index.html
```diff
@@ -126,13 +126,17 @@
...
   </p>
   <p>
     This document defines an API to enable selected use-cases based on
-    NFC technology. The current scope of the specification is <a>NDEF</a>.
-    Other NFC technologies may be supported in the future.
+    NFC technology. The current scope of this specification is <a>NDEF</a>.
+  </p>
+  <p class="advisement">
+    Low-level I/O operations with the ISO-DEP protocol and Host-based Card
+    Emulation (HCE) are <b>not</b> supported in the current scope.
```
zolkis: Good clarification in the abstract.

index.html
```diff
@@ -161,7 +165,7 @@
 <!-- - - - - - - - - - - - - - -  Introduction  - - - - - - - - - - - - - - -->
 <section class="informative"> <h2>Introduction</h2>
   <div>
-    NFC user scenarios can be grouped as follows:
+    Web NFC user scenarios can be grouped as follows:
```
zolkis, line 168: In the Introduction section it would not be out of place to enumerate a few NFC use cases, while clarifying which ones are currently implemented by Web NFC. The classification was meant to provide generic information. Once we remove the generic use cases, I don't know if it makes sense to put the use cases here, and have a full Use Cases section as well (with basically the same content). Probably doesn't hurt.

index.html
```diff
@@ -1007,17 +944,6 @@ <h4>
...
-        NFC-A and NFC-F depending on the secondary device.
-      </li>
-      <li>
-        [future] Allow <a>NFC handover</a> to Bluetooth or WiFi.
-      </li>
-      <li>
-        [future] Allow card emulation with secure element or host card
-        emulation.
-      </li>
     </ol>
```
zolkis, line 946: In general, I am thinking should we somehow keep these somewhere as informal content, to define the use case coverage of NFC in general (with possible future adoption into Web NFC).

zolkis (commented):
I don't take a stance on this, need to think what would be the best way to put/organize these.
Feel free to merge if @kenchris approves.

kenchris (approved):
Can we store these in a use-cases.md file?

beaufortfrancois:
@kenchris @zolkis  I've updated existing use-cases.html document. See https://github.com/w3c/web-nfc/pull/513/commits/ff527e69730f6a720ed97d6881f7d7cbce7abbd0

archive/use-cases.html
```diff
@@ -62,6 +62,10 @@
     This specification defines an API to manage selected NFC use-cases from Web
     applications and pages, and to enable new use-cases based on NFC technology.
   </p>
+  <p class="advisement">
+    Low-level I/O operations with the ISO-DEP protocol and Host-based Card
```
kenchris: I don't know why we always say ISO-DEP - there are other low level protocols, but anyway that is a separate thing

archive/use-cases.html
```diff
@@ -62,6 +62,10 @@
...
+    Emulation (HCE) are <b>not</b> supported in the current scope.
```
kenchris: are not within the current scope?

archive/use-cases.html
```diff
@@ -139,6 +144,28 @@
      Other use cases may include bulk transfer of contacts cards, photos,
      and other data.
     </p>
+    <p class="note">
+      This use case is not supported in this version of the specification.
```
kenchris: Some places it is "not within scope" but here we talk about the spec - confusing

archive/use-cases.html
beaufortfrancois: Shall I say "Low-level I/O operations such as with the ISO-DEP protocol" instead?

archive/use-cases.html
```diff
@@ -139,6 +144,28 @@
...
+    </p>
+  </section>
+  <section><h3>Payment scenarios</h3>
```
kenchris: Rename? "Scenarios along side payment"?

archive/use-cases.html
```diff
@@ -139,6 +144,28 @@
...
+    <p>
+      Payment scenarios with Web NFC generally do not refer to supporting
```
kenchris: Payment using NFC should be handled at a platform level due to security considerations and is thus not within the scope of Web NFC. On the other hand it is possible to associate...

archive/use-cases.html
beaufortfrancois: Done.

archive/use-cases.html
```diff
@@ -139,6 +144,28 @@
...
+      the payment process itself, but associating the payment status with
+      a web page in order to have secondary actions. For instance,
+      the user buys goods in a store, and payment options include contactless
+      payment using NFC technology.
```
kenchris, line 159: new paragraph?

index.html
```diff
@@ -126,13 +126,17 @@
...
     The hardware standard is defined in [[[NFC-STANDARDS]]].
   </p>
   <p>
     This document defines an API to enable selected use-cases based on
-    NFC technology. The current scope of the specification is <a>NDEF</a>.
-    Other NFC technologies may be supported in the future.
+    NFC technology. The current scope of this specification is <a>NDEF</a>.
+  </p>
+  <p class="advisement">
+    Low-level I/O operations with the ISO-DEP protocol and Host-based Card
```
kenchris: as ISO-DEP, NFC-A and the like, ?

archive/use-cases.html
beaufortfrancois: You're right. I'll fix it to be consistent.

index.html
```diff
@@ -699,9 +682,9 @@ <h4>
       <dfn>NDEF Signature</dfn> is defined [[NDEF-SIGNATURE]].
       Its <a>TYPE field</a> contains "`Sig`" (`0x53`, `0x69`, `0x67`) and its
       <a>PAYLOAD field</a> contains version, signature and a certificate chain.
-      In this version of the API, Web NFC only provides the raw byte content
-      of the payload (see this
-      <a href="https://github.com/w3c/web-nfc/issues/363">issue</a>).
+    </p>
+    <p class="issue" data-number="363">
+      Web NFC only provides the raw byte content of the payload at the moment.
```
kenchris: As currently spec'ed,

index.html
```diff
@@ -869,34 +854,17 @@ <h4>
     A few NFC user scenarios have been enumerated
     <a href="http://www.w3.org/2009/dap/wiki/Near_field_communications_%28NFC%29#Use_cases_submitted_to_DAP_mailing_list">
     here</a> and in the
-    <a href="https://w3c.github.io/web-nfc/use-cases.html">Web NFC Use Cases</a>
-    document. The basic NFC interactions are the following.
+    <a href="/archive/use-cases.html">Web NFC Use Cases</a>
```
kenchris: I wonder if we should somehow highlight this better - graphics?

index.html
```diff
@@ -869,34 +854,17 @@ <h4>
...
+    document. The basic Web NFC interactions are the following.
```
kenchris: rudimentary?

index.html
```diff
@@ -869,34 +854,17 @@ <h4>
...
-      <li>
-        Reading an <a>NFC tag</a> when no {{Document}} using Web NFC is
-        <a>visible</a>.
-        <p class="note">
-          This use case is not supported in this version of the specification.
-        </p>
-      </li>
-    </ol>
+    <p>
+      Reading an <a>NFC tag</a> containing an <a>NDEF message</a>, when the
```
kenchris: s/when/while?

index.html
```diff
@@ -1007,17 +942,6 @@ <h4>
         devices, such as writable tags, when they come in range, as
         <a>NDEF message</a>s.
       </li>
-      <li>
-        [future] Allow manual connection for various technologies such as
-        NFC-A and NFC-F depending on the secondary device.
```
kenchris, line 1012: So there were removed but not mentioned together with ISO-DEP... Also should such be links? I think we mention these while discussion the tag technologies

archive/use-cases.html
kenchris: I would also mention NFC-A/B, NFC-F or so - to make it obvious that more exist

beaufortfrancois: Done.

archive/use-cases.html
beaufortfrancois: Done.

archive/use-cases.html
beaufortfrancois: Done.

archive/use-cases.html
beaufortfrancois: Done.

index.html
beaufortfrancois: Done.

index.html
beaufortfrancois: This seems to suggest the same thing as "at the moment" for me. This is an issue.

index.html
beaufortfrancois: I'm not sure how graphics would highlight this better.
Shall we track this in another issue ?

index.html
beaufortfrancois: Done.

index.html
beaufortfrancois: Done.

index.html
beaufortfrancois: What do you suggest?

beaufortfrancois (commented):
Thanks for reviewing @kenchris!

index.html
kenchris: I suggested it as a replacement for that. I think it is more clear

index.html
```diff
@@ -713,8 +696,10 @@ <h4>
       The negotiated communication carrier would then be used (separately) to
       perform certain activities between the two devices, such as sending photos
       to the other device, printing to a Bluetooth printer or streaming video to
-      a television set. Web NFC does not support this at the moment (see this
-      <a href="https://github.com/w3c/web-nfc/issues/364">issue</a>.
+      a television set.
+    </p>
+    <p class="issue" data-number="364">
+      Web NFC does not support this at the moment.
```
kenchris: As currently spec'ed, this is not supported.

index.html
kenchris: sure

index.html
kenchris: I think adding is above (as you did) is fine

kenchris (approved):

index.html
zolkis: It was archived because we moved all content in the spec. However, now we moved the content back and it might make sense to move back the use cases doc to the root (informal and showing a roadmap to further development).

index.html
zolkis: I think we could just leave this removed in this place. The "Allow manual connection for ..." sentence could be moved to the use-cases doc (probably already there).

zolkis (approved):
WDYT about restoring the use-cases.html to the root (for info and roadmap)?

index.html
beaufortfrancois: Done.

index.html
beaufortfrancois: Done.

index.html
beaufortfrancois: I moved back the use-cases doc to the root.

PR #514, opened 2020-01-03
Fix #511: validate external type

### LINKED ISSUE #511: Fix external type validation algorithm
As noted [here](https://github.com/w3c/web-nfc/issues/350#issuecomment-570529776)

> Probably what we should do instead is figure out if `dns-part` is literal IPv6 and take the _first_ `:` as delimiter after it, allowing `:` in the `name-part` of external types.

zolkis:

***
<a href="https://pr-preview.s3.amazonaws.com/zolkis/web-nfc/pull/514.html" title="Last updated on Jan 6, 2020, 10:50 AM UTC (b77c8f0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/514/b1a2853...zolkis:b77c8f0.html" title="Last updated on Jan 6, 2020, 10:50 AM UTC (b77c8f0)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2812,73 +2812,33 @@
       </p>
       <ol class=algorithm id="validate-external-type">
         <li>
-          If |input:string| is not an [=ASCII string=] or it is empty, or its
+          If |input:string| is not an [=ASCII string=], is empty, or its
           length exceeds 255 bytes, return `false`.
         </li>
         <li>
-          If input starts with `U+005B` (`[`), then:
-          <ol>
-            <li>
-              Let |address| be the part of input between `U+005B` (`[`) and
-              the first occurrence of `U+005D` (`]`) or `null` if that is not
-              found.
-            </li>
-            <li>
-              If |address| is `null`, return `false`.
-            </li>
-            <li>
-              Let |domain| be the result of running
-              <a href="https://url.spec.whatwg.org/#concept-ipv6-parser">
-              IPv6 parsing</a> on |address|.
-            </li>
-            <li>
-              If |domain| is failure, return `false`.
-            </li>
-            <li>
-              Let |rest| be the |input| after the first occurrence, if any, of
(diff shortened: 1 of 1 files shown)


kenchris (commented):
Can you explain the [] to me?

zolkis:
> Can you explain the [] to me?

In an URL, the IPv6 address should be enclosed in "[" and "]" brackets, IIUC. [[rfc2732]](https://tools.ietf.org/html/rfc2732)
Since the NDEF spec uses this prose (referring to URIs), if this convention does not apply in your opinion in the external type RTD, then we can disallow using it in Web NFC and take out the whole step 2.

Edit: My interpretation is that even though the NFC RTD spec calls the type format RTD-URI in section 3 (RTD Type Names ABNF), but it does not require URL conventions in the `dns-part` (instead, defines its own format), so it's disputable if we should support IPv6 there.

index.html
```diff
@@ -2812,43 +2812,33 @@ <h3><dfn>Writing or pushing content</dfn></h3>
       </p>
       <ol class=algorithm id="validate-external-type">
         <li>
-          If |input:string| is not a {{USVString}} or it is empty, or its
+          If |input:string| is not an [=ASCII string=] or it is empty, or its
```
beaufortfrancois: Nit: "If |input:string| is not an [=ASCII string=], is empty, or its length exceeds 255 bytes, return `false`."

beaufortfrancois (approved):
LGTM with nit

zolkis:
Okay, this became a very simple update, in the end.
As discussed in #511, we don't need to support IPv6 in the external type, since domain name is mainly used for namespacing and we don't expect relationship between the `dns-name` part used in external types and the document origin.

index.html
zolkis: Fixed, thanks.

zolkis:
Merging with keeping both commits.

PR #518, opened 2020-01-07
Reworded security section about NDEF signature

beaufortfrancois:
***
<a href="https://pr-preview.s3.amazonaws.com/beaufortfrancois/web-nfc/pull/518.html" title="Last updated on Jan 7, 2020, 10:30 AM UTC (e8ba435)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/518/0a518fc...beaufortfrancois:e8ba435.html" title="Last updated on Jan 7, 2020, 10:30 AM UTC (e8ba435)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4080,6 +4080,10 @@
 <!-- - - - - - - - - - - - - Security and Privacy - - - - - - - - - - - - - -->
 <section> <h2 id="security">Security and Privacy</h2>
   <p>
+    As general security measure, the whole software stack needs to be security
+    hardened.
+  </p>
+  <p>
     In order to protect the integrity and authenticity of NDEF messages, the NFC
     Forum introduced [[NDEF-SIGNATURE]].
   </p>
--- a/index.html
+++ b/index.html
@@ -4080,11 +4080,9 @@
 <!-- - - - - - - - - - - - - Security and Privacy - - - - - - - - - - - - - -->
 <section> <h2 id="security">Security and Privacy</h2>
   <p>
-    NFC technology involves multiple levels of security. Payments done with NFC
-    are considered to be secure at hardware level, but the whole software stack
-    needs to be security hardened.
-    However, NFC data transfers involve additional threats. In order to mitigate
-    these, the NFC Forum introduced [[NDEF-SIGNATURE]].
+    In order to protect the integrity and authenticity of NDEF messages, the NFC
+    Forum introduced [[NDEF-SIGNATURE]]. Signing NDEF records prevents malicious
+    use of NFC tags.
   </p>
 
   <section> <h3>Chain of trust</h3>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -4080,11 +4080,9 @@ <h3>Parsing content</h3>
...
 <section> <h2 id="security">Security and Privacy</h2>
   <p>
-    NFC technology involves multiple levels of security. Payments done with NFC
-    are considered to be secure at hardware level, but the whole software stack
-    needs to be security hardened.
-    However, NFC data transfers involve additional threats. In order to mitigate
-    these, the NFC Forum introduced [[NDEF-SIGNATURE]].
+    In order to protect the integrity and authenticity of NDEF messages, the NFC
+    Forum introduced [[NDEF-SIGNATURE]]. Signing NDEF records prevents malicious
+    use of NFC tags.
```
kenchris: but it is also the other way around.

This sounds like you harden the tag instead, but it also allows the reader to only accept such hardened tags and thus hardens the reader as well

beaufortfrancois: That's not how I intended to be understood ;)
How would you rephrase it?

zolkis: I would remove "Signing NDEF records prevents malicious use of NFC tags." 
That is inaccurate.

index.html
```diff
@@ -4080,11 +4080,9 @@ <h3>Parsing content</h3>
 <!-- - - - - - - - - - - - - Security and Privacy - - - - - - - - - - - - - -->
 <section> <h2 id="security">Security and Privacy</h2>
   <p>
-    NFC technology involves multiple levels of security. Payments done with NFC
-    are considered to be secure at hardware level, but the whole software stack
-    needs to be security hardened.
```
zolkis, line 4085: I would add a separate sentence, like

"As general security measure,  the whole software stack needs to be security hardened".

And place where you think it adds most clarity/value.

index.html
beaufortfrancois: I've removed it.

zolkis (approved):

index.html
beaufortfrancois: This seems obvious to me and not specific to Web NFC, don't you think?

zolkis: IIRC security reviewers insisted on stating the obvious :), but it's the editors' call, indeed.

beaufortfrancois: WDYT of https://github.com/w3c/web-nfc/pull/518/commits/ae9e5471a4ef736b1a1311ba19558fbd17e6685b

zolkis:
We already take that data as untrusted.

beaufortfrancois:
I'll let you update the explainer and I'll remove it.

PR #524, opened 2020-01-09
Remove peer target mode

beaufortfrancois:
Android NFC P2P APIs (Android Beam) are [deprecated](https://source.android.com/setup/start/android-10-release#android_beam_deprecated). Moreover, it looks like P2P NFC can be achieved with Host-Card Emulation aka HCE. In that case, I'd argue we should remove support for NDEFPushOptions `"peer"` target and investigate how HCE could be added to Web NFC.

Let me know what you think.

***
<a href="https://pr-preview.s3.amazonaws.com/beaufortfrancois/web-nfc/pull/524.html" title="Last updated on Jan 10, 2020, 4:21 AM UTC (985104f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/524/d12fac4...beaufortfrancois:985104f.html" title="Last updated on Jan 10, 2020, 4:21 AM UTC (985104f)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -164,26 +164,19 @@
 
 <!-- - - - - - - - - - - - - - -  Introduction  - - - - - - - - - - - - - - -->
 <section class="informative"> <h2>Introduction</h2>
-  <div>
-    Web NFC user scenarios can be grouped as follows:
-    <ul>
-      <li>
-        Hold a device in close proximity to a passively powered tag, such as
-        a plastic card or sticker, in order to read and/or write data.
-      </li>
-      <li>
-        Hold two active devices, e.g. phones or tablets, in close proximity
-        in order to push data from one device to the other.
-      </li>
-    </ul>
-  </div>
+  <p>
+    Web NFC user scenario is as follows: Hold a device in close proximity to a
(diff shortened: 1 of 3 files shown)


index.html
```diff
@@ -231,13 +224,6 @@
       manufacturer. They may also expose an <a>NDEF</a> message.
     </p>
   </div>
-  <p>
```
zolkis, line 234: This is informative content and could stay IMHO. We could add a note that given (link) reasons, we don't support P2P at the moment.

zolkis (commented):
Appears this PR overlaps with #526?

index.html
beaufortfrancois: Done.

beaufortfrancois:
As stated in https://github.com/w3c/web-nfc/pull/526#issue-360887509, #526 sits on top this PR (#524). That's because we'll only rename NDEFWriter push to write if/when we remove peer-to-peer support.

kenchris (approved):

zolkis (approved):

PR #534, opened 2020-01-13
Updated explainer

beaufortfrancois:
This PR updates the introduction of the Web NFC explainer by adding user flow.

### DIFF
--- a/EXPLAINER.md
+++ b/EXPLAINER.md
@@ -10,19 +10,27 @@ Participate
 - https://lists.w3.org/Archives/Public/public-web-nfc/
 
 ## Introduction
+
 Web NFC aims to provide sites the ability to read and write to nearby NFC (Near
-Field Communication) devices (e.g., tags and phones) in a secure and privacy
-preserving manner.
+Field Communication) tags in a secure and privacy preserving manner.
 
 NFC consists of a rather large set of technologies, so for this first iteration
 of the API the focus has been on supporting the NFC Data Exchange Format aka
 [NDEF](https://www.oreilly.com/library/view/beginning-nfc/9781449324094/ch04.html),
 a lightweight binary message format, as it works across different tag formats.
 
-The focus has been on making an easy to use, and very webby API that web
-developers would be comfortable with, and at the same time making sure that it
-is low-level and flexible enough so that users can get the full potential out of
-NDEF and even integrate with legacy solutions.
+Web NFC is only available to top-level, secure browsing contexts, and origins
+must first request the “nfc” permission while handling a user gesture. To then
+perform a read or write, the web page must be visible when the user touches
+their device to an NFC tag. The browser uses haptic feedback to indicate a tap.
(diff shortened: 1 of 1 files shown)


zolkis (approved):

EXPLAINER.md
```diff
@@ -10,19 +10,27 @@ Participate
...
 of the API the focus has been on supporting the NFC Data Exchange Format aka
 [NDEF](https://www.oreilly.com/library/view/beginning-nfc/9781449324094/ch04.html),
 a lightweight binary message format, as it works across different tag formats.
 
-The focus has been on making an easy to use, and very webby API that web
-developers would be comfortable with, and at the same time making sure that it
-is low-level and flexible enough so that users can get the full potential out of
-NDEF and even integrate with legacy solutions.
+Web NFC is only available to top-level, secure browsing contexts, and origins
+must first request the “nfc” permission while handling a user gesture. To then
```
kenchris, line 23: remove then

EXPLAINER.md
```diff
@@ -10,19 +10,27 @@ Participate
...
+perform a read or write, the web page must be visible when the user touches
```
kenchris: you dont "touch to"

when the user touches an NFC tag with their device - or brings an NFC tag within reading range of their device..

EXPLAINER.md
```diff
@@ -10,19 +10,27 @@ Participate
...
+their device to an NFC tag. The browser uses haptic feedback to indicate a tap.
```
kenchris: a tap = a tag of some kind was read

EXPLAINER.md
```diff
@@ -10,19 +10,27 @@ Participate
...
+Access to the NFC radio is blocked if the display is off or the device is
+locked. For backgrounded web pages, receiving and pushing NFC content are
+suspended.
+
+We believe Web NFC is an easy to use, and very webby API that web developers
```
kenchris: webby? an API that fits well with the API design of the web platform?

EXPLAINER.md
beaufortfrancois: "then" indicate the action happens after the "nfc" permission was granted. I think we should keep it for this.

EXPLAINER.md
beaufortfrancois: Fixed.

EXPLAINER.md
beaufortfrancois: I've removed "webby" as we already mention that "web developers will be comfortable with"

EXPLAINER.md
beaufortfrancois: a tap is pretty universal to understand.

EXPLAINER.md
kenchris: And at that point...

EXPLAINER.md
kenchris: ok

PR #536, opened 2020-01-22
Improve parsing smart poster: add missing steps

### LINKED ISSUE #505: How to read/parse local types
The ["parse an NDEF record"](https://w3c.github.io/web-nfc/#dfn-parse-an-ndef-record) algorithm used when reading "sub records" doesn't say explicitly how to deal with local types such as ":act", ":s", etc. It only deals with "T", "U", and "Sp".

@zolkis Is that intentional ?

image

This issue was caught by https://github.com/w3c/web-nfc/pull/502#issuecomment-570468555

zolkis:
Changed the parse record steps to also receive a context.
Fixes #505 .
Likely improvements are needed on formulating the steps, @beaufortfrancois any suggestions?
@leonhsl PTAL

***

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Mar 10, 2020, 3:39 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Frawcdn.githack.com%2Fzolkis%2Fweb-nfc%2F92d49fdb05d177af265dd2f5bc06cdb5865f73ed%2Findex.html%3FisPreview%3Dtrue)

```
Navigation timeout of 30000 ms exceeded
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/web-nfc%23536.)._
</details>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2628,9 +2628,9 @@
             If |record|'s <a>recordType</a> starts with a colon `U+003A` (`:`):
             <ul>
               <li>
-                If |record| is not a payload to another <a>NDEF record</a>
-                and if |context| is not `"smart-poster"`,
-                reject |promise| with a {{TypeError}} and abort these steps.
+                If |context| is `""` (i.e. |record| is not a payload to another
+                <a>NDEF record</a>), reject |promise| with a {{TypeError}} and
+                abort these steps.
               </li>
               <li>
                 If running the <a>validate local type</a> steps on |record|'s
--- a/index.html
+++ b/index.html
@@ -648,6 +648,10 @@
             just provides the values.
           </p>
         </li>
+        <li>
+          A <a>smart poster</a> MAY also contain other records, which can be
+          handled in an application specific manner.
+        </li>
       </ul>
     </div>
 
@@ -1509,9 +1513,14 @@
         <a>recordType</a> attribute.
       </li>
       <li>
-        If the |recordType| value is "`smart-poster`", or if running
-        <a>validate external type</a> on |recordType| returns `true`,
-        then return the result of running <a>parse records from bytes</a> on |bytes|.
+        If the |recordType| value is "`smart-poster`", then return the result of
+        running <a>parse records from bytes</a> given |bytes| and
+        `"smart-poster"`.
+      </li>
+      <li>
+        If running <a>validate external type</a> on |recordType| returns `true`,
+        then return the result of running <a>parse records from bytes</a> given
+        |bytes| and `"external"`.
       </li>
       <li>
         Otherwise, [= exception/throw =] a
@@ -2324,8 +2333,8 @@
               </li>
               <li>
                 Let |output| be the notation for the <a>NDEF message</a>
-                to be created by UA, as the result of passing
-                |message| to <a>create NDEF message</a>.
+                to be created by UA, as the result of invoking
+                <a>create NDEF message</a> with |message| and `""`.
                 If this throws an exception, reject |p| with that
                 exception and abort these steps.
               </li>
@@ -2431,7 +2440,7 @@
   <section><h3>Creating NDEF message</h3>
     <div>
       To <dfn>create NDEF message</dfn> given a
-      |source:NDEFMessageSource| run these steps:
+      |source:NDEFMessageSource| and |context:string|, run these steps:
       <ol class=algorithm id="create-web-nfc-message">
         <li>Switch on |source:NDEFMessageSource|'s type:
           <dl>
@@ -2451,7 +2460,7 @@
             <dt>{{BufferSource}}</dt>
             <ul>
                 <li>
-                  Let |mimeRecord| be an <a>NDEFrecord</a> initialized with its
+                  Let |mimeRecord| be an <a>NDEFRecord</a> initialized with its
                   |recordType| set to "`mime`", |data| set to |source|, and
                   |mediaType| set to "`application/octet-stream`".
                 </li>
@@ -2465,8 +2474,8 @@
             <dt>{{NDEFMessageInit}}</dt>
             <ul>
               <li>
-                If |source|'s records [= list/is empty =], throw a
-                {{TypeError}} and abort these steps.
+                If |source|'s records [= list/is empty =], [= exception/throw =]
+                a {{TypeError}} and abort these steps.
               </li>
             </ul>
             <dt>unmatched type</dt>
@@ -2487,8 +2496,8 @@
           <ol>
             <li>
               Let |ndef| be the result of running <a>create NDEF record</a>
-              given |record:NDEFRecordInit|, or make sure the underlying
-              platform provides equivalent values to |ndef|.
+              given |record:NDEFRecordInit| and |context|, or make sure the
+              underlying platform provides equivalent values to |ndef|.
               If the algorithm throws an exception |e|, reject |promise| with
               |e| and abort these steps.
             </li>
@@ -2498,15 +2507,48 @@
           </ol>
         </li>
         <li>
+          If running <a>check created records</a> given |output| and |context|
+          throw an |error: Error|, reject |promise| with |error| and abort these
+          steps.
+        </li>
+        <li>
           Return |output|.
         </li>
       </ol>
     </div>
 
+    <section><h3>Check created records</h3>
+      <div>
+        To <dfn>check created records</dfn> given |records: NDEFRecord sequence|
+        and |context: string|, run these steps:
+        <ol class=algorithm>
+          <li>
+            If |context| is `"smart-poster"` and |records| does not contain
+            exactly one <a>URI record</a>, or if it contains more than one
+            <a>type record</a>, <a>size record</a> or <a>action record</a>,
+            throw {{TypeError}} and abort these steps.
+          </li>
+          <li>
+            Return `true`;
+          </li>
+        </ol>
+        <p class="note">
+          Icon record media types could be limited to `"image/"` or `"video/"`,
+          but the [[NDEF-SMARTPOSTER]] specification does actually allow other
+          media type records in a <a>smart poster</a>, which can be treated in
+          an application-specific manner, for instance a vCard contact card
+          using one of its associated
+          <a href="https://www.iana.org/assignments/media-types/media-types.xhtml">
+          MIME types</a>. Applications MAY ignore any extra records inside
+          the <a>smart poster</a>.
+        </p>
+      </div>
+    </section>
+
     <section><h3>Creating NDEF record</h3>
       <div>
-        To <dfn>create NDEF record</dfn> given |record:NDEFRecordInit|, run
-        these steps:
+        To <dfn>create NDEF record</dfn> given |record:NDEFRecordInit|
+        and |context:string|, run these steps:
         <ol data-link-for="NDEFRecordInit">
           <li>
             Let |ndef| be the representation of an <a>NDEF record</a> to be
@@ -2534,7 +2576,8 @@
             If |record|'s <a>recordType</a> starts with a colon `U+003A` (`:`):
             <ul>
               <li>
-                If |record| is not a payload to another <a>NDEF record</a>,
+                If |record| is not a payload to another <a>NDEF record</a>
+                or if |context| is `"smart-poster"`,
                 reject |promise| with a {{TypeError}} and abort these steps.
               </li>
               <li>
@@ -2548,13 +2591,17 @@
               </li>
             </ul>
           <li>
-            Switching on |record|'s <a>recordType</a>, pass |record| and |ndef|
-            to one of the following algorithms and return the result.
+            Switching on |record|'s <a>recordType</a>, invoke one of the
+            following algorithms and return the result.
             If the algorithm throws an exception |e|, reject |promise| with |e|
             and abort these steps.
             <dl>
               <dt>"`empty`"</dt>
               <ul>
+                <li>
+                  If |context| is `"smart-poster"`,
+                  [= exception/throw =] a {{TypeError}} and abort these steps.
+                </li>
                 <li>
                   Return <a>map empty record to NDEF</a> given |record| and |ndef|.
                 </li>
@@ -2580,12 +2627,16 @@
               <dt>"`smart-poster`"</dt>
               <ul>
                 <li>
-                  Return <a>map smart poster to NDEF</a> given |record| and
-                  |ndef|.
+                  Return <a>map smart poster to NDEF</a> given |record|, |ndef|
+                  and |context|.
                 </li>
               </ul>
               <dt>"`absolute-url`"</dt>
               <ul>
+                <li>
+                  If |context| is `"smart-poster"`,
+                  [= exception/throw =] a {{TypeError}} and abort these steps.
+                </li>
                 <li>
                   Return <a>map absolute-URL to NDEF</a> given |record| and
                   |ndef|.
@@ -2595,8 +2646,16 @@
           </li>
           <li>
             If running <a>validate external type</a> on |record|'s
-            <a>recordType</a> returns `true`, return
-            <a>map external data to NDEF</a> given |record| and |ndef|.
+            <a>recordType</a> returns `true`,
+            <ol>
+              <li>
+                If |context| is `"smart-poster"`,
+                [= exception/throw =] a {{TypeError}} and abort these steps.
+              </li>
+              <li>
+                Return <a>map external data to NDEF</a> given |record| and |ndef|.
+              </li>
+            </ol>
           </li>
           <li>
             Otherwise, [= exception/throw =] a {{TypeError}} and abort
@@ -2988,17 +3047,22 @@
             and abort these steps.
           </li>
           <li>
-            Let |mimeTypeRecord| be the <a>MIME type</a>
+            Let |mimeType| be the <a>MIME type</a>
             returned by running <a>parse a MIME type</a> on
             |record|'s <a>mediaType</a>.
-            <ol>
-              <li>
-                If |mimeTypeRecord| is failure, let |mimeTypeRecord| be a new
-                <a>MIME type record</a> whose type is "`application`", and
-                subtype is "`octet-stream`".
-              </li>
-            </ol>
           </li>
+          <li>
+            If |mimeType| is failure, let |mimeTypeRecord| be a new
+            <a>MIME type record</a> whose type is "`application`", and
+            subtype is "`octet-stream`".
+          </li>
+          <!--
+          <li>
+            If |context| is `"smart-poster"` and |mimeType| does not
+            start with `"image/"` or `"video/"`, throw <a>TypeError</a>
+            and abort these steps.
+          </li>
+          -->
           <li>
             Set |arrayBuffer| to |record|'s <a>data</a>.
           </li>
@@ -3013,7 +3077,7 @@
           </li>
           <li>
             Set the |ndef|'s <a>TYPE field</a> to the result of
-            <a>serialize a MIME type</a> with |mimeTypeRecord| as
+            <a>serialize a MIME type</a> with |mimeType| as
             the input.
           </li>
           <li>
@@ -3078,7 +3142,8 @@
             <ol>
               <li>
                 Set the |ndef|'s <a>PAYLOAD field</a> to the result of running
-                the <a>create NDEF message</a> given |record|'s <a>data</a>.
+                the <a>create NDEF message</a> given |record|'s <a>data</a>
+                and `"external"`.
               </li>
               <li>
                 Set the |ndef|'s <a>PAYLOAD LENGTH field</a> to the length of
@@ -3146,7 +3211,8 @@
             <ol>
               <li>
                 Set the |ndef|'s <a>PAYLOAD field</a> to the result of running
-                the <a>create NDEF message</a> given |record|'s <a>data</a>.
+                the <a>create NDEF message</a> given |record|'s <a>data</a>
+                and `"local"`.
               </li>
               <li>
                 Set the |ndef|'s <a>PAYLOAD LENGTH field</a> to the length of
@@ -3184,7 +3250,8 @@
           </li>
           <li>
             Set |ndef|'s <a>PAYLOAD field</a> to the result of running the
-            <a>create NDEF message</a> given |record|'s <a>data</a>.
+            <a>create NDEF message</a> given |record|'s <a>data</a> and
+            `"smart-poster"`.
           </li>
           <li>
             Set |ndef|'s <a>PAYLOAD LENGTH field</a> to the length of
@@ -3433,7 +3500,7 @@
           </li>
           <li>
             Let |record:NDEFRecord| be the result of <a>parse an NDEF record</a>
-            on |ndef|.
+            given |ndef| and `""`.
           </li>
           <li>
              If |record| is not `null`, <a>append</a> |record| to |message|'s
@@ -3512,8 +3579,8 @@
   <h3>Parsing content</h3>
 
   <section><h3>Parsing records from bytes</h3>
-    To <dfn>parse records from bytes</dfn> given |bytes:byte sequence|,
-    run these steps:
+    To <dfn>parse records from bytes</dfn> given |bytes:byte sequence| and
+    |context: string|, run these steps:
     <ol class=algorithm>
       <li>
         If the length of |bytes| is `0`, return `null` and abort these steps.
@@ -3602,13 +3669,23 @@
           </li>
           <li>
             Let |record:NDEFRecord| be the result of <a>parse an NDEF record</a>
-            on |ndef|.
+            given |ndef| and |context|.
           </li>
           <li>
             If |record| is not `null`, <a>append</a> |record| to |records|.
           </li>
           <li>
-            If |messageEnd| is `true`, abort these sub-steps.
+            If |messageEnd| is `true`,
+            <ol>
+              <li>
+                If <a>check parsed records</a> given |records| and |context|
+                throws an |error|, reject |promise| with |error| and abort these
+                steps.
+              </li>
+              <li>
+                Otherwise abort these sub-steps (terminate the loop).
+              </li>
+            </ol>
           </li>
         </ol>
       </li>
@@ -3618,9 +3695,25 @@
     </ol>
   </section>
 
+  <section><h3>Check parsed records</h3>
+    To <dfn>check parsed records</dfn> given |records: NDEFRecord sequence|
+    and |context: string|, run these steps:
+    <ol class=algorithm>
+      <li>
+        If |context| is `"smart-poster"` and |records| does not contain
+        exactly one <a>URI record</a>, or if it contains more than one
+        <a>type record</a>, <a>size record</a> or <a>action record</a>,
+        throw {{TypeError}} and abort these steps.
+      </li>
+      <li>
+        Otherwise return `true`.
+      </li>
+    </ol>
+  </section>
+
   <section><h3>Parsing NDEF records</h3>
   <div>
-    To <dfn>parse an NDEF record</dfn> given |ndef| into a
+    To <dfn>parse an NDEF record</dfn> given |ndef| and |context:string| into a
     |record:NDEFRecord|, run these steps:
     <ol class=algorithm data-link-for="NDEFRecord">
       <li>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -3651,39 +3656,81 @@ <h3>Parsing content</h3>
...
-              <ul>
+              <ol>
                 <li>
-                  Running <a>parse an NDEF smart-poster record</a> on |ndef|.
+                  Set |record| to the result of running
+                  <a>parse an NDEF smart-poster record</a> on |ndef|.
                 </li>
-              </ul>
+              </ol>
+              <dt>"`:s`" (`0x53` `0x70`)</dt>
```
beaufortfrancois: Please update hex representation.

index.html
```diff
@@ -3651,39 +3656,81 @@ <h3>Parsing content</h3>
...
+                <li>
+                  If |context| is not equal to `"smart-poster"`,
+                  [= exception/throw =] a {{TypeError}} and abort these steps.
+                </li>
+                <li>
+                  Set |record| to the result of running
+                  <a>parse a smart-poster size record</a> on |ndef|.
+                </li>
+              </ol>
+              <dt>"`:t`" (`0x53` `0x70`)</dt>
```
beaufortfrancois: Please update hex representation.

index.html
```diff
@@ -3651,39 +3656,81 @@ <h3>Parsing content</h3>
...
+                <li>
+                  If |context| is not equal to `"smart-poster"`,
+                  [= exception/throw =] a {{TypeError}} and abort these steps.
+                </li>
+                <li>
+                  Set |record| to the result of running
+                  <a>parse a smart-poster type record</a> on |ndef|.
+                </li>
+              </ol>
+              <dt>"`:act`" (`0x53` `0x70`)</dt>
```
beaufortfrancois: Please update hex representation.

index.html
```diff
@@ -3651,39 +3656,81 @@ <h3>Parsing content</h3>
...
+              <ol>
+                <li>
+                  If |context| is not equal to `"smart-poster"`,
+                  [= exception/throw =] a {{TypeError}} and abort these steps.
+                </li>
+                <li>
+                  Set |record| to the result of running
+                  <a>parse a smart-poster action record</a> on |ndef|.
+                </li>
+              </ol>
```
beaufortfrancois: How about other local type records that may contained in a smart-poster record.
Nothing prevents that if I understand it correctly.

index.html
```diff
@@ -3867,9 +3914,121 @@ <h3>Parsing content</h3>
...
+        </p>
       </li>
     </ol>  <!-- parsing NDEF smart-poster record -->
   </div>
+  <div>
+    To <dfn>parse a smart-poster size record</dfn> given an |ndefRecord| into a
+    |record:NDEFRecord|, run these steps:
+    <ol class=algorithm data-link-for="NDEFRecord">
+      <li>
+        Set |record|'s <a>recordType</a> to "`s`".
```
beaufortfrancois: Shouldn't it be `:s`?

beaufortfrancois: Same for type and action record.

index.html
```diff
@@ -3867,9 +3914,121 @@ <h3>Parsing content</h3>
...
+      </li>
+      <li>
+        Set |record|'s <a>mediaType</a> to `null`.
+      </li>
+      <li>
+        Let |buffer:byte sequence| be the <a>byte sequence</a> of
+        |ndefRecords|'s <a>PAYLOAD field</a>.
+        <p class="note">
+          Applications can parse this value as a string that contains an
+          [[RFC2048]] media type that denotes the medi type of the object the
```
beaufortfrancois: Typo: `media`

index.html
zolkis: Fixed, thanks. I had to rewrite it quite much :).

index.html
zolkis: Right. And we need an extra algo for inserting `:` before local types.

index.html
```diff
@@ -3651,39 +3656,73 @@ <h3>Parsing content</h3>
...
+            <a>parse a smart-poster type record</a> on |ndef|.
+          </li>
+          <li>
+            If |ndef|'s |type:string| is <strong>"`act`" (`0x65` `0x63` `0x74`)
+            </strong> and if |context| is equal to `"smart-poster"`,
+            set |record| to the result of running
+            <a>parse a smart-poster action record</a> on |ndef|.
+          </li>
+          <li>
+            Otherwise, if running the <a>validate local type</a> steps on
```
beaufortfrancois: Shall we explicitly call out what happens if it's not a valid local type?

index.html
```diff
@@ -3867,9 +3906,147 @@ <h3>Parsing content</h3>
...
+        Return |record|.
+      </li>
+    </ol>
+  </div>
+  <div>
+    To <dfn>validate smart poster icon type</dfn> given |type:string|,
+    run these steps:
+    <ol class=algorithm data-link-for="NDEFRecord">
+      <li>
+        If |type| starts with `"image"` or `"video"`, return `true`.
```
beaufortfrancois: Shall we add an extra `/` after `image` and `video`?

index.html
zolkis: Yes, fixed.

index.html
zolkis: Yes, fixed (amended).

index.html
```diff
@@ -3657,63 +3657,59 @@ <h3>Parsing content</h3>
...
+            <a>parse a smart-poster type record</a> on |ndef|.
+          </li>
+          <li>
+            If |ndef|'s |type:string| is <strong>"`act`" (`0x65` `0x63` `0x74`)
+            </strong> and if |context| is equal to `"smart-poster"`,
+            set |record| to the result of running
+            <a>parse a smart-poster action record</a> on |ndef|.
+          </li>
+          <li>
+            Otherwise, if running the <a>validate local type</a> steps on
```
beaufortfrancois: Please remove "Otherwise, " now that we have a final otherwise at the end.

beaufortfrancois (commented):
As suggested by @jakearchibald, we should make sure Web NFC web platform tests cover this PR before merging it. If not covered, we should update our web platform tests before.

index.html
zolkis: Done

zolkis:
> As suggested by @jakearchibald, we should make sure Web NFC web platform tests cover this PR before merging it. If not covered, we should update our web platform tests before.

To me it seems that [Web NFC WPT](https://github.com/web-platform-tests/wpt/tree/master/web-nfc) has not been covering local types and smart poster yet. 
I am not sure I understand the suggestion (link?), so does this mean the spec cannot advance before implementations+tests catch up? That would be awkward. If there needs to be clarity on spec+impl+tests, let's make a spec release for impl and a draft branch for the spec to be able to move forward.

Right now not the whole spec is implemented yet, but that should not stop the spec from fixing issues.

beaufortfrancois:
According to https://whatwg.org/working-mode#changes, "each normative change made to the standard needs to have corresponding test changes, either in the form of new tests or modifications to existing tests."
@riju We are missing web platform tests for local types and smart poster.

zolkis:
Well, that has not been applied in the past. If we apply it here, until tests catch up we should move spec development to a new `draft` branch. Then stuff can be moved to gh-pages when they have test coverage.

EDIT: To be consistent, we should also probably trim the spec in `gh-pages` to the level of existing tests + impl and move the rest to a `draft` branch.

index.html
```diff
@@ -3433,7 +3438,7 @@ <h3><dfn>Writing content</dfn></h3>
           </li>
           <li>
             Let |record:NDEFRecord| be the result of <a>parse an NDEF record</a>
-            on |ndef|.
+            given |ndef| and `"root"`.
```
Honry: Why pass `root` here? Does it mean nothing?

Honry:
@leonhsl and I will take care of the WPT tests.

Honry:
@zolkis, spec says a smart poster record should have a mandatory URI record at:

>Smart poster is defined in [NDEF-SMARTPOSTER] to describe a given web content as an NDEF record that contains an NDEF message as payload, including the following records: 
> - A single mandatory URI record that refers to the smart poster content.

This should be also described in smart poster's parsing algorithm.

Honry:
Moreover, smart poster's type, size, and action records are restricted at most 1 respectively.

zolkis:
@Honry thanks for the comments, all valid, will address soon. Will use `""` instead of `"root"` and fix the rest, too.

beaufortfrancois:
https://github.com/w3c/web-nfc/pull/536/commits/ad5c6d2845266867058d86d76fe0344cd195ed6f seems to address reading a smart-poster record.
Should it also raise an issue when web developers try to create a smart-poster without a URL record for instance?

zolkis:
> Should it also raise an issue when web developers try to create a smart-poster without a URL record for instance?

It already does.

beaufortfrancois:
> > Should it also raise an issue when web developers try to create a smart-poster without a URL record for instance?
> 
> It already does.

I can't find where sorry ;(

zolkis:
Line [3622](https://github.com/w3c/web-nfc/pull/536/commits/ad5c6d2845266867058d86d76fe0344cd195ed6f#diff-eacf331f0ffc35d4b482f1d15a887d3bR3622).
Please tell if it needs better formulation.

beaufortfrancois:
This line IIUC is for reading/parsing smart poster records from NFC tags.
What I'm saying is that we should raise validation errors when web developers try to write a smart-poster without a URL record to a NFC tag.

```js
writer.write({ records: [
  {
    recordType: "smart-poster",
    data: { records: [] }
  }
]}).catch(error => {
// This should raise a validation error!
});
```

zolkis:
Thanks for noticing. I included the write path as well.
Note that similar to reading, I pass a context (as internal parameter to the spec) to the create steps. 
This context is defined for smart-poster, local and external, but at the moment only the first has checks.
Still this felt more logical then passing `""` as context to the steps creating external and local type records. Implementors will "optimize" them out for now :) and we might use these in later checks.

Honry:
@zolkis, per smart-poster record only support limit embedded record type at https://w3c.github.io/web-nfc/#smart-poster-record, we should also throw TypeError for the other unsupported type, e.g. empty record, absolute-url record, external record, unknown record, other local type records.

zolkis:
>  we should also throw TypeError for the other unsupported type,

Hmm, yes, we could do that. Just thinking should we do similar checks for other contexts than smart-poster (local and external).

zolkis:
Creating `unknown` type is not supported currently, though we have an example for it. 
I make a separate issue for that as this needs discussion (we once discussed it and decided not to support that, only for reading). We need to put this in the spec, if that is the case.

Honry:
WPT tests added:
- smart-poster: https://chromium-review.googlesource.com/c/chromium/src/+/2049831
- local types: https://chromium-review.googlesource.com/c/chromium/src/+/1996946

beaufortfrancois:
@kenchris Does this look good to you as well?

index.html
```diff
@@ -2497,6 +2502,13 @@ <h3><dfn>Writing content</dfn></h3>
             </li>
           </ol>
         </li>
+        <li>
+          If |context| is `"smart-poster"` and |records| does not contain a
+          <a>URI record</a> for which the protocol is supported by the UA, or
```
leonhsl: I think we should reject if more than one URI records exist.

index.html
```diff
@@ -3602,13 +3634,24 @@ <h3>Parsing content</h3>
...
           </li>
           <li>
-            If |messageEnd| is `true`, abort these sub-steps.
+            If |messageEnd| is `true`,
+              <ol>
+                <li>
+                  If |context| is not `"smart-poster"`, abort these sub-steps.
+                </li>
+                <li>
+                  If |records| does not contain a <a>URI record</a>
```
leonhsl: Ditto. More than one URI record should NOT be OK.

index.html
```diff
@@ -2497,6 +2502,13 @@ <h3><dfn>Writing content</dfn></h3>
             </li>
           </ol>
         </li>
+        <li>
+          If |context| is `"smart-poster"` and |records| does not contain a
+          <a>URI record</a> for which the protocol is supported by the UA, or
+          if it contains more than one <a>type record</a>, <a>size record</a>
+          or <a>action record</a>, reject |promise| with {{TypeError}} and
```
leonhsl: Should we also reject if there are records other than URI record, title record, icon record, size record, type record, action record? e.g. an external type record is embedded.

index.html
```diff
@@ -3602,13 +3634,24 @@ <h3>Parsing content</h3>
...
+              <ol>
+                <li>
+                  If |context| is not `"smart-poster"`, abort these sub-steps.
+                </li>
+                <li>
+                  If |records| does not contain a <a>URI record</a>
+                  for which the protocol is supported by the UA, or if it
+                  contains more than one <a>type record</a>, <a>size record</a>
+                  or <a>action record</a>,
+                  discard all records in |records| and abort these sub-steps.
```
leonhsl: How would we do if |records| contains some other records than URI record, title record, icon record, size record, type record, action record? e.g. an external type record.

index.html
kenchris: Why does the UA need to support the protocol... That makes it impossible for writing an application for just configuring tags for other purposes, like if I write a generic smart poster app so create smart posters and to inspect them

index.html
```diff
@@ -3602,13 +3634,24 @@ <h3>Parsing content</h3>
...
           <li>
-            If |messageEnd| is `true`, abort these sub-steps.
+            If |messageEnd| is `true`,
+              <ol>
+                <li>
+                  If |context| is not `"smart-poster"`, abort these sub-steps.
+                </li>
+                <li>
+                  If |records| does not contain a <a>URI record</a>
+                  for which the protocol is supported by the UA, or if it
```
kenchris: Same

index.html

index.html

index.html
zolkis: The NFC spec allows that.

> There MAY be other records, which can be treated in an application-specific manner.

index.html
zolkis: The NFC spec allows that.

index.html

index.html

index.html
zolkis: Now passing `""`.

index.html
Honry: @zolkis, looks like Web NFC spec doesn't describe this, we should add it to the smart-poster definition.

zolkis: Makes sense, although we cannot aim to reproduce all details of the NFC Forum specs.
Normative text should be in algorithms, and in general it's not a problem if they are not duplicated in informative sections.

leonhsl: Is a smart-poster record allowed to embed another smart-poster record in the payload? I'm unsure how the NFC spec regulates about that.

zolkis: It doesn't regulate that. If you wonder how the context works with recursive data, well, the algorithms should work AFAICT, even when we embed into an external record a smart poster with an extra smart poster or external which again embed others. The contexts are stacked.

Honry: @zolkis, does it mean smart-poster allows another mime type? If so, following restriction should be removed.

> If ndef's typeNameField is 2 (MIME type record), then
> If context is equal to "smart-poster" and if running validate smart poster icon type on ndef's type returns false, throw a TypeError and abort these steps.

zolkis: > does it mean smart-poster allows another mime type? 

Yes, it does, but MIME type records have to be representing an icon, see below. 

> A Smart Poster may include an icon by including one or many MIME-
typed image records within the Smart Poster. If the device supports images, it SHOULD
select and display one of these, depending on the device capabilities. The device SHOULD
display only one. The Icon record is optional.

and,

> The Smart Poster may also contain a number of Icon records that have an image NDEF MIME
type [NDEF], [RFC2046]. For example, “image/jpeg”, “image/png”, etc. There might also be
animated Icon records, using common media types such as “video/mpeg”. A reader device
SHOULD display one (and only one) of these Icons to the user prior to acting on the URI record.
A reader device is not required to support any particular image format, but it is
RECOMMENDED that implementers stick to well-known, well-compressed formats such as
PNG [PNG] and JPEG [JPEG].

This pretty much implies that MIME type records in smart poster are _meant_ to represent icons. If it was otherwise, it would have had to be specified explicitly. Therefore I conclude the text should be OK.

Honry: Got it, thanks for your clarification.

leonhsl: By reading the quoted description for defining icon records, it looks to me that MIME records are NOT necessarily acting as icon records.

It just says a Smart Poster can include some MIME records to represent icons, without saying that MIME records must be for icons, i.e. they can be for other application-specific things as well. Otherwise, the spec should have made such a restriction explicitly.

zolkis: So how can you tell if a MIME record was meant as icon or as application-specific thing? I wager _that_ would have had to be explicitly specified by the NFC spec, because it's a confusing situation.

But if MIME type records are considered meant for icons, that is not confusing.

leonhsl: If the MIME record's mime type is something like "image/*" or "video/*" that can represent icon data, then it's an icon record. Otherwise, it might be representing some other application-specific data by using various mime types, e.g. JSON data by using "application/json".  If we restrict MIME type to be only for icon data, users lose all other possibilities which seems a surprise and not that friendly.

Now we all agree that the NFC spec allows Smart Poster to include any kind of records,  and it says explicitly that there must have one URI record referring to the smart poster's content and no other url record is allowed, then, when it comes to MIME record, it also should regulate explicitly if there is a similar restriction on that.

zolkis: Well, there can be multiple MIME type records, of which one should be selected by the readers to represent the smart poster icon. So all MIME type records are accepted that qualify for an icon. 

It's true the NFC spec leaves it a bit open what qualifies for an icon. But if we accept any MIME types in smart poster that are _not_ meant to be icons, that will be confusing for readers, since there is no algorithm specified by NFC Forum that would tell how to tell  icons from compatible, but not-meant-for-icon records.

Remember the smart poster use case is meant to describe content available on a network resource (like MIME type, but extended), specifying URL, title, size, type, icon, etc. There is no point having other MIME records than icons, since the content is supposed to be fetched from network and its type is given by the type record.

Apps could also use external records for strictly application purpose media included in smart poster, so there is no limitation

I suggest sticking to the specified algorithm.

leonhsl: Well I noticed this note in https://pr-preview.s3.amazonaws.com/zolkis/web-nfc/pull/536.html#dfn-validate-smart-poster-icon-type seems to be saying what I wanted to say :) 

> NOTE
The [NDEF-SMARTPOSTER] specification allows an "image" or "video" MIME type that is supported by applications. Examples are given as "image/jpeg", "image/png", "video/mpeg", but there are no hard requirements on accepted media types. If an application does not find a supported MIME type, it can choose to not display any icon for the smart-poster.

"image/*" and "video/*" are just some examples of mime types to represent icons, not hard requirements. We can allow all mime types and leave applications to choose from them.

zolkis: I meant that `image/jpeg` is an example but expect that all icon media types start with either `image/` or `video/`. Are you arguing for removing that restriction and accept any media types, remove the "validate icon type" steps and move that note?

leonhsl: Ah I see what "examples" does mean. And yes I feel that removing this restriction and leaving applications to do whatever they want might bring us (WebNFC) more flexibility, given that NFC forum spec does not restrict that explicitly.

zolkis: I need to run that through with security people. Allowing all MIME types when expecting an icon. Of course it's app responsibility. The NFC Forum is specific, as it allows "**_image_** NDEF MIME" and "**_animated_** Icon records, using common media types such as video/mpeg. 
I am not convinced we should leave it completely free.

leonhsl: Well I just feel we do not necessarily need to add a restriction that is not regulated by the NFC forum, i.e. image or video MIME records CAN serve as icon records but other MIME records should still be able to serve for any other purposes.

NFC forum might be specific that only `image` and `video` MIME records are allowed to represent icons, but it is NOT specific on that all MIME records must be for icons.

Security people might have more insights around this.

zolkis: Additional point: from purely use case point of view, other MIME records don't make sense, apps could include those other MIME records outside of the smart-poster. If they want other records inside the smart poster, they could use external or local. So I don't _think_ the NFC Forum _meant_ to allow other MIME records, i.e. I consider it a bug in the NFC Forum spec. Maybe members could ask? @beaufortfrancois ?

zolkis: The threat from security point of view would be this: since apps have to select one MIME record for icon, if the chosen MIME record is not truly compatible with the platform for representing as icon, it might be an attack vector. So restricting icons to certain types would make sense _for the Web_.

leonhsl: You mentioned before that NFC forum is saying
> There MAY be other records, which can be treated in an application-specific manner.

I figure this **_other records_** include all record types, not only external or local records?

leonhsl: I think restricting the mime type to be only "image/*" or "video/*" does NOT guarantee it must be truly compatible with the platform for icon representation.  The application is supposed to choose an appropriate icon record and even choose whether to display an icon or not.

zolkis: Yes, that is right.

Since this discussion refers writing, we should be extra careful (compared to reading) about what we allow to put there. IMHO it is better to restrict writing of known compatible MIME types for icons (since they figure in the NFC Forum spec).

On the reading side, I think if there is no extra security threat, we could read and expose all records that have been written (supposedly the writing end knew what they were doing). But it would be good to ask the NFC Forum in this matter.

zolkis: Outcome of discussion with security people:
- we should ask NFC Forum how to handle this corner case
- meanwhile, we could have no restrictions on reads (but implementation can harden handling of mime records handling even without the spec requiring it)
- on writes, the problem is that smart poster possibly has associated automatic actions on readers (any kind, not just Web NFC). Therefore allowing any MIME type may result in utilizing a gap in NFC Forum spec or its interpretation among reader implementations. We should try to minimize possible attack surface. So makes sense that Web NFC limits accept smart poster icon mime types to those that are known to be expected to work (image/jpeg, jpg, png etc, video/... etc), if that is not a major restriction for web apps (and should not be).

So I propose that for the moment we restrict on writes only, and we ask/get response from NFC Forum on what is safe to do here.

zolkis: However, we can interpret the NFC Forum spec also in a way that permits extra MIME type records. Note that currently we don't allow external and absolute-url records inside SP.

> There MAY be other records, which can be treated in an application-specific manner. For
example, some application might include a vCard contact card using the proper MIME type.
Applications MAY ignore any extra records inside the Smart Poster.

zolkis: For now removed the restrictions, but we need to track this. It would perhaps be better to limit now and relax later, than to restrict later.

zolkis:
This has been open for too long. It was not meant to be kept open until we finish all SP-related concerns, other PRs may also address the emerging issues.
Should we set a deadline for comments before merging?

index.html
```diff
@@ -2548,13 +2591,17 @@ <h3><dfn>Writing content</dfn></h3>
...
-            to one of the following algorithms and return the result.
+            Switching on |record|'s <a>recordType</a>, invoke one of the
+            following algorithms and return the result.
             If the algorithm throws an exception |e|, reject |promise| with |e|
             and abort these steps.
             <dl>
               <dt>"`empty`"</dt>
               <ul>
+                <li>
+                  If |context| is `"smart-poster"`,
```
Honry: This looks wrong that you restrict smart poster's creation to not include empty type, absolute-url type, external type and local type. Moreover, we should also check the  PAYLOAD field's size of smart-poster's action type and size type in creation algorithm.

zolkis: The NFC FOrum spec states that SP is basically a URL record with other records as metadata, either WKT or local type records.

Empty record and abs URL fall out from the primary intent of smart poster. While NFC Forum allows other records like ones containing vcards, we should not encourage all possible aberrations of "other records", when we have other mechanisms to include those records and when it is quite sure it will not interwork with other readers because it's not standard. If someone wants such a wildly more complex smart poster, why not use their own external type for that? 

Do we have interoperability tests of smart poster with absolute URL records and external records?

I can imagine local types should be allowed, and perhaps even external (though I'm not fully convinced on that). But empty and absolute URL? Why?

About checking size for action and size records, I agree. Moreover, it seems some readers expect the URL record to be the first record in the SP. I wonder should we include a clause in the spec that implementations should put the URL record first in the SP message?

leonhsl: I feel when the NFC forum spec (the standard) says 'other records', it just means **ALL other records**, if we want to make it clear that `empty` and `absolute-url` are not allowed, we might want to change the NFC forum spec first, which is source of truth for WebNFC. Although WebNFC provides some more of web things, I assume it's not expected to interpret the standard in its own way.

About the external type record, actually it is able to replace all other types of records if users want.

zolkis: We don't need to change the NFC Forum spec for defining what "other records" we accept in SP. The NFC Forum spec is rather undefined in this respect, but the intent and use case for SP is clear.
Web NFC implementation is an "application" from NFC Forum pov, and the spec says that applications MAY ignore other records in the SP.
The discussion here is what to allow there in Web NFC.

It does not make sense to allow just anything, as it was explained so many times in previous comments. Could you please show me a SP tag that contains absolute URL or empty records, and tell how many SP tags contain external or even other local type records? 

BTW there is a hard restriction on SP that's not captured in the Web NFC spec: if there are other URI records on the same level as SP, they MUST be ignored and SP MUST be used (i.e. SP overrides all other URI records). 

So absolute URL records MUST NOT be accepted at least.
Empty and unknown records make no sense to accept.
External and other local are debatable, but I have no hard opinion on that. We could accept them.
Other MIME records are fine and they are the primary examples in the SP spec.

I am curious what do you @kenchris and @beaufortfrancois think.

zolkis: Also, as already said above, it's better policy to restrict now (in debatable situations) and allow later than vice versa.

zolkis: Continued in #547.

index.html
```diff
@@ -2531,72 +2581,76 @@ <h3><dfn>Writing content</dfn></h3>
...
                 </li>
               </ul>
             </dl>
           </li>
+          <li>
+            If |record|'s <a>recordType</a> starts with a colon `U+003A` (`:`):
+            <ul>
+              <li>
+                If |record| is not a payload to another <a>NDEF record</a>
+                and if |context| is not `"smart-poster"`,
```
Honry: @zolkis, looks like we don't need to check `if |context| is not "smart-poster"`.

zolkis: I know that smart-poster contains other records, this is more like a shortcut (not to run a full check on containment since we already know the context). But it should be the first condition then.

zolkis: Usually "not a payload to another NDEF record" should be a defined algorithm, actually this might be an issue. We may want to extend the usage of `context` to help that. But frankly I don't like very much these meta-arguments (just for the sake of algorithms).

Honry: OK, make sense, thanks!

Honry: > Usually "not a payload to another NDEF record" should be a defined algorithm, actually this might be an issue. We may want to extend the usage of `context` to help that. But frankly I don't like very much these meta-arguments (just for the sake of algorithms).

why not just create an issue for further improvement? 😄

Honry (approved):
Looks pretty good now! @zolkis, thank you for your great effort!

leonhsl (approved):

index.html
zolkis: I think it would be enough to use 

> If |context| is `""` (i.e. |record| is not a payload to another  <a>NDEF record</a>), reject ...

index.html
```diff
@@ -575,7 +575,8 @@ <h4>
       <ul>
         <li>
           A single mandatory <a>URI record</a> that refers to the
-          <a>smart poster</a> content.
+          <a>smart poster</a> content. A <a>smart poster</a> overrides any
+          other <a>URI records</a> in an <a>NDEF message</a>.
```
kenchris: according to NFC spec link? or somthing like that

zolkis: The Smart Poster RTD, obviously. The whole section is about that one and it's stated/linked in the previous sentence. I don't expect to link it in every sentence in that section?

beaufortfrancois: Spec says 

> If an NDEF message contains one or multiple URI records (see [RTD_URI]) in addition to the
Smart Poster record at the top level (i.e., not nested), the Smart Poster record overrides the other
URI records. The NDEF application SHALL use only the Smart Poster record.

However, I'm not sure we should actually add this bit to the spec, as it is up to the app to actually consider a second URI record.

beaufortfrancois (approved):
LGTM with nit

index.html
beaufortfrancois: Saying it another way, I don't think Web NFC should delete/remove URI records from NDEF messages with smart poster.

zolkis: I thought Web NFC as an application from the NFC Forum point of view. As I see it,  an app doesn't really have the right to consider a second URI record in a smart poster, unless it wants to be non-conformant to the NFC spec.

kenchris:
I'm.with Francois on this. We can add a note about how web devs are supposed to handle this

zolkis:
I guess I disagree but commit.

zolkis:
Related to the disagreement on smart poster handling vs URI records, it goes quite deep.

I need more specific input on what is the consequence on the "check created records" and "check parsed records" and also the create steps (bs absolute URL in "smart-poster" context).

Since on creation we check there should be only one URI record, it seems logical to include the "why" in the smart poster informative section. (Whether it's a Note or text doesn't really matter, as the Smart Poster RTD is normative).

It makes no sense to check on unique type, size and action records, and ignore multiple URI records (including absolute URL records) which is as well a normative coming from SP RTD.

IMHO we have 2 options: either Web NFC doesn't do any checks on SP (leaving it to apps) or do all of the ones mandated by SP RTD.

kenchris (approved):

zolkis:
Thanks everyone! This was quite a ride. :)

PR #565, opened 2020-04-26
Fix #560: remove filtering

### LINKED ISSUE #560: Get rid of filtering
It turns out that filtering was not the greatest idea

* it is a bit confusing as it is not filtering the records (not sure that would actually be that useful without complex rules) and it just filters whether such a record exist 
* Newer standards like TNEP have other ways of selecting services as a way of filtering/communicating with IoT devices.

So let's get rid of it and simplify spec and implementation. @zolkis

zolkis:
@leonhsl @Honry  @DonnaWuDongxia  PTAL.

***
<a href="https://pr-preview.s3.amazonaws.com/zolkis/web-nfc/pull/565.html" title="Last updated on Apr 27, 2020, 10:33 AM UTC (d560974)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/565/9f15d9a...zolkis:d560974.html" title="Last updated on Apr 27, 2020, 10:33 AM UTC (d560974)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1065,15 +1065,17 @@
   <section> <h3>Save and restore game progress with an NFC tag</h3>
     <p>
       Filtering of relevant data sources can be done by the use of
-      the <a>NDEFScanOptions</a>. Below we use the custom record identifier
-      "`my-game-progress`" as a relative URL so that when we read the data, we
-      immediately update the game progress by issuing a write with a custom NDEF
-      data layout.
+      a custom record identifier, in this case "`my-game-progress`".
+      When we read the data, we immediately update the game progress by issuing
+      a write with a custom NDEF data layout.
     </p>
     <pre class="example">
       const reader = new NDEFReader();
-      await reader.scan({ id: "my-game-progress" });
+      await reader.scan();
       reader.onreading = async event => {
+        if (event.message.id !== "my-game-progress")
+          return;
+
         console.log(`Game state: ${ JSON.stringify(event.message.records) }`);
 
         const encoder = new TextEncoder();
@@ -1102,9 +1104,7 @@
     </p>
     <pre class="example">
       const reader = new NDEFReader();
-      await reader.scan({
-        mediaType: "application/*json"
-      });
+      await reader.scan();
       reader.onreading = event => {
         const decoder = new TextDecoder();
         for (const record of event.message.records) {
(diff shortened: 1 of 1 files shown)


kenchris (approved):

index.html
```diff
@@ -2191,60 +2164,18 @@ <h3>The <dfn>NDEFWriteOptions</dfn> dictionary</h3>
   <section data-dfn-for="NDEFScanOptions">
     <h3>The <dfn>NDEFScanOptions</dfn> dictionary</h3>
       <p>
-        To describe which messages an application is interested in, the
+        To pass an [= AbortSignal/aborted flag =], the
```
beaufortfrancois: Shall we express the intent of the aborted flag instead?
```suggestion
        To stop listening to NDEF messages, the

```

index.html
```diff
@@ -2191,60 +2164,18 @@ <h3>The <dfn>NDEFWriteOptions</dfn> dictionary</h3>
...
         <a>NDEFScanOptions</a> dictionary is used:
       </p>
       <pre class="idl">
         dictionary NDEFScanOptions {
-          USVString id;
-          USVString recordType;
-          USVString mediaType;
           AbortSignal? signal;
```
beaufortfrancois, line 2172: Shall we now make `AbortSignal` mandatory as user would not pass a dictionary otherwise?
```suggestion
          AbortSignal signal;
```

kenchris: That is a good questions. We never know how things develop in the future, so I think it is fine that it is optional. Or do you want `scan({})` to fail?

index.html
zolkis: OK! Thanks

index.html
zolkis: Formally it's still permissible to pass an empty dictionary as options, but if we want to stop that, yes, we could make it this way.  If we ever add back other options, we'll revert.

kenchris: Then the user code making use of the new API will fail outright in older browsers... seems bad

index.html
zolkis: What about this?
```
        To stop listening to NDEF messages, an [= AbortSignal/aborted flag =]
        can be passed to the <a>NDEFScanOptions</a> dictionary:
```

index.html
zolkis: OK, no big deal, could keep it nullable (force pushed only this revert).

beaufortfrancois: Let's keep it optional then as I don't feel strongly about it.

beaufortfrancois (approved):

zolkis:
Thank you!

PR #568, opened 2020-04-30
Fix #561: join NDEFReader and NDEFWriter

### LINKED ISSUE #561: Join Reader and Writer objects
I take the blame for this. It originally seemed like a good idea with separate objects, but I am getting more and more convinced that it was a bad idea.

* If we enhance to be able to select tag in the case of multiple tags read, this is required
* TNEP also requires this
* We can get rid of ignoreRead
* User is more in control of what happens.

Suggested name `NDEFPoller` or `NDEFPollerDevice`

zolkis:
A first attempt to join NDEFReader and NDEFWriter into NDEFReader.

Updated the reader slot  `[[\Signal]]` to `[[\ScanOptions]]` like it's been with `NDEFWriter`.

Haven't touched yet `ignoreRead`.

***
<a href="https://pr-preview.s3.amazonaws.com/zolkis/web-nfc/pull/568.html" title="Last updated on Apr 30, 2020, 4:45 PM UTC (a8f2148)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/568/c6eaae2...zolkis:a8f2148.html" title="Last updated on Apr 30, 2020, 4:45 PM UTC (a8f2148)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -943,12 +943,10 @@
   <section><h3>Feature support</h3>
     <p>
       Detecting if Web NFC is supported can be done by checking NDEFReader
-      and/or NDEFWriter objects. Note that this does not guarantee that NFC
-      hardware is available.
+      objects. Note that this does not guarantee that NFC hardware is available.
     </p>
     <pre class="example">
-      if ('NDEFReader' in window) { /* ... Scan NDEF Tags */ }
-      if ('NDEFWriter' in window) { /* ... Write NDEF Tags */ }
+      if ('NDEFReader' in window) { /* ... Scan and/or write NDEF Tags */ }
     </pre>
   </section>
 
@@ -957,8 +955,8 @@
       Writing a text string to an NFC tag is straightforward.
     </p>
     <pre class="example">
-      const writer = new NDEFWriter();
-      writer.write(
+      const ndef = new NDEFReader();
+      ndef.write(
         "Hello World"
       ).then(() => {
         console.log("Message written.");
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -943,12 +943,10 @@ <h4>
...
     <p>
       Detecting if Web NFC is supported can be done by checking NDEFReader
-      and/or NDEFWriter objects. Note that this does not guarantee that NFC
-      hardware is available.
+      objects. Note that this does not guarantee that NFC hardware is available.
     </p>
     <pre class="example">
-      if ('NDEFReader' in window) { /* ... Scan NDEF Tags */ }
-      if ('NDEFWriter' in window) { /* ... Write NDEF Tags */ }
+      if ('NDEFReader' in window) { /* ... Scan and/or write NDEF Tags */ }
```
kenchris, line 949: write *to*

tags not Tags

index.html
```diff
@@ -3304,19 +3278,20 @@ <h3><dfn>Writing content</dfn></h3>
             <ol>
               <li>
                 If |key| equals "`signal`" and |value| is not `undefined`, set
-                |reader|.<a>[[\Signal]]</a> to |value|.
+                |reader|.<a>[[\ScanOptions]]</a>'s |signal:AbortSignal| to |value|.
```
kenchris, line 3281: We are getting rid of scan options so why not just [[\ScanSignal]] to |value|

kenchris (approved):
Minor nits

index.html
zolkis: To be symmetric with write, and to be future-proof if we will eventually add other scan options later.

index.html
zolkis: Just copied the original text...

zolkis: fixed

kenchris:
Going to merge this so that I can update the write examples to better

PR #570, opened 2020-04-30
Remove ignoreRead and add info on how to implement manually

kenchris:
Also add more info about how writing works and an example on how to timeout a write, as well as how to construct a method that allows reading just one tag.

***
<a href="https://pr-preview.s3.amazonaws.com/kenchris/web-nfc/pull/570.html" title="Last updated on Apr 30, 2020, 9:27 PM UTC (ccb0bfb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/570/e5ecb1b...kenchris:ccb0bfb.html" title="Last updated on Apr 30, 2020, 9:27 PM UTC (ccb0bfb)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -950,9 +950,40 @@
     </pre>
   </section>
 
+  <section><h3>General information about writing data</h3>
+    <p>
+      Writing data is generally straight-forward, but there are a few
+      gotcha's around how writing works with NFC.
+    </p>
+    <p>
+      An NFC reader works by polling, so in order to be able to
+      write to a tag, a tag first needs to be found and read, which
+      means that polling needs to be initialized first.
+    </p>
+    <p>
+      If polling is not initiated already by first calling `scan()`,
+      then the `write()` method will initiate it temporarily until a
+      tag was found and read, and writing to it was attempted.
+    </p>
+    <p>
+      This means that the flow is that first a read it performed
+      as the tag is first found, then followed by a write.
+    </p>
+    <p>
+      This means that if `scan()` is running and you have an event
+      listener for the `reading` event, it will be dispatched once
+      during a `write()` operation, which might not be the intended
+      behavior.
+    </p>
+    <p>
+      In the following sections we will discuss how you can easily
+      deal with this behavior, but first a few simple examples.
+    </p>
+  </section>
+
   <section><h3>Write a text string</h3>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -950,9 +950,40 @@ <h4>
...
+    <p>
+      In the following sections we will discuss how you can easily
+      deal with this behavior, but first a few simple examples.
+    </p>
+  </section>
+
   <section><h3>Write a text string</h3>
     <p>
-      Writing a text string to an NFC tag is straightforward.
+      Writing a text string to an NFC tag is straight-forward.
```
zolkis, line 986: straightforward was fine
https://www.merriam-webster.com/dictionary/straightforward

index.html
```diff
@@ -1180,7 +1322,7 @@ <h4>
       };
 
       // Stop listening to NDEF messages after 3s.
-      setTimeout(() => controller.abort(), 3000);
+      setTimeout(() => controller.abort(), 3_000);
```
zolkis, line 1325: typo?

index.html
```diff
@@ -950,9 +950,40 @@ <h4>
...
+    <p>
+      An NFC reader works by polling, so in order to be able to
+      write to a tag, a tag first needs to be found and read, which
+      means that polling needs to be initialized first.
+    </p>
+    <p>
+      If polling is not initiated already by first calling `scan()`,
+      then the `write()` method will initiate it temporarily until a
+      tag was found and read, and writing to it was attempted.
+    </p>
```
zolkis, line 967: This is not explicitly said in the write() steps, but it's indeed implicitly assumed.

index.html
zolkis: Ah, https://github.com/tc39/proposal-numeric-separator

zolkis (approved):

index.html
kenchris: Then I need to fix in more places

index.html
kenchris: It's not said in write steps either which might be the reason why we have the read after writing if scan is not called

kenchris: Also we should probably talk about the polling and NFC Read operation as used in the NFC specs (at least tnep)

Issue #540, opened 2020-02-06
Provide TypeScript Declaration File for Web NFC


beaufortfrancois:
I'm sure TypeScript developers would appreciate if they'd had a `d.ts` file. 

For info, https://github.com/takefumi-yoshii/webnfc-media-memo/blob/master/src/types/global.ts  already exists.

kenchris:
Makes sense to have as part of this repo. I totally agree! @takefumi-yoshii do you mind contributing it to this repo?

takefumi-yoshii:
@beaufortfrancois @kenchris Of course! I'll contribute as soon as possible 😄

beaufortfrancois:
https://github.com/w3c/web-nfc/blob/gh-pages/web-nfc.d.ts is part of the repo. I'll close this issue.

PR #577, opened 2020-05-22
Add TypeScript Declaration File

takefumi-yoshii:
This PR is for https://github.com/w3c/web-nfc/issues/540

This type definitions referenced to WebIDL.
https://w3c.github.io/web-nfc/#actual-idl-index

@beaufortfrancois @kenchris Please check if there is any shortage.


web-nfc.d.ts
```diff
@@ -0,0 +1,90 @@
+// Type definitions for Web NFC
+// Project: https://github.com/w3c/web-nfc
+// Definitions by: Takefumi Yoshii <https://github.com/takefumi-yoshii>
+// TypeScript Version: 3.9
+
+// This type definitions referenced to WebIDL.
+// https://w3c.github.io/web-nfc/#actual-idl-index
+
+interface Window {
+  NDEFMEssage: NDEFMEssage
```
kenchris: Spelling mistake... uppercase E

web-nfc.d.ts
```diff
@@ -0,0 +1,90 @@
...
+  scan: (options?: NDEFScanOptions) => Promise<void>
+}
+
+interface Window {
+  NDEFReadingEvent: NDEFReadingEvent
+}
+declare class NDEFReadingEvent extends Event {
+  constructor(type: string, readingEventInitDict: NDEFReadingEventInit)
+  serialNumber: string
+  message: NDEFMEssage
```
kenchris: Spelling uppercase E

kenchris (approved):
Approve given the mistakes are fixed.

I am not a TS expert so maybe one should have a look

web-nfc.d.ts
```diff
@@ -0,0 +1,90 @@
...
+
+// This type definitions referenced to WebIDL.
+// https://w3c.github.io/web-nfc/#actual-idl-index
+
+interface Window {
+  NDEFMessage: NDEFMessage
+}
+declare class NDEFMessage {
+  constructor(messageInit: NDEFMessageInit)
+  readonly records: NDEFRecord[]
```
fsubal: [Question] IDL says it's `FrozenArray<T>` rather than `sequence`. Should this be `ReadonlyArray<Readonly<NDEFRecord>>` then ?

web-nfc.d.ts
```diff
@@ -0,0 +1,90 @@
...
+interface NDEFWriteOptions {
+  ignoreRead?: boolean
+  overwrite?: boolean
+  signal?: AbortSignal
+}
+interface NDEFScanOptions {
+  id?: string
+  recordType?: string
+  mediaType?: string
+  signal?: any
```
fsubal: Why is this `any` instead of `AbortSignal` ?

web-nfc.d.ts
fsubal: I don't know what kind of type is used if typings were auto generated from IDL, and it's okay to keep this if they would use `T[]`.

manekinekko: I think `ReadonlyArray<NDEFRecord>` should be fine.

web-nfc.d.ts
manekinekko: Correct, this should be `AbortSignal`

manekinekko:
I left some comments but the PR LGTM.

web-nfc.d.ts
takefumi-yoshii: 👍

takefumi-yoshii:
@fsubal @manekinekko Thanks for review! Fixed some mistake.

fsubal (approved):
Looks good to me!

web-nfc.d.ts
```diff
@@ -0,0 +1,90 @@
...
+}
+
+interface Window {
+  NDEFReader: NDEFReader
+}
+declare class NDEFReader extends EventTarget {
+  constructor()
+  onerror: (error: any) => void
+  onreading: (event: NDEFReadingEvent) => void
+  scan: (options?: NDEFScanOptions) => Promise<void>
```
beaufortfrancois, line 64: The promise may reject with a DOMException error. Can we handle this?

web-nfc.d.ts
```diff
@@ -0,0 +1,90 @@
...
+  constructor(type: string, readingEventInitDict: NDEFReadingEventInit)
+  serialNumber: string
+  message: NDEFMessage
+}
+interface NDEFReadingEventInit extends EventInit {
+  serialNumber?: string
+  message: NDEFMessageInit
+}
+
+interface NDEFWriteOptions {
```
beaufortfrancois, line 80: Is there a way to add default values for `ignoreRead` and `overwrite` as in the spec?
```
  boolean ignoreRead = true;
  boolean overwrite = true;
```

web-nfc.d.ts
```diff
@@ -0,0 +1,90 @@
...
+
+interface Window {
+  NDEFWriter: NDEFWriter
+}
+declare class NDEFWriter {
+  constructor()
+  write: (
+    message: NDEFMessageSource,
+    options?: NDEFWriteOptions
+  ) => Promise<void>
```
beaufortfrancois, line 54: The promise may reject with a DOMException error. Can we handle this?

web-nfc.d.ts
```diff
@@ -0,0 +1,90 @@
...
+  readonly lang?: string
+  toRecords?: () => NDEFRecord[]
+}
+declare interface NDEFRecordInit {
+  recordType: string
+  mediaType?: string
+  id?: string
+  encoding?: string
+  lang?: string
+  data?: NDEFRecordDataSource
```
beaufortfrancois, line 41: Those are not optional in the `dictionary NDEFRecordInit`.

web-nfc.d.ts
```diff
@@ -0,0 +1,90 @@
...
+}
+declare class NDEFRecord {
+  constructor(recordInit: NDEFRecordInit)
+  readonly recordType: string
+  readonly mediaType?: string
+  readonly id?: string
+  readonly data?: DataView
+  readonly encoding?: string
+  readonly lang?: string
+  toRecords?: () => NDEFRecord[]
```
beaufortfrancois, line 33: Shall we use `ReadonlyArray` as well there?

beaufortfrancois (commented):
Thanks for tackling this @takefumi-yoshii!

Here are some nits below. Note that I'm no TS expert ;)

web-nfc.d.ts
kenchris: It's not FrozenArray though, just a sequence

web-nfc.d.ts
kenchris: I don't believe so no

https://github.com/microsoft/TypeScript/issues/13219

web-nfc.d.ts
takefumi-yoshii: Unfortunately, it is not possible to determine the type of error.
We have to handle with error at the runtime level, not the type level.

web-nfc.d.ts
takefumi-yoshii: This cannot be achieved.Expressed as a type that has an initial value,It causes the following problems.

```typescript
const options: NDEFWriteOptions = {}
const cond = options.ignoreRead // cond eq "undefined"
if (cond) {
  // The type tells "true", but ...
  console.log("Can't reach")
}
```

web-nfc.d.ts
takefumi-yoshii:
```typescript
const myRecord = new NDEFRecord({ recordType: 'text' })
// Error! Property 'data' is missing in type '{ recordType: string; }'
// but required in type 'NDEFRecordInit'.
```

Above compile error is correct? If so, I should fix with another PR.

web-nfc.d.ts
takefumi-yoshii: Looking at WebIDL, it looks like a sequence.

takefumi-yoshii:
@beaufortfrancois  Thank you for the review! I returned a comment.

web-nfc.d.ts
Xample: I confirm we cannot "yet"

PR #586, opened 2020-06-24
Improve Security and Privacy section

kenchris:
***
<a href="https://pr-preview.s3.amazonaws.com/kenchris/web-nfc/pull/586.html" title="Last updated on Jun 25, 2020, 12:32 PM UTC (5949ed7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/586/b73b57e...kenchris:5949ed7.html" title="Last updated on Jun 25, 2020, 12:32 PM UTC (5949ed7)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4411,7 +4411,7 @@
     <p>
       Data written is readable by other apps and sites with granted read access
       Any NDEF data on a tag can be read by any app or web site with the proper
-      access, so if that is not intended then the data should be encoded in a
+      access, so if that is not intended then the data should be encrypted in a
       secure manner that only who is supposed to read it can.
     </p>
     <h4>Multiple tags may be within the reading field at the same time</h4>
--- a/index.html
+++ b/index.html
@@ -4435,6 +4435,9 @@
       do so by cloning the data of the original tag and modifying it -
       either by changing the URL to load a malicious app/site, or by
       changing the data to inject malicious data in the right app/site.
+      Example: the tag is supposed to take you to <a href="https://example.com">https://example.com</a> but is modified
+      to take you to <a href="https://exаmple.com">https://exаmple.com</a> (that is with a Cyrillic а) - it looks legitimate
+      and you might now to giving sensitive data to a malicious site.
     </p>
     <p>
       Loading web sites from a tag is outside the scope of Web NFC, but it
--- a/index.html
+++ b/index.html
@@ -4431,7 +4431,10 @@
     <p>
       There is an attack vector, where someone places another malicious NFC
       tag/sticker on top of a legitimate tag, in order to load the wrong
-      app/site, or inject wrong data into the right app/site.
+      app/site, or inject wrong data into the right app/site. They can
+      do so by cloning the data of the original tag and modifying it -
+      either by changing the URL to load a malicious app/site, or by
+      changing the data to inject malicious data in the right app/site.
     </p>
     <p>
       Loading web sites from a tag is outside the scope of Web NFC, but it
--- a/index.html
+++ b/index.html
@@ -4387,7 +4387,7 @@
     <p>
       When a site has access to read NFC content, then the data of the scanned
       tags is shared with the site, in a similar way to uploading files and images.
-      The user has to trust that the site handles this data properly and in the
+      As with any site, it is up to the user whether to trust that the site handles this data properly and in the
       intended manner.
     </p>
     <h4>A site may modify and overwrite data of tags that are not made read-only</h4>
--- a/index.html
+++ b/index.html
@@ -4352,6 +4352,9 @@
 <section> <h2 id="security">Security and Privacy</h2>
   <section> <h3>Chain of trust</h3>
     <p>
+      Implementations need to make sure that when the user authorizes a method which is part of the Web NFC API, only that action is run, without side effects.
+    </p>
+    <p>
       By default, NDEF doesn't provide any way to make the content trusted
       beyond allowing tags to be made permanently read-only after writing
       data to them. This can even be done from a factory setting.
--- a/index.html
+++ b/index.html
@@ -4352,17 +4352,13 @@
 <section> <h2 id="security">Security and Privacy</h2>
   <section> <h3>Chain of trust</h3>
     <p>
-      Web sites and applications using Web NFC are not trusted.
-      This means that the user needs to be made aware of what NFC functionality
-      a web page intends to do. Implementations need to
-      make sure that when the user authorizes a method of this API, only
-      that action is run, without side effects, and exactly in the context and
-      the number of times the user allows the execution of NFC related
-      operations, according to the algorithmic steps detailed in this
-      specification.
+      By default, NDEF doesn't provide any way to make the content trusted
+      beyond allowing tags to be made permanently read-only after writing
+      data to them. This can even be done from a factory setting.
     </p>
     <p>
-      Web NFC does not sign <a>NFC content</a>. In order to protect the
+      Data written by the use of this API is not signed or encrypted automatically,
+      which follows existing native NFC APIs. In order to protect the
       integrity and authenticity of NDEF messages, the NFC Forum introduced
       [[NDEF-SIGNATURE]]. Using <a>NDEF signature</a> and
       key management is the responsibility of the application.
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -4381,6 +4377,75 @@ <h3>Parsing content</h3>
...
+  <section> <h3>Things that users should be made aware of when using NFC</h3>
+    <p>
+      This section details some of the things that users ought to be aware
+      of when using NFC. It is recommendated that implementations help educate
+      the users of given facts before or when related NFC actions are performed. 
+    </p>
+    <h4>Data that is read is shared with site</h4>
+    <p>
+      When a site has access to read NFC content, then the data of the scanned
+      tags is shared with the site, in a similar way to upload files and images.
```
kenchris:
```suggestion
      tags is shared with the site, in a similar way to uploading files and images.
```

zolkis:
@ereshetova could you please take a look?

index.html
```diff
@@ -4381,6 +4377,75 @@ <h3>Parsing content</h3>
...
+    <p>
+      A fixed tag may encode its ID or location in the data, meaning that reading
+      it exposes that information to the site, which then can deduct the location
+      the read took place. That combined with being logged into a service, can
+      share your location data with the site.
+    </p>
+    <p>
+      Data written is readable by other apps and sites with granted read access
+      Any NDEF data on a tag can be read by any app or web site with the proper
+      access, so if that is not intended then the data should be encoded in a
```
ereshetova: I would suggest the term "encrypted" and not "encoded" since encoding usually does not protect against confidentiality disclosure

ereshetova (commented):
Looks good to me overall

index.html
```diff
@@ -4381,6 +4377,75 @@ <h3>Parsing content</h3>
...
```
kenchris:
```suggestion
      access, so if that is not intended then the data should be encrypted in a
```

index.html
```diff
@@ -4352,17 +4352,13 @@ <h3>Parsing content</h3>
...
   <section> <h3>Chain of trust</h3>
     <p>
-      Web sites and applications using Web NFC are not trusted.
-      This means that the user needs to be made aware of what NFC functionality
-      a web page intends to do. Implementations need to
-      make sure that when the user authorizes a method of this API, only
-      that action is run, without side effects, and exactly in the context and
-      the number of times the user allows the execution of NFC related
-      operations, according to the algorithmic steps detailed in this
-      specification.
```
zolkis, line 4362: I would keep this, as it provides clarity for implementations. @ereshetova ?

index.html
```diff
@@ -4352,17 +4352,13 @@ <h3>Parsing content</h3>
...
+      By default, NDEF doesn't provide any way to make the content trusted
+      beyond allowing tags to be made permanently read-only after writing
+      data to them. This can even be done from a factory setting.
     </p>
```
zolkis, line 4361: We can move this to the beginning (and keep the previous, deleted paragraph, too?)

index.html
```diff
@@ -4381,6 +4377,75 @@ <h3>Parsing content</h3>
...
+      This section details some of the things that users ought to be aware
+      of when using NFC. It is recommendated that implementations help educate
+      the users of given facts before or when related NFC actions are performed. 
+    </p>
+    <h4>Data that is read is shared with site</h4>
+    <p>
+      When a site has access to read NFC content, then the data of the scanned
+      tags is shared with the site, in a similar way to uploading files and images.
+      The user has to trust that the site handles this data properly and in the
+      intended manner.
```
zolkis, line 4394: Not sure if the "user has to trust" is proper language for security :)

index.html
kenchris: Some of it was added elsewhere... and it is implied that you need to implement the algorithms as described. I also find it hard to understand

index.html
kenchris: will have to? :)

index.html
```diff
@@ -4381,6 +4377,75 @@ <h3>Parsing content</h3>
...
+    <h4>A site may modify and overwrite data of tags that are not made read-only</h4>
+    <p>
+      Deployed NFC solutions, like tags in stores etc, should always be made
+      read-only in order to ensure they are not modified by mistake or as part
+      of a malicious act.
+    </p>
+    <p>
+      Private tags and stickers are often unlocked (writable) from the factory
+      and the user should be aware that such tags might be overwritten/modified
+      by scanning them.
```
zolkis, line 4405: by scanning?

index.html
```diff
@@ -4381,6 +4377,75 @@ <h3>Parsing content</h3>
...
+    </p>
+    <h4>Reading a fixed (e.g. mounted) tag may expose reading location</h4>
+    <p>
+      A fixed tag may encode its ID or location in the data, meaning that reading
+      it exposes that information to the site, which then can deduct the location
+      the read took place. That combined with being logged into a service, can
+      share your location data with the site.
```
zolkis, line 4412: if the site planted the id or the tag  - it has to have prior knowledge about the tag in order to determine location

index.html
kenchris: its called `.scan()` and you touch it... suggestions?

index.html
```diff
@@ -4381,6 +4377,75 @@ <h3>Parsing content</h3>
...
+    <p>
+      This is mostly useful for payment cards and travel cards that are read
+      by external hardware and thus not a use-case for Web NFC. For Web NFC,
+      we do not allow reading when there are multiple tags available.
+    </p>
+    <p>
+      There is an attack vector, where someone places another malicious NFC
+      tag/sticker on top of a legitimate tag, in order to load the wrong
+      app/site, or inject wrong data into the right app/site.
+    </p>
```
zolkis, line 4444: The sites should be prepared to read _any_ tags - there is no such thing as "reading an expected tag". So I am not sure if this threat is real.

index.html
kenchris: that is implied. I am trying to keep this simple

index.html
kenchris: sure there is... you can clone a tag and change the url by adding some russian latin look alike chars in the url instead... or just change the data

zolkis (commented):
I will leave the proper review to @ereshetova , just made comments.
Where I have not commented, I agree with the changes/additions. Thanks!

index.html
```diff
@@ -4381,6 +4377,75 @@ <h3>Parsing content</h3>
...
+    <p>
+      Private tags and stickers are often unlocked (writable) from the factory
+      and the user should be aware that such tags might be overwritten/modified
+      by scanning them.
+    </p>
+    <h4>Reading a fixed (e.g. mounted) tag may expose reading location</h4>
+    <p>
+      A fixed tag may encode its ID or location in the data, meaning that reading
+      it exposes that information to the site, which then can deduct the location
```
kenchris:
```suggestion
      it exposes that information to the site that knows the physical location of the tag, which then can deduct the location
```

index.html
zolkis: How do you define "the wrong app/site" or "wrong data"? Looks like you have some transactional use case in mind. It's not clear what is the threat.

index.html
zolkis: That's an important distinction, since this threat is not generic: in the vast majority of the cases sites won't deduct location by reads. The precondition is important for the attack.

index.html
zolkis: Maybe "tags might be overwritten/modified by this API"?
since it's possible to just read.

index.html
```diff
@@ -4381,6 +4377,75 @@ <h3>Parsing content</h3>
...
+    </p>
+    <p>
+      This is mostly useful for payment cards and travel cards that are read
+      by external hardware and thus not a use-case for Web NFC. For Web NFC,
+      we do not allow reading when there are multiple tags available.
+    </p>
+    <p>
+      There is an attack vector, where someone places another malicious NFC
+      tag/sticker on top of a legitimate tag, in order to load the wrong
+      app/site, or inject wrong data into the right app/site.
```
kenchris:
```suggestion
      app/site, or inject wrong data into the right app/site. They can
      do so by cloning the data of the original tag and modifying it -
      either by changing the URL to load a malicious app/site, or by
      changing the data to inject malicious data in the right app/site.
```

index.html
zolkis: The trust relationship between the user and the site needs to be established independently of using this API.
(complex, but expresses the point)

index.html
kenchris: maybe it is supposed to load www.skat.dk but it loads www.skаt.dk (that is with cyrillic а) ... looks like you loaded the right site and now you make a payment or share data you are not supposed to

zolkis (approved):
Small improvements could be made, based on the comments.

index.html
zolkis: So the fraud is with the tags themselves, since the URL shown in the permission prompt or in the read data will mislead the user on visual similarities. Could we filter out / warn about this kind of attack?
Is there a standard way to vet URLs?

index.html
```diff
@@ -4381,6 +4377,78 @@ <h3>Parsing content</h3>
...
+      by external hardware and thus not a use-case for Web NFC. For Web NFC,
+      we do not allow reading when there are multiple tags available.
+    </p>
+    <p>
+      There is an attack vector, where someone places another malicious NFC
+      tag/sticker on top of a legitimate tag, in order to load the wrong
+      app/site, or inject wrong data into the right app/site. They can
+      do so by cloning the data of the original tag and modifying it -
+      either by changing the URL to load a malicious app/site, or by
+      changing the data to inject malicious data in the right app/site.
```
kenchris, line 4440:
```suggestion
      changing the data to inject malicious data in the right app/site.
      Example: the tag is supposed to take you to <a href="https://example.com">https://example.com</a> but is modified
      to take you to <a href="https://exаmple.com">https://exаmple.com</a> (that is with a Cyrillic а) - it looks legitimate
      and you might now to giving sensitive data to a malicious site.
```

index.html
kenchris: The fraud is that you add a cloned and modified tag on top of a real one. People have done this with QR codes as well. It might take you to a whole different url and some users just don't notice. We don't allow reading if there are more tags close by so we can avoid this issue pretty much

index.html
```diff
@@ -4381,6 +4377,81 @@ <h3>Parsing content</h3>
...
+      and one of the tags can be selected as the tag to communicate with.
+    </p>
+    <p> 
+      Use cases for this could be having multiple smart cards (NFC based) in
+      your wallet and not wanting to take the card out.
+    </p>
+    <p>
+      This is mostly useful for payment cards and travel cards that are read
+      by external hardware and thus not a use-case for Web NFC. For Web NFC,
+      we do not allow reading when there are multiple tags available.
```
kenchris:
```suggestion
      we do not allow reading when there are multiple tags available, preventing the following attack vector.
```

index.html
zolkis: We can avoid the overlapping, but I wonder could we also warn about _possible_ fake URLs even when reading single tags? It is a well known attack, so there might be some standard solution to that already.

index.html
kenchris: but I think it is not clear to users that a site might actually write doing a touch/scan

index.html
kenchris: Sounds very cryptic to me. @anssiko

index.html
kenchris: Chrome itself will soon do that! Like if you mix character sets like cyrillic and latin

index.html
zolkis: OK, do as you wish :) 
Though this sentence makes sense:
"Implementations need to make sure that when the user authorizes a method of this API, only that action is run, without side effects"

index.html
zolkis: IoW, the trust between the user and the site is separate from using Web NFC. 
First, the user needs to trust the site (including data handling), then allows using NFC (including NFC data handling).

index.html
zolkis: Then replace "by scanning them" with "by touching them".

zolkis:
Resolved the done conversations, 3 small things left to check/improve.

index.html
```diff
@@ -4381,6 +4377,81 @@ <h3>Parsing content</h3>
...
+    <p>
+      This section details some of the things that users ought to be aware
+      of when using NFC. It is recommendated that implementations help educate
+      the users of given facts before or when related NFC actions are performed. 
+    </p>
+    <h4>Data that is read is shared with site</h4>
+    <p>
+      When a site has access to read NFC content, then the data of the scanned
+      tags is shared with the site, in a similar way to uploading files and images.
+      The user has to trust that the site handles this data properly and in the
```
kenchris:
```suggestion
      As with any site, it is up to the user whether to trust that the site handles this data properly and in the
```

index.html
```diff
@@ -4352,17 +4352,13 @@ <h3>Parsing content</h3>
...
     <p>
-      Web sites and applications using Web NFC are not trusted.
-      This means that the user needs to be made aware of what NFC functionality
-      a web page intends to do. Implementations need to
-      make sure that when the user authorizes a method of this API, only
-      that action is run, without side effects, and exactly in the context and
-      the number of times the user allows the execution of NFC related
-      operations, according to the algorithmic steps detailed in this
-      specification.
+      By default, NDEF doesn't provide any way to make the content trusted
```
kenchris, line 4358:
```suggestion
      Implementations need to make sure that when the user authorizes a method which is part of the Web NFC API, only that action is run, without side effects.
    </p>
    <p>
      By default, NDEF doesn't provide any way to make the content trusted
```

PR #601, opened 2020-09-22
Join NDEFReader and NDEFWriter

zolkis:
Re-applying #568.

***
<a href="https://pr-preview.s3.amazonaws.com/zolkis/web-nfc/pull/601.html" title="Last updated on Nov 19, 2020, 7:08 PM UTC (e82d4b2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/601/934c053...zolkis:e82d4b2.html" title="Last updated on Nov 19, 2020, 7:08 PM UTC (e82d4b2)">Diff</a>

### DIFF
--- a/EXPLAINER.md
+++ b/EXPLAINER.md
@@ -92,11 +92,9 @@ when required, about granting access to Web NFC. It means sending and receiving
 info when users tap NFC devices can be done smoothly once permission is granted.
 
 ```js
-const reader = navigator.nfc.ndef;
-
 async function startScanning() {
-  await reader.scan();
-  reader.onreading = event => {
+  await navigator.nfc.ndef.scan();
+  navigator.nfc.ndef.onreading = event => {
     /* handle NDEF messages */
   };
 }
@@ -126,10 +124,8 @@ cases where the text might even be in UTF-16 do to existing real life
 implementations.
 
 ```js
-const reader = navigator.nfc.ndef;
-
-await reader.scan({ recordType: "example.com:smart-poster" });
-reader.onreading = event => {
+await navigator.nfc.ndef.scan();
+navigator.nfc.ndef.onreading = event => {
   const externalRecord = event.message.records.find(
     record => record.type == "example.com:smart-poster"
   );
@@ -163,10 +159,9 @@ abortController.signal.onabort = event => {
   // All NFC operations have been aborted.
 };
 
-const reader = navigator.nfc.ndef;
-await ndef.scan({ signal: abortController.signal });
+navigator.nfc.ndef.scan({ signal: abortController.signal });
 
-await ndef.push("foo", { signal: abortController.signal });
+navigator.nfc.ndef.write("foo", { signal: abortController.signal });
 
 document.querySelector("#abortButton").onclick = event => {
   abortController.abort();
@@ -222,17 +217,19 @@ object/namespace, like Web Bluetooth, Web USB, etc. Unlike these APIs, the newer
 Generic Sensor APIs don’t attach themselves to the navigator object/namespace
 and have separate and dedicated objects like Accelerometer, Gyroscope, etc.
 
-We decided to follow this newer pattern as it allows node.js to implement the
+The newer pattern allows node.js to implement the
 same API, and have it loaded as a separate module. People working on Web
 Assembly are also advocating for this patterns as it might be able to turn such
 globals into modules in the future, at least when accessed from WASM.
 
+However, since Web NFC is close to Web Bluetooth and Web USB, and because scan
+filters were abandoned after Origin Trials feedback, attaching to the navigator
+object seemed to be a proper alignment.
+
 ### Separate objects for reader/writer
 
-The reader and writer objects could have been merged into one single object, but
-as we allow multiple scans with filters to be active at the same time (maybe in
-multiple places/view of the app/site) then it makes more sense to be able to use
-separate objects.
+The reader and writer objects existed separately when scan filters were supported,
+but after filters have been abandoned, they have been merged into one single object.
 
 ## Considered alternatives
 
--- a/EXPLAINER.md
+++ b/EXPLAINER.md
@@ -92,9 +92,11 @@ when required, about granting access to Web NFC. It means sending and receiving
 info when users tap NFC devices can be done smoothly once permission is granted.
 
 ```js
+const reader = new NDEFReader();
+
 async function startScanning() {
-  await navigator.nfc.ndef.scan();
-  navigator.nfc.ndef.onreading = event => {
+  await reader.scan();
+  reader.onreading = event => {
     /* handle NDEF messages */
   };
 }
@@ -124,8 +126,10 @@ cases where the text might even be in UTF-16 do to existing real life
 implementations.
 
 ```js
-await navigator.nfc.ndef.scan();
-navigator.nfc.ndef.onreading = event => {
+const reader = new NDEFReader();
+
+await reader.scan({ recordType: "example.com:smart-poster" });
+reader.onreading = event => {
   const externalRecord = event.message.records.find(
     record => record.type == "example.com:smart-poster"
   );
@@ -159,9 +163,10 @@ abortController.signal.onabort = event => {
   // All NFC operations have been aborted.
 };
 
-navigator.nfc.ndef.scan({ signal: abortController.signal });
+const ndef = new NDEFReader();
+await ndef.scan({ signal: abortController.signal });
 
-navigator.nfc.ndef.write("foo", { signal: abortController.signal });
+await ndef.push("foo", { signal: abortController.signal });
 
 document.querySelector("#abortButton").onclick = event => {
   abortController.abort();
@@ -217,19 +222,17 @@ object/namespace, like Web Bluetooth, Web USB, etc. Unlike these APIs, the newer
 Generic Sensor APIs don’t attach themselves to the navigator object/namespace
 and have separate and dedicated objects like Accelerometer, Gyroscope, etc.
 
-The newer pattern allows node.js to implement the
+We decided to follow this newer pattern as it allows node.js to implement the
 same API, and have it loaded as a separate module. People working on Web
 Assembly are also advocating for this patterns as it might be able to turn such
 globals into modules in the future, at least when accessed from WASM.
 
-However, since Web NFC is close to Web Bluetooth and Web USB, and because scan
-filters were abandoned after Origin Trials feedback, attaching to the navigator
-object seemed to be a proper alignment.
-
 ### Separate objects for reader/writer
 
-The reader and writer objects existed separately when scan filters were supported,
-but after filters have been abandoned, they have been merged into one single object.
+The reader and writer objects could have been merged into one single object, but
+as we allow multiple scans with filters to be active at the same time (maybe in
+multiple places/view of the app/site) then it makes more sense to be able to use
+separate objects.
 
 ## Considered alternatives
 
--- a/EXPLAINER.md
+++ b/EXPLAINER.md
@@ -6,7 +6,7 @@ Authors:
 - Francois Beaufort
 
 Participate
-- https://github.com/w3c/web-nfc/issues/ 
+- https://github.com/w3c/web-nfc/issues/
 - https://lists.w3.org/Archives/Public/public-web-nfc/
 
 ## Introduction
@@ -93,14 +93,14 @@ info when users tap NFC devices can be done smoothly once permission is granted.
 
 ```js
 const reader = new NDEFReader();
- 
+
 async function startScanning() {
   await reader.scan();
   reader.onreading = event => {
     /* handle NDEF messages */
   };
 }
- 
+
 const nfcPermissionStatus = await navigator.permissions.query({ name: "nfc" });
 if (permissionStatus.state === "granted") {
   // NFC access was previously granted, so we can start NFC scanning now.
@@ -123,7 +123,7 @@ USB](https://wicg.github.io/webusb/) in that it offers secure low level access
 with no magic and instead exposes data as a DataView as well as requires text to
 be encoded and decoded using TextEncoder/TextDecoder, which allows handling
 cases where the text might even be in UTF-16 do to existing real life
-implementations. 
+implementations.
 
 ```js
 const reader = new NDEFReader();
@@ -162,13 +162,12 @@ const abortController = new AbortController();
 abortController.signal.onabort = event => {
   // All NFC operations have been aborted.
 };
- 
-const reader = new NDEFReader();
-await reader.scan({ signal: abortController.signal });
- 
-const writer = new NDEFWriter();
-await writer.push("foo", { signal: abortController.signal });
- 
+
+const ndef = new NDEFReader();
+await ndef.scan({ signal: abortController.signal });
+
+await ndef.push("foo", { signal: abortController.signal });
+
 document.querySelector("#abortButton").onclick = event => {
   abortController.abort();
 };
@@ -234,7 +233,7 @@ The reader and writer objects could have been merged into one single object, but
 as we allow multiple scans with filters to be active at the same time (maybe in
 multiple places/view of the app/site) then it makes more sense to be able to use
 separate objects.
- 
+
 ## Considered alternatives
 
 ### Restrict to NDEF and name accordingly
--- a/EXPLAINER.md
+++ b/EXPLAINER.md
@@ -128,7 +128,7 @@ implementations.
 ```js
 const ndef = new NDEFReader();
 
-await ndef.scan({ recordType: "example.com:smart-poster" });
+await ndef.scan();
 ndef.onreading = event => {
   const externalRecord = event.message.records.find(
     record => record.type == "example.com:smart-poster"
--- a/EXPLAINER.md
+++ b/EXPLAINER.md
@@ -92,7 +92,7 @@ when required, about granting access to Web NFC. It means sending and receiving
 info when users tap NFC devices can be done smoothly once permission is granted.
 
 ```js
-const reader = new NDEFReader();
+const reader = navigator.nfc.ndef;
 
 async function startScanning() {
   await reader.scan();
@@ -126,7 +126,7 @@ cases where the text might even be in UTF-16 do to existing real life
 implementations.
 
 ```js
-const reader = new NDEFReader();
+const reader = navigator.nfc.ndef;
 
 await reader.scan({ recordType: "example.com:smart-poster" });
 reader.onreading = event => {
@@ -163,7 +163,7 @@ abortController.signal.onabort = event => {
   // All NFC operations have been aborted.
 };
 
-const ndef = new NDEFReader();
+const reader = navigator.nfc.ndef;
 await ndef.scan({ signal: abortController.signal });
 
 await ndef.push("foo", { signal: abortController.signal });
--- a/EXPLAINER.md
+++ b/EXPLAINER.md
@@ -92,17 +92,17 @@ when required, about granting access to Web NFC. It means sending and receiving
 info when users tap NFC devices can be done smoothly once permission is granted.
 
 ```js
-const reader = new NDEFReader();
+const ndef = new NDEFReader();
 
 async function startScanning() {
-  await reader.scan();
-  reader.onreading = event => {
+  await ndef.scan();
+  ndef.onreading = event => {
     /* handle NDEF messages */
   };
 }
 
 const nfcPermissionStatus = await navigator.permissions.query({ name: "nfc" });
-if (permissionStatus.state === "granted") {
+if (nfcPermissionStatus.state === "granted") {
   // NFC access was previously granted, so we can start NFC scanning now.
   startScanning();
 } else {
@@ -126,10 +126,10 @@ cases where the text might even be in UTF-16 do to existing real life
 implementations.
 
 ```js
-const reader = new NDEFReader();
+const ndef = new NDEFReader();
 
-await reader.scan({ recordType: "example.com:smart-poster" });
-reader.onreading = event => {
+await ndef.scan({ recordType: "example.com:smart-poster" });
+ndef.onreading = event => {
   const externalRecord = event.message.records.find(
     record => record.type == "example.com:smart-poster"
   );
@@ -166,7 +166,7 @@ abortController.signal.onabort = event => {
 const ndef = new NDEFReader();
 await ndef.scan({ signal: abortController.signal });
 
-await ndef.push("foo", { signal: abortController.signal });
+await ndef.write("foo", { signal: abortController.signal });
 
 document.querySelector("#abortButton").onclick = event => {
   abortController.abort();
@@ -227,12 +227,10 @@ same API, and have it loaded as a separate module. People working on Web
 Assembly are also advocating for this patterns as it might be able to turn such
 globals into modules in the future, at least when accessed from WASM.
 
-### Separate objects for reader/writer
+### Common vs separate objects for reader/writer
 
-The reader and writer objects could have been merged into one single object, but
-as we allow multiple scans with filters to be active at the same time (maybe in
-multiple places/view of the app/site) then it makes more sense to be able to use
-separate objects.
+The reader and writer objects existed separately when scan filters were supported,
+but after filters have been abandoned, they have been merged into one single object.
 
 ## Considered alternatives
 
--- a/index.html
+++ b/index.html
@@ -980,14 +980,11 @@
 
   <section><h3>Feature support</h3>
     <p>
-      Detecting if Web NFC is supported can be done by checking the {{Navigator}}
+      Detecting if Web NFC is supported can be done by checking the {{NDEFReader}}
       object. Note that this does not guarantee that NFC hardware is available.
-      The example checks if {{NDEFReader}} is available.
     </p>
     <pre class="example">
-      if (navigator.nfc?.ndef) {
-        /* ... Scan and write NDEF Tags */
-      }
+      if ('NDEFReader' in window) { /* ... Scan and write NDEF Tags */ }
     </pre>
   </section>
 
@@ -1027,7 +1024,8 @@
       Writing a text string to an NFC tag is straightforward.
     </p>
     <pre class="example">
-      navigator.nfc.ndef.write(
+      const ndef = new NDEFReader();
+      ndef.write(
         "Hello World"
       ).then(() => {
         console.log("Message written.");
@@ -1043,8 +1041,9 @@
       we rely on async/await.
     </p>
     <pre class="example">
+      const ndef = new NDEFReader();
       try {
-        navigator.nfc.ndef.write({
+        ndef.write({
           records: [{ recordType: "url", data: "https://w3c.github.io/web-nfc/" }]
         });
       } catch {
@@ -1067,9 +1066,10 @@
       handler and one used specifically for a single write.
     </p>
     <pre class="example">
+      const ndef = new NDEFReader();
       let ignoreRead = false;
 
-      navigator.nfc.ndef.onreading(event => {
+      ndef.onreading(event => {
         if (ignoreRead) {
           return; // write pending, ignore read.
         }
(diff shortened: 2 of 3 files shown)


kenchris (approved):

beaufortfrancois:
Quick question: Is it clear for web developers that `onerror` will get fired only when scan has started.
I can see why some web developers would think this code would work.

```js
ndef.onerror = event => {
  /* Was there an error while writing? */
}
ndef.write('hello world');
```

zolkis:
Good question, we need to clarify that with the two objects merged.

zolkis:
So should we rename `onerror` to `onreaderror`, or get rid of it alltogether and rely on the Promise, like with write()?

index.html
```diff
@@ -1980,30 +1974,26 @@ <h2>The <dfn>record type</dfn> string</h2>
...
-      Promise&lt;undefined&gt; write(NDEFMessageSource message,
-                                     optional NDEFWriteOptions options={});
-    };
-
     [SecureContext, Exposed=Window]
     interface NDEFReader : EventTarget {
       constructor();
 
-      attribute EventHandler onerror;
+      attribute EventHandler onreadingerror;
```
beaufortfrancois: Nit: Move `onreadingerror` below `onreading`

index.html
```diff
@@ -2041,10 +2031,11 @@ <h2>The <dfn>record type</dfn> string</h2>
     The serial number usually consists of 4 or 7 numbers, separated by `:`.
   </p>
   <p>
-    The <dfn>NDEFWriter</dfn> is an object used for writing data to an <a>NFC tag</a>.
+    The <dfn>NDEFReader</dfn> is an object used for reading or writing data when
```
beaufortfrancois: I believe this is redundant with the first paragraph of "The NDEFReader object"

index.html
```diff
@@ -1980,30 +1974,26 @@ <h2>The <dfn>record type</dfn> string</h2>
...
 
-<section> <h2>The NDEFReader and NDEFWriter objects</h2>
-  The objects provide a way for the <a>browsing context</a> to
-  use NFC functionality.
-  They allow for writing <a>NDEF message</a>s to <a>NFC tag</a>s within range,
-  and to act on incoming <a>NDEF message</a>s from <a>NFC tag</a>s.
+<section> <h2>The NDEFReader object</h2>
+  <p>
+    The <dfn>NDEFReader</dfn> is an object that exposes NFC functionality to the
+    <a>browsing context</a>: reading data when a device, such as a tag, is within
```
beaufortfrancois: I believe we should explicitly mention NDEF Message for reading like we do for writing

index.html
```diff
@@ -981,12 +980,10 @@ <h4>
   <section><h3>Feature support</h3>
     <p>
       Detecting if Web NFC is supported can be done by checking NDEFReader
-      and/or NDEFWriter objects. Note that this does not guarantee that NFC
-      hardware is available.
+      objects. Note that this does not guarantee that NFC hardware is available.
```
beaufortfrancois: Nit: `object`, not `objects`.

index.html
```diff
@@ -981,12 +980,10 @@ <h4>
...
     </p>
     <pre class="example">
-      if ('NDEFReader' in window) { /* ... Scan NDEF Tags */ }
-      if ('NDEFWriter' in window) { /* ... Write NDEF Tags */ }
+      if ('NDEFReader' in window) { /* ... Scan or write NDEF Tags */ }
```
beaufortfrancois: Nit: `Scan and...`

beaufortfrancois (commented):
Would you mind also updating the the EXPLAINER.md file and other files where NDEFWriter is referenced?

index.html

index.html
zolkis: Removed.

index.html
zolkis: Done.

index.html

index.html

zolkis:
Amended the TypeScript definition to also remove `ignoreRead` and scan options no longer supported.

arskama:
> Quick question: Is it clear for web developers that `onerror` will get fired only when scan has started.
> I can see why some web developers would think this code would work.
> 
> ```js
> const ndef = new NDEFReader();
> ndef.onerror = event => {
>   /* Was there an error while writing? */
> }
> ndef.write('hello world');
> ```

@beaufortfrancois 
I would like to have a clarification on your comment (it s me not understanding :) )

On the chromium implementation, there are 2 places where onerror is fired.
https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/modules/nfc/ndef_reader.cc;l=217;drc=126a458910eb392beb0c56601c89233cd82d80b5

and

https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/modules/nfc/nfc_proxy.cc;l=122;drc=126a458910eb392beb0c56601c89233cd82d80b5

which are 2 different cases, right?
So I think we need two different handling for those, since the first one is not really a readingerror?

zolkis:
Since we merged NDEFReader (which had an `onerror` event) and NDEFWriter (which did not have it), someone may assume that in the joined NDEFReader `onerror` is also relevant to writing, which is not the case. Therefore we renamed `onerror` to `onreadingerror`.

In the impl `onerror` conveys the NFC read errors (either due to access problems, or technical faults) to the readers.

Looks like since filters are gone, we have only one reader, and `NFCProxy` could be simplified (`readers_` removed etc), in conformance with the new internal slots (`PendingReader` instead of `ActivatedReaderList`) of `NDEFReader`.

beaufortfrancois:
Argh I see now... You're right @arskama, the first one "NDEFReader::OnMojoConnectionError()" is not per-say a reading error but an internal communication issue that can happen in Chromium.
If I remember correctly though, there's no way to recover from this.

The old behaviour still triggered an error even though it was NOT specifically a reading error.

With that in mind, maybe keeping `onerror` makes sense.
For info, this will also get triggered if an internal error happens during a write NFC operation and `write()` promise will reject as well...

Let's keep `onerror`.
What do you folks think?

zolkis:
> Let's keep onerror. What do you folks think?

So that error could have happened with the write(), too.
When does that error come up typically? Can we handle that via the scan() and write() promise?

zolkis:
>  "NDEFReader::OnMojoConnectionError()" is not per-say a reading error but an internal communication issue that can happen in Chromium.

Do I read that correctly: it seems to reject the promise with that error and _also_ fire an `onerror`? Is the latter needed?
Edit: it's covering the case the error comes _after_ the promise is settled. But does that actually (typically) happen?

beaufortfrancois:
> > Let's keep onerror. What do you folks think?
> 
> So that error could have happened with the write(), too.
> When does that error come up typically? Can we handle that via the scan() and write() promise?

scan will succeed when scan has started successfully, but this error could happen way after.
re: write, it won't be an issue as if this error happens, the promise will reject right away, and the onerror event will also get fired.

> > "NDEFReader::OnMojoConnectionError()" is not per-say a reading error but an internal communication issue that can happen in Chromium.
> 
> Do I read that correctly: it seems to reject the promise with that error and _also_ fire an `onerror`? Is the latter needed?
> Edit: it's covering the case the error comes _after_ the promise is settled. But does that actually (typically) happen?

it is an internal issue which "shouldn't" happen but can happen.

zolkis:
OK, I will add a commit the revert to `onerror` and explain the behavior for reading and writing.

zolkis:
Question, @beaufortfrancois @kenchris 

With filters gone, we can still create multiple NDEFReader instances, each of which has a scan() method that operate on the same NFC HW and have an `onreading` event that can take callbacks.

How should we handle this? 
1. Allow all the NDEFReader object instances, but keep a single reader in the internal slot and add the event listeners to its event.
2. Allow all the object instance, and keep all of these in the `activated reader list`. In the impl, there is a single NFC proxy, though.
3. Allow a single instance of NDEFReader in the given context and decide how to expose it: provide a getter/factory method for it, or let it be a static object, e.g. `navigator.nfc` (we used to be there).

zolkis:
Sorry, I needed to amend since I forgot to update the explainer and the TS definitions.
So, as agreed, now there is only one NDEFReader object that is accessed via `navigator.nfc.ndef`.
NFC state could be also moved to the `navigator.nfc` object, but for the moment it looked good on the settings object.
However, I moved all relevant content (NFC state, handling NFC adapters, obtaining permission, handling visibility change. releasing NFC) under `navigator.nfc`.

EXPLAINER.md
```diff
@@ -92,15 +92,15 @@ when required, about granting access to Web NFC. It means sending and receiving
 info when users tap NFC devices can be done smoothly once permission is granted.
 
 ```js
-const reader = new NDEFReader();
- 
+const reader = navigator.nfc.ndef;
```
beaufortfrancois: I think it's fair to use `navigator.nfc.ndef` inline.

```js

async function startScanning() {
  await navigator.nfc.ndef.scan();
  navigator.nfc.ndef.onreading = event => {
    /* handle NDEF messages */
  };
}
```

EXPLAINER.md
```diff
@@ -123,10 +123,10 @@ USB](https://wicg.github.io/webusb/) in that it offers secure low level access
 with no magic and instead exposes data as a DataView as well as requires text to
 be encoded and decoded using TextEncoder/TextDecoder, which allows handling
 cases where the text might even be in UTF-16 do to existing real life
-implementations. 
+implementations.
 
 ```js
-const reader = new NDEFReader();
+const reader = navigator.nfc.ndef;
```
beaufortfrancois: Same here.
```js
await navigator.nfc.ndef.scan({ recordType: "example.com:smart-poster" });
navigator.nfc.ndef.onreading = event => {
  const externalRecord = event.message.records.find(
    record => record.type == "example.com:smart-poster"
  );

  let action, text;

  for (const record of externalRecord.toRecords()) {
    if (record.recordType == "text") {
      const decoder = new TextDecoder(record.encoding);
      text = decoder.decode(record.data);
    } else if (record.recordType == "act") {
      action = record.data.getUint8(0);
    }
  }

  console.log({text, action});
};
```

EXPLAINER.md
```diff
@@ -123,10 +123,10 @@ USB](https://wicg.github.io/webusb/) in that it offers secure low level access
...
 
 await reader.scan({ recordType: "example.com:smart-poster" });
```
beaufortfrancois: `recordType` has been deprecated. Let's use `  await navigator.nfc.ndef.scan();`

EXPLAINER.md
```diff
@@ -162,13 +162,11 @@ const abortController = new AbortController();
...
 };
- 
-const reader = new NDEFReader();
-await reader.scan({ signal: abortController.signal });
- 
-const writer = new NDEFWriter();
-await writer.push("foo", { signal: abortController.signal });
- 
+
+await navigator.nfc.ndef.scan({ signal: abortController.signal });
```
beaufortfrancois: I'd remove those two `await` otherwise `document.querySelector("#abortButton").onclick` declaration wouldn't be valid.

index.html
```diff
@@ -980,13 +979,14 @@ <h4>
...
-      and/or NDEFWriter objects. Note that this does not guarantee that NFC
-      hardware is available.
+      Detecting if Web NFC is supported can be done by checking the {{Navigator}}
+      object. Note that this does not guarantee that NFC hardware is available.
+      The example checks if {{NDEFReader}} is available.
     </p>
     <pre class="example">
-      if ('NDEFReader' in window) { /* ... Scan NDEF Tags */ }
-      if ('NDEFWriter' in window) { /* ... Write NDEF Tags */ }
+      if (navigator.nfc && navigator.nfc.ndef) {
```
beaufortfrancois:
```suggestion
      if (navigator.nfc?.ndef) {
```

index.html
```diff
@@ -1043,9 +1042,8 @@ <h4>
       we rely on async/await.
     </p>
     <pre class="example">
-      const ndef = new NDEFReader();
       try {
-        ndef.write({
+        navigator.nfc.ndef.write({
```
beaufortfrancois: Use `await` to catch exception.
```suggestion
        await navigator.nfc.ndef.write({
```

index.html
```diff
@@ -1082,14 +1079,15 @@ <h4>
       function write(data) {
         ignoreRead = true;
         return new Promise((resolve, reject) => {
-          reader.addEventListener("reading", event => {
+          navigator.nfc.ndef.addEventListener("reading", event => {
             // Check if we want to write to this tag, or reject.
```
beaufortfrancois, line 1064: Where do we check here?

index.html
```diff
@@ -1106,8 +1104,8 @@ <h4>
       a certain amount of time, then you time out.
     </p>
     <pre class="example">
-      const reader = new NDEFReader();
-      reader.onreading(event => console.log("We read a tag!"));
+      const ndef = navigator.nfc.ndef;
```
beaufortfrancois: May we try without the shortcut `const ndef` for consistency and use `navigator.nfc.ndef` everywhere?

index.html
```diff
@@ -1167,17 +1165,16 @@ <h4>
...
     <pre class="example">
-      const reader = new NDEFReader();
-
       function read() {
         return new Promise((resolve, reject) => {
           const ctlr = new AbortController();
           ctlr.signal.onabort = reject;
-          reader.addEventListener("reading", event => {
+          navigator.nfc.ndef.addEventListener("reading", event => {
             ctlr.abort();
```
beaufortfrancois, line 1155: I'm not sure about this. How does this work exactly?

index.html
```diff
@@ -2139,25 +2012,26 @@ <h2>The <dfn>record type</dfn> string</h2>
       </td>
      </tr>
      <tr>
-      <td><dfn>[[\ActivatedReaderList]]</dfn></td>
+      <td><dfn>[[\PendingRead]]</dfn></td>
       <td>empty <a>set</a></td>
       <td>
-        A <a>set</a> of {{NDEFReader}} instances.
+        A &lt;|promise:Promise|, |reader:NDEFReader|&gt; tuple where |promise|
+        holds a {{Promise}} and |reader| holds an {{NDEFReader}}.
```
beaufortfrancois, line 2085: Why not `pending` as in "PendingWrite" description?

index.html
```diff
@@ -1979,139 +1962,29 @@ <h2>The <dfn>record type</dfn> string</h2>
...
+    interface NFC {
+      readonly attribute NDEFReader ndef;
     };
   </pre>
-  <p>
-    The <dfn>NDEFMessageSource</dfn> is a union type representing argument types
-    accepted by the {{NDEFWriter/write()}} method.
+  <p>The
+    <dfn>ndef</dfn> property is an {{NDEFReader}} object that provides
+    <a>NDEF</a> tag read and write functionality.
```
beaufortfrancois: Shall we use NFC tags consistently?

web-nfc.d.ts
```diff
@@ -6,6 +6,12 @@
 // This type definitions referenced to WebIDL.
 // https://w3c.github.io/web-nfc/#actual-idl-index
 
+interface Navigator {
+  nfc: NFC
```
beaufortfrancois:
```suggestion
  readonly nfc: NFC
```

web-nfc.d.ts
```diff
@@ -6,6 +6,12 @@
...
+}
+declare class NFC {
+  ndef: NDEFReader
```
beaufortfrancois:
```suggestion
  readonly ndef: NDEFReader
```

beaufortfrancois (commented):
Here's my first feedback to get your thoughts.

beaufortfrancois:
What does happen with code below?
```js
const promise1 = navigator.nfc.ndef.scan();
const promise2 = navigator.nfc.ndef.scan();
```
Which `promise1` of `promise2` reject?
It is not clear to me.

Same for `write()`...

beaufortfrancois (commented):
We shall update `Separate objects vs navigator.nfc` and `Separate objects for reader/writer` sections in EXPLAINER.md

index.html
zolkis: True, changed. I was thinking not all NFC tags were NDEF, so using "NDEF tag" was more appropriate for NDEFReader.

index.html
zolkis: fixed

zolkis:
Summarized the discussion about constructors vs navigator.nfc in #602.
Also, some options to address remaining issues like the ones @beaufortfrancois mentioned above.

zolkis:
OK, now we have `navigator.nfc` that contains `ndef` of type `NDEFReader` (exposing scan(), write(), onreading, onreadingerror).

I will make a branch from here to explore option 2 from #602 [comment](https://github.com/w3c/web-nfc/issues/602#issuecomment-707945243). That is, `navigator.nfc` would contain `ontagfound`, `onerror`,  and `ndef` of type `NDEFReader` that contains `read()` and `write()`. 
Scanning (NFC polling) is separated from reads, `ignoreRead` is not needed any more since there is no automatic content read from tags, and internal slots for pending read and write are not needed, either. Reading and dispatching algorithms become much simpler. Also, this would be closer to the Android NFC API usage flow, and more future proof.

This would permit later improvements, such as the `ontagfound` event could contain NFC technologies supported, but since that is restricted to NDEF at the moment, we can use a simple event. However, we could use flags already available from Android, such as `writeable`, `formattable` etc.
Future NFC tech could get their API namespace under `navigator.nfc`, for instance `navigator.nfc.isodep` exposing `transceive()` etc. The same `navigator.nfc.ontagfound` could be used (with expanded content, so backwards compatible).
Also, a `navigator.nfc.ontagreleased` event could be added later, if a developer need arises (it's quite common in other APIs, but will only add here when there are clear use cases for it).

The minimal delta from this point that supports the current functionality would be quite small, but let's see how

zolkis:
Reverted attaching to `navigator`. That version is saved in PR #605

EXPLAINER.md
```diff
@@ -162,13 +162,12 @@ const abortController = new AbortController();
...
-await reader.scan({ signal: abortController.signal });
- 
-const writer = new NDEFWriter();
-await writer.push("foo", { signal: abortController.signal });
- 
+
+const ndef = new NDEFReader();
+await ndef.scan({ signal: abortController.signal });
+
+await ndef.push("foo", { signal: abortController.signal });
```
beaufortfrancois: It should  be `write` not `push`

index.html
zolkis: The example & comment was by @kenchris

EXPLAINER.md
```diff
@@ -93,14 +93,14 @@ info when users tap NFC devices can be done smoothly once permission is granted.
 
 ```js
```
beaufortfrancois: Use `const ndef` here and below.

EXPLAINER.md
```diff
@@ -93,14 +93,14 @@ info when users tap NFC devices can be done smoothly once permission is granted.
...
 async function startScanning() {
   await reader.scan();
   reader.onreading = event => {
     /* handle NDEF messages */
   };
 }
- 
+
 const nfcPermissionStatus = await navigator.permissions.query({ name: "nfc" });
 if (permissionStatus.state === "granted") {
```
beaufortfrancois: Use `nfcPermissionStatus`

EXPLAINER.md
```diff
@@ -123,7 +123,7 @@ USB](https://wicg.github.io/webusb/) in that it offers secure low level access
 with no magic and instead exposes data as a DataView as well as requires text to
 be encoded and decoded using TextEncoder/TextDecoder, which allows handling
 cases where the text might even be in UTF-16 do to existing real life
-implementations. 
+implementations.
 
 ```js
```
beaufortfrancois: Use `ndef` instead of `reader`.

EXPLAINER.md
```diff
@@ -234,7 +233,7 @@ The reader and writer objects could have been merged into one single object, but
 as we allow multiple scans with filters to be active at the same time (maybe in
 multiple places/view of the app/site) then it makes more sense to be able to use
 separate objects.
- 
+
 ## Considered alternatives
```
beaufortfrancois, line 235: Please remove `### Separate objects for reader/writer` section above.

beaufortfrancois (changes requested):
Here's my feedback for the explainer.
I'll now review spec changes.

EXPLAINER.md
zolkis: Yes.

zolkis: Wait, that's only a title for a problem (explaining the API choice). So what should be the new name?

EXPLAINER.md
zolkis: Hmm. Looks like these changes have been lost in the revert, need to re-apply.

index.html
```diff
@@ -1068,10 +1065,10 @@ <h4>
       handler and one used specifically for a single write.
     </p>
     <pre class="example">
-      const reader = new NDEFReader();
+      const ndef = new NDEFReader();
       let ignoreRead = false;
 
-      reader.onreading(event => {
+      ndef.onreading(event => {
```
beaufortfrancois: This is not valid in JS. Please do not forget to remove extra `)` at the end of this function.

```suggestion
      ndef.onreading = (event) => {

File: index.html [hunk 25]
```diff
@@ -1106,8 +1103,8 @@ <h4>
       a certain amount of time, then you time out.
     </p>
     <pre class="example">
-      const reader = new NDEFReader();
-      reader.onreading(event => console.log("We read a tag!"));
+      const ndef = new NDEFReader();
+      ndef.onreading(event => console.log("We read a tag!"));
```
beaufortfrancois:
```suggestion
      ndef.onreading = (event) => console.log("We read a tag!");

File: index.html [hunk 26]
```diff
@@ -2056,12 +2043,20 @@ <h2>The <dfn>record type</dfn> string</h2>
...
      </tr>
     </thead>
-    <tbody data-link-for="NDEFWriter">
+    <tbody data-link-for="NDEFReader">
+     <tr>
+      <tr>
+        <td><dfn>[[\ScanAbortSignal]]</dfn></td>
+        <td>`undefined`</td>
+        <td>
+          The {{NDEFScanOptions}}.signal to abort the operation.
```
beaufortfrancois: Shall we specify it's only about the scan operation?

index.html
```diff
@@ -3478,27 +3446,25 @@ <h3><dfn>Writing content</dfn></h3>
             <ol>
               <li>
                 If |key| equals "`signal`" and |value| is not `undefined`, set
-                |reader|.<a>[[\Signal]]</a> to |value|.
+                |reader|.<a>[[\ScanAbortSignal]]</a> to |value|.
```
beaufortfrancois: How about doing the same as NDEFWriter and not use ScanAbortSignal, simply the signal from the dictionary member?

image

beaufortfrancois (changes requested):
See my comments.

index.html
zolkis: Done

EXPLAINER.md
zolkis: Updated

index.html
zolkis: fixed, thanks

index.html
zolkis: Should we then also remove the internal slot? (my pleasure)

index.html
```diff
@@ -1106,8 +1103,8 @@ <h4>
       a certain amount of time, then you time out.
     </p>
     <pre class="example">
-      const reader = new NDEFReader();
-      reader.onreading(event => console.log("We read a tag!"));
+      const ndef = new NDEFReader();
+      ndef.onreading = event => console.log("We read a tag!"));
```
beaufortfrancois:
```suggestion
      ndef.onreading = event => console.log("We read a tag!");
```

beaufortfrancois (commented):
Re: https://github.com/w3c/web-nfc/pull/601#discussion_r525211173 We could remove it if we decide to remove NDEFWriteOptions as well from the slots.

index.html
zolkis: Done

zolkis:
>  We could remove it if we decide to remove NDEFWriteOptions as well from the slots.

done

beaufortfrancois:
> > We could remove it if we decide to remove NDEFWriteOptions as well from the slots.
> 
> done

Shall we remove WriteMessage as well then?
image

beaufortfrancois:
> OK, I will add a commit the revert to `onerror` and explain the behavior for reading and writing.

FYI I'm still seeing `readingerror` in latest patch

zolkis:
> BTW the linking is broken somewhere, as it is actually used in the "start the NFC write" algorithm.

Thanks for catching!

beaufortfrancois:
@zolkis To fix linking use `<a>[[\WriteOptions]]</a>` instead of `[[\WriteOptions]]` only. Same for `WriteMessage`.

zolkis:
> To fix linking use <a>[[\WriteOptions]]</a> instead of [[\WriteOptions]] only. Same for WriteMessage.

Thanks, they are fixed now. Also checked the other internal slots.

index.html
beaufortfrancois: It is still not fixed.

index.html
beaufortfrancois: @kenchris Do we need `catch(err => reject(err));`?

beaufortfrancois: Let's keep it like this.

web-nfc.d.ts
```diff
@@ -43,25 +43,18 @@ declare interface NDEFRecordInit {
...
-}
-
 interface Window {
   NDEFReader: NDEFReader
 }
 declare class NDEFReader extends EventTarget {
   constructor()
-  onerror: (error: any) => void
   onreading: (event: NDEFReadingEvent) => void
+  onreadingerror: (error: any) => void
```
beaufortfrancois: `onreadingerror` doesn't expose error or anything actually.

beaufortfrancois (approved):
With few nits, this change looks good to me.

index.html
kenchris: I think we need to remove the ctlr.signal.onabort because you don't need to reject if it actually succeeds getting a reading.

kenchris: It works the way that if you get an event it resolves with the data and aborts the reader (stops reading).

if it fails for some reason, it will just reject

kenchris: it might actually be working today with ctlr.signal.onabort = reject; because abort is running async code and only one of reject/resolve can be called (the next is ignored)

kenchris: Maybe we should explain how it works :)

beaufortfrancois: Please open a new PR for this ;)

web-nfc.d.ts
zolkis: should be `onreadingerror: (event: Event) => void` ?

web-nfc.d.ts
```diff
@@ -43,25 +43,18 @@ declare interface NDEFRecordInit {
...
-}
-
 interface Window {
   NDEFReader: NDEFReader
 }
 declare class NDEFReader extends EventTarget {
   constructor()
-  onerror: (error: any) => void
   onreading: (event: NDEFReadingEvent) => void
+  onreadingerror: (error: Event) => void
```
beaufortfrancois: We should ask TS experts but I'd go with what others do.

See https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/w3c-web-usb/index.d.ts#L110-L112 and https://github.com/DefinitelyTyped/DefinitelyTyped/blob/927064b9a3a5f1075f3560eefc923860be7e0b53/types/web-bluetooth/index.d.ts#L163

```
declare class NDEFReader extends EventTarget {
  constructor()
  onreading: (this: this, ev: NDEFReadingEvent) => any
  onreadingerror: (this: this, ev: Event) => any
```

zolkis: right, https://www.valentinog.com/blog/this/

zolkis: amended the TS commit

PR #610, opened 2020-11-25
Disallow explicitly non top-level browsing contexts

beaufortfrancois:
Spec says Web NFC is allowed only in a top-level browsing context, but `scan` and `write` algorithm don't include those steps explicitly. This PR fixes this.

PTAL @zolkis

***
<a href="https://pr-preview.s3.amazonaws.com/beaufortfrancois/web-nfc/pull/610.html" title="Last updated on Nov 25, 2020, 12:49 PM UTC (3142d03)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/610/8ae5979...beaufortfrancois:3142d03.html" title="Last updated on Nov 25, 2020, 12:49 PM UTC (3142d03)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2312,6 +2312,11 @@
             Let |p:Promise| be a new {{Promise}} object.
           </li>
           <li>
+            If not currently executing in the currently active <a>top-level
+            browsing context</a>, then reject |p| with and
+            {{"InvalidStateError"}} {{DOMException}} and return |p|.
+          </li>
+          <li>
             Let |message:NDEFMessageSource| be the first argument.
           </li>
           <li>
@@ -3408,6 +3413,11 @@
         <ol class=algorithm>
           <li>
             Let |p:Promise| be a new {{Promise}} object.
+          </li>
+          <li>
+            If not currently executing in the currently active <a>top-level
+            browsing context</a>, then reject |p| with and
+            {{"InvalidStateError"}} {{DOMException}} and return |p|.
           </li>
           <li>
             Let |reader:NDEFReader| be the {{NDEFReader}} instance.


zolkis (approved):
Makes sense. Though the general text about this is normative, the algorithms should quantify it.

kenchris (approved):

beaufortfrancois:
@domenic The [Keyboard Lock](https://wicg.github.io/keyboard-lock/#h-keyboard-lock) and [Contact Picker](https://wicg.github.io/contact-api/spec/#contacts-manager-select) Web APIs  reject with `InvalidStateError` when the current browsing context is not a top-level browsing context.

Shall we do the same here in Web NFC for consistency with the web platform, or shall we introduce a new error type for this specific type of error?
What are your thoughts?

zolkis:
My take: from this [keyboard lock issue discussion](https://github.com/WICG/keyboard-lock/issues/35) it seems the reason to choose `InvalidStateError` was specific to their case. I'd use `NotAllowedError` as suggested [in this comment](https://github.com/WICG/keyboard-lock/issues/35#issuecomment-387230960), or `SecurityError`, since it's a failure because of a security policy, and invalid state error is a bit confusing on what exactly it refers to (and where/how that state is defined).

On ContactPicker looks like a copy from the previous choice :). 
I'd avoid making a de-facto standard choice from an application specific choice. IMHO we need a more generic relevant error type. I agree we should do it in similar way in the Web Platform, but then this requires a discussion/decision in the Web Platform.

domenic:
Any of those makes sense. Inventing a new error doesn't seem like a great idea.

In general error types are not a big deal on the web platform. They are only useful when your function throws multiple different errors, *and* it's important for web developers to distinguish between those error types, and react differently to them in a programmatic manner (not just logging).

PR #454, opened 2019-11-27
Do not use 'any' to define NDEFRecordInit#data

beaufortfrancois:
This PR replaces NDEFRecord `any data` with `NDEFRecordSource data`. The new `NDEFRecordSource` is `typedef (DOMString or BufferSource or NDEFMessageInit)`

```js
const textRecord = new NDEFRecord({ recordType: "text", data: "foo" });
```

FIX #453

***
<a href="https://pr-preview.s3.amazonaws.com/beaufortfrancois/web-nfc/pull/454.html" title="Last updated on Nov 27, 2019, 8:27 AM UTC (8c08a71)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/454/61d614b...beaufortfrancois:8c08a71.html" title="Last updated on Nov 27, 2019, 8:27 AM UTC (8c08a71)">Diff</a>

PR #621, opened 2021-09-15
Add records depth check

### LINKED ISSUE #620: Invalid Web IDL loop between NDEFRecordInit and NDEFMessageInit
https://heycam.github.io/webidl/#idl-dictionaries says:

> The type of a dictionary member must not include the dictionary it appears on. A type includes a dictionary D if at least one of the following is true:
> 1. the type is D
> 1. the type is a dictionary that inherits from D
> 1. the type is a nullable type whose inner type includes D
> 1. the type is a sequence type or frozen array whose element type includes D
> 1. the type is a union type, one of whose member types includes D
> 1. the type is a dictionary, one of whose members or inherited members has a type that includes D
> 1. the type is record<K, V> where V includes D

#454 introduced a case of 4, 5 and 6, as right now `NDEFRecordInit.data` can be an `NDEFMessageInit`, which is a dictionary with a sequence of `NDEFRecordInit`s. This creates a loop that is not allowed by Web IDL and which can cause implementation issues like https://bugs.chromium.org/p/chromium/issues/detail?id=1242274

beaufortfrancois:
This PR aims to solve infinite Web IDL loop introduced by https://github.com/w3c/web-nfc/pull/454. It does so by using `any` for `NDEFRecordInit.data` and adding a depth-check step.

FIX https://github.com/w3c/web-nfc/issues/620

***
<a href="https://pr-preview.s3.amazonaws.com/beaufortfrancois/web-nfc/pull/621.html" title="Last updated on Sep 17, 2021, 9:50 AM UTC (2a78694)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/621/36f05d6...beaufortfrancois:2a78694.html" title="Last updated on Sep 17, 2021, 9:50 AM UTC (2a78694)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2523,9 +2523,15 @@
                 Increment |recordsDepth| by one.
               </li>
               <li>
-                If |recordsDepth| > `2`, [= exception/throw =] a {{TypeError}} and abort these steps.
+                If |recordsDepth| > `2`, [= exception/throw =] a {{TypeError}}
+                and abort these steps.
                 <p class="issue" data-number="620">
-                  Infinite loops are not allowed by Web IDL.
+                  To prevent infinite loops, Web IDL bans dictionaries from
+                  including themselves as members. As the `data` dictionaries
+                  will have to have been created by a developer before
+                  initiating the function call, these will in practice be
+                  finite, so we work around this by using `any` as the
+                  member type and enforcing limits on allowed sub-records.
                 </p>
               </li>
             </ul>
--- a/index.html
+++ b/index.html
@@ -1551,8 +1551,6 @@
       <a>NDEFRecord</a> interface:
     </p>
     <pre class="idl">
-      typedef (DOMString or BufferSource or NDEFMessageInit) NDEFRecordDataSource;
-
       [SecureContext, Exposed=Window]
       interface NDEFRecord {
         constructor(NDEFRecordInit recordInit);
@@ -1576,7 +1574,7 @@
         USVString encoding;
         USVString lang;
 
-        NDEFRecordDataSource data;
+        any data;
       };
     </pre>
     <p>
@@ -2376,7 +2374,7 @@
               <li>
                 Let |output| be the notation for the <a>NDEF message</a>
                 to be created by UA, as the result of invoking
-                <a>create NDEF message</a> with |message| and `""`.
+                <a>create NDEF message</a> with |message|, `""`, and 0.
                 If this throws an exception, reject |p| with that
                 exception and abort these steps.
               </li>
@@ -2481,8 +2479,9 @@
 
   <section><h3>Creating NDEF message</h3>
     <div>
-      To <dfn>create NDEF message</dfn> given a
-      |source:NDEFMessageSource| and |context:string|, run these steps:
+      To <dfn>create NDEF message</dfn> given |source:NDEFMessageSource|,
+      |context:string|, and |recordsDepth:unsigned short|,
+      run these steps:
       <ol class=algorithm id="create-web-nfc-message"
           data-link-for="NDEFMessageSource">
         <li>Switch on |source:NDEFMessageSource|'s type:
@@ -2520,6 +2519,15 @@
                 If |source|'s records [= list/is empty =], [= exception/throw =]
                 a {{TypeError}} and abort these steps.
               </li>
+              <li>
+                Increment |recordsDepth| by one.
+              </li>
+              <li>
+                If |recordsDepth| > `2`, [= exception/throw =] a {{TypeError}} and abort these steps.
+                <p class="issue" data-number="620">
+                  Infinite loops are not allowed by Web IDL.
+                </p>
+              </li>
             </ul>
             <dt>unmatched type</dt>
             <ul>
@@ -2539,8 +2547,8 @@
           <ol>
             <li>
               Let |ndef| be the result of running <a>create NDEF record</a>
-              given |record:NDEFRecordInit| and |context|, or make sure the
-              underlying platform provides equivalent values to |ndef|.
+              given |record:NDEFRecordInit|, |context|, and |recordsDepth| or
+              make sure the underlying platform provides equivalent values to |ndef|.
               If the algorithm throws an exception |e|, reject |promise| with
               |e| and abort these steps.
             </li>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1576,7 +1574,7 @@ <h4>
         USVString encoding;
         USVString lang;
 
-        NDEFRecordDataSource data;
+        any data;
```
kenchris: Maybe for documentation purposes like MDN we should add a note that the accepted types are actually DOMString, BufferSource and NDEFMessageInit and that currently all other values will be ignored?

index.html
```diff
@@ -2376,7 +2374,7 @@ <h3><dfn>Writing content</dfn></h3>
               <li>
                 Let |output| be the notation for the <a>NDEF message</a>
                 to be created by UA, as the result of invoking
-                <a>create NDEF message</a> with |message| and `""`.
+                <a>create NDEF message</a> with |message|, `""`, and 0.
```
kenchris: Shouldn't be back tick 0?

index.html
```diff
@@ -2520,6 +2519,15 @@ <h3><dfn>Writing content</dfn></h3>
                 If |source|'s records [= list/is empty =], [= exception/throw =]
                 a {{TypeError}} and abort these steps.
               </li>
+              <li>
+                Increment |recordsDepth| by one.
+              </li>
+              <li>
+                If |recordsDepth| > `2`, [= exception/throw =] a {{TypeError}} and abort these steps.
+                <p class="issue" data-number="620">
+                  Infinite loops are not allowed by Web IDL.
```
kenchris: They will never really be infinite in practice as they have to be described by a user, using JavaScript

kenchris (approved):

index.html
beaufortfrancois: I've added a comment.

index.html
beaufortfrancois: Fixed

index.html
beaufortfrancois: Would you update this text?

kenchris: To prevent infinite loops, Web IDL bans dictionaries from including themselves as members. As the data dictionaries will have to have been created by a developer before initiating the function call, these will in practice be finite, so we work around this by using `any` as the member type and enforcing limits on allowed sub-records.

beaufortfrancois: THanks!

kenchris:
Want me to merge? Or feel free to do it yourself

beaufortfrancois:
I'll wait for others to chime in before doing so. Thank you!

zolkis:
So the idea is to enforce a depth level hardcoded to 2 when creating NDEF messages from a source.

There is no depth restriction on reads, right? 

Coming from there, I was pondering whether is it worth to expose the max depth to developers (up to a hardcoded ceiling bigger than 2), as a depth of 2 seems a good Gaussian value, but there might be use cases for more. I guess if that is the case, we'd get issues and could extend the API later. I am fine with 2.

If all this is intended, LGTM.

reillyeon:
I think it's reasonable to allow a much deeper recursion depth (e.g. 32) on writes. On reads we don't need to worry because the depth is effectively limited by the length of the data received from the device.

index.html
```diff
@@ -2520,6 +2519,21 @@ <h3><dfn>Writing content</dfn></h3>
...
+              </li>
+              <li>
+                If |recordsDepth| > `2`, [= exception/throw =] a {{TypeError}}
+                and abort these steps.
+                <p class="issue" data-number="620">
+                  To prevent infinite loops, Web IDL bans dictionaries from
+                  including themselves as members. As the `data` dictionaries
+                  will have to have been created by a developer before
+                  initiating the function call, these will in practice be
+                  finite, so we work around this by using `any` as the
```
reillyeon: There are tricky ways that a developer can craft an object that changes as the algorithm processes it. What I would say here instead is that for practical purposes there is only so much room for storing an NDEF message and so we can apply a limit on the depth of nested messages which developers are unlikely to ever hit.

index.html
```diff
@@ -2804,8 +2818,9 @@ <h3><dfn>Writing content</dfn></h3>
 
     <section><h3>Mapping empty record to NDEF</h3>
       <div>
-        To <dfn>map empty record to NDEF</dfn> given a |record:NDEFRecordInit|
-        |ndef| and |context:string|, run these steps:
+        To <dfn>map empty record to NDEF</dfn> given |record:NDEFRecordInit|,
+        |ndef|, |context:string|, and |recordsDepth:unsigned short|,
```
reillyeon: Can we skip passing the depth to algorithms like this which don't use it?

index.html
```diff
@@ -3192,7 +3209,7 @@ <h3><dfn>Writing content</dfn></h3>
               <li>
                 Set the |ndef|'s <a>PAYLOAD field</a> to the result of running
                 the <a>create NDEF message</a> given |record|'s <a>data</a>
```
reillyeon, line 3207: `data` here is `any` while the "create NDEF message" record expects `NDEFMessageSource`. It feels like we need to acknowledge this but maybe not because those steps do try to match on the parameter type and have a default case which will throw a TypeError.

beaufortfrancois: I believe it's okay to leave it as it. As you said, an `unmatched type` will throw an error in the "create NDEF message" algorithm

index.html
beaufortfrancois: I've done this for `recordsDepth` and `context`.

index.html
beaufortfrancois: Applied

beaufortfrancois (commented):
I've changed limit to `32`.

index.html
```diff
@@ -2520,6 +2519,19 @@ <h3><dfn>Writing content</dfn></h3>
                 If |source|'s records [= list/is empty =], [= exception/throw =]
                 a {{TypeError}} and abort these steps.
               </li>
+              <li>
+                Increment |recordsDepth| by one.
+              </li>
+              <li>
+                If |recordsDepth| > `32`, [= exception/throw =] a {{TypeError}}
+                and abort these steps.
+                <p class="issue" data-number="620">
```
reillyeon: Should we be using `class="issue"` here? Merging this change will resolve the issue so this should probably be a regular note, with reference to the issue that caused the change.

reillyeon (approved):

index.html
beaufortfrancois: Done

Issue #626, opened 2021-10-15
Heads up: refs to `pagevisibility` (external now visibile) about to be removed


noamr:
See w3c/page-visibility#73
Most of the page visibility spec is moving into the HTML Spec (see whatwg/html#7153), and calls are to be made explicitly instead of the "hook" (external-now-visible-algorithm).

Since this spec appears to rely on this ref, opening this issue to track switching to being called directly.

beaufortfrancois:
Do you have examples of other specs that have switched to the HTML spec for page visibility?

noamr:
> Do you have examples of other specs that have switched to the HTML spec for page visibility?

Not yet
It's all quite new
I can assist with advice if someone starts a PR etc

yoavweiss:
It might be worthwhile to create a helper algorithm inside HTML that would enable specs that want to add processing when the page is hidden/visible to do that more easily (without mucking around with history traversal).

noamr:
> It might be worthwhile to create a helper algorithm inside HTML that would enable specs that want to add processing when the page is hidden/visible to do that more easily (without mucking around with history traversal).

I think if I add that algorithm I might as well add the 3 calls to the other specs

noamr:
See https://github.com/whatwg/html/pull/7238

noamr:
https://html.spec.whatwg.org/#page-visibility-change-steps

beaufortfrancois:
Thank you @noamr!
See https://github.com/w3c/web-nfc/pull/627

PR #627, opened 2021-11-04
Switch to HTML page visibility refs

beaufortfrancois:
This PR is an attempt to address https://github.com/w3c/web-nfc/issues/626 as the Page Visibility spec is moving to the HTML spec.

@noamr What do you think?

***
<a href="https://pr-preview.s3.amazonaws.com/beaufortfrancois/web-nfc/pull/627.html" title="Last updated on Nov 5, 2021, 12:13 PM UTC (4616ff0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/627/760581a...beaufortfrancois:4616ff0.html" title="Last updated on Nov 5, 2021, 12:13 PM UTC (4616ff0)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -135,7 +135,7 @@
   </style>
 </head>
 
-<body data-cite="MIMESNIFF SECURE-CONTEXTS PERMISSIONS ENCODING PAGE-VISIBILITY">
+<body data-cite="MIMESNIFF SECURE-CONTEXTS PERMISSIONS ENCODING">
 
 <!-- - - - - - - - - - - - - - - - Abstract - - - - - - - - - - - - - - - - -->
 <section id="abstract">
@@ -2167,7 +2167,7 @@
   <section><h3>Handling visibility change</h3>
     <p>
       When the user agent determines that the
-      <dfn data-cite="PAGE-VISIBILITY#dfn-visibility-states">visibility
+      <dfn data-cite="html#visibility-state">visibility
       state</dfn> of the [=environment settings object / responsible document=]
       of the <a>current settings object</a> changes, it must run these steps:
     </p>
@@ -4844,7 +4844,7 @@
       <p>
         Web NFC functionality is allowed only for the {{Document}} of the
         <a>top-level browsing context</a>, which must be
-        <dfn data-cite="PAGE-VISIBILITY#dom-visibilitystate-visible">visible</dfn>.
+        <dfn data-cite="html#visibility-state">visible</dfn>.
       </p>
       <p>
         This also means that UAs should block access to the NFC radio if
--- a/index.html
+++ b/index.html
@@ -885,7 +885,7 @@
     <p>
       Reading an <a>NFC tag</a> containing an <a>NDEF message</a>, while the
       {{Document}} of the <a>top-level browsing context</a> using Web NFC is
-      <a>visible</a>. For instance, a web page instructs the user to tap an NFC
+      visible. For instance, a web page instructs the user to tap an NFC
       tag, and then receives information from the tag.
     </p>
   </section>
@@ -2166,9 +2166,8 @@
 
   <section><h3>Handling visibility change</h3>
     <p>
-      When the user agent determines that the
-      <dfn data-cite="html#visibility-state">visibility
-      state</dfn> of the [=environment settings object / responsible document=]
+      When the user agent determines that the {{Document/visibilityState}}
+      of the [=environment settings object / responsible document=]
       of the <a>current settings object</a> changes, it must run these steps:
     </p>
     <ol class="algorithm">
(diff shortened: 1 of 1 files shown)


noamr:
> This PR is an attempt to address #626 as the Page Visibility spec is moving to the HTML spec.
> 
> @noamr What do you think?
> 
> [Preview](https://pr-preview.s3.amazonaws.com/beaufortfrancois/web-nfc/pull/627.html) | [Diff](https://pr-preview.s3.amazonaws.com/w3c/web-nfc/627/760581a...beaufortfrancois:8536386.html)

Can't you use `{{Document/visibilityState}}` instead of `data-cite`?

In addition, the heads up was about this: https://w3c.github.io/web-nfc/#handling-visibility-change.
It "monkey patches" the HTML spec by saying that "X and Y happens when the visibility changes".
The proposal was to make a call from the HTML spec to this spec [here](https://html.spec.whatwg.org/#page-visibility-change-steps), which would allow specifying the order between reactions to visibility change.

beaufortfrancois:
> Can't you use `{{Document/visibilityState}}` instead of `data-cite`?

Done. Thanks!
 
> In addition, the heads up was about this: https://w3c.github.io/web-nfc/#handling-visibility-change. It "monkey patches" the HTML spec by saying that "X and Y happens when the visibility changes". The proposal was to make a call from the HTML spec to this spec [here](https://html.spec.whatwg.org/#page-visibility-change-steps), which would allow specifying the order between reactions to visibility change.

Since Web NFC is implemented only in Chromium, I don't think it makes sense to update the HTML spec for now. What do you think?

noamr:
> > Can't you use `{{Document/visibilityState}}` instead of `data-cite`?
> 
> 
> 
> Done. Thanks!
> 
>  
> 
> > In addition, the heads up was about this: https://w3c.github.io/web-nfc/#handling-visibility-change. It "monkey patches" the HTML spec by saying that "X and Y happens when the visibility changes". The proposal was to make a call from the HTML spec to this spec [here](https://html.spec.whatwg.org/#page-visibility-change-steps), which would allow specifying the order between reactions to visibility change.
> 
> 
> 
> Since Web NFC is implemented only in Chromium, I don't think it makes sense to update the HTML spec for now. What do you think?
> 
> 
> 
> 

It's a good question. Usually HTML spec changes require two interested implementors.

/cc @yoavweiss

noamr:
In any case, if we stay with monkey-patching, I would suggest to refer to [this](https://html.spec.whatwg.org/#update-the-visibility-state) rather than to the old `page visibility` hook.

beaufortfrancois:
> In any case, if we stay with monkey-patching, I would suggest to refer to [this](https://html.spec.whatwg.org/#update-the-visibility-state) rather than to the old `page visibility` hook.

How would you update this spec then?

noamr:
> > In any case, if we stay with monkey-patching, I would suggest to refer to [this](https://html.spec.whatwg.org/#update-the-visibility-state) rather than to the old `page visibility` hook.
> 
> How would you update this spec then?

I would add a link inside `web-nfc` referring to that place in HTML [here](https://w3c.github.io/web-nfc/#handling-visibility-change).

beaufortfrancois:
Can you suggest some wording? I'm not sure how to do this.

Here's what we have for now:

> When the user agent determines that the visibilityState of the responsible document of the current settings object changes, it must run these steps:

noamr:
Perhaps something along the lines of: 
when the user-agent [updates the visibility](https://html.spec.whatwg.org/#update-the-visibility-state) of {{Document}} doc, if doc is the active document, run these steps:

beaufortfrancois:
https://respec.org/xref/?term=update+the+visibility+state is not exported sadly. Is that intended?

noamr:
> https://respec.org/xref/?term=update+the+visibility+state is not exported sadly. Is that intended?

It is, because the intention was to invert the reference (direct calls rather than hooks). But I want to make sure that's OK for specs that have not reached consensus like this one.

beaufortfrancois:
Ok. I'm not familiar enough with respec to know the best way to proceed. Shall I simply link to  https://html.spec.whatwg.org/#update-the-visibility-state with an anchor or is there another way?

noamr:
> Ok. I'm not familiar enough with respec to know the best way to proceed. Shall I simply link to https://html.spec.whatwg.org/#update-the-visibility-state with an anchor or is there another way?

AFAIK that's the way to do it if it's not exported.

beaufortfrancois:
Thanks. I've updated this PR to link to https://html.spec.whatwg.org/#update-the-visibility-state

noamr (approved):
LGTM FWIW :)

beaufortfrancois:
Thanks @noamr!

@kenchris or @zolkis Can you review as well?

kenchris (approved):

PR #629, opened 2021-11-18
Add type NDEFRecordType to web-nfc.d.ts

drewmw5:
This pull request is for adding types to recordTypes.
- It adds a new type: "NDEFRecordTypes"
- It includes all valid record types for recordTypes referenced from https://developer.mozilla.org/en-US/docs/Web/API/NDEFReader/write
- Values consist of: 
   "absolute-url"
   "empty"
   "mime"
   "smart-poster"
   "text"
   "unknown"
   "URL"

### DIFF
--- a/web-nfc.d.ts
+++ b/web-nfc.d.ts
@@ -22,9 +22,11 @@ declare type NDEFRecordDataSource = string | BufferSource | NDEFMessageInit
 interface Window {
   NDEFRecord: NDEFRecord
 }
+type NDEFRecordTypes = "absolute-url" | "empty" | "mime" | "smart-poster" | "text" | "unknown" | "URL";
+
 declare class NDEFRecord {
   constructor(recordInit: NDEFRecordInit)
-  readonly recordType: string
+  readonly recordType: NDEFRecordTypes
   readonly mediaType?: string
   readonly id?: string
   readonly data?: DataView
@@ -33,7 +35,7 @@ declare class NDEFRecord {
   toRecords?: () => NDEFRecord[]
 }
 declare interface NDEFRecordInit {
-  recordType: string
+  recordType: NDEFRecordTypes
   mediaType?: string
   id?: string
   encoding?: string
--- a/web-nfc.d.ts
+++ b/web-nfc.d.ts
@@ -22,7 +22,7 @@ declare type NDEFRecordDataSource = string | BufferSource | NDEFMessageInit
 interface Window {
   NDEFRecord: NDEFRecord
 }
-type NDEFRecordTypes = string | "absolute-url" | "empty" | "mime" | "smart-poster" | "text" | "unknown" | "URL";
+type NDEFRecordTypes = string | "absolute-url" | "empty" | "mime" | "smart-poster" | "text" | "unknown" | "url";
 
 declare class NDEFRecord {
   constructor(recordInit: NDEFRecordInit)
--- a/web-nfc.d.ts
+++ b/web-nfc.d.ts
@@ -22,7 +22,7 @@ declare type NDEFRecordDataSource = string | BufferSource | NDEFMessageInit
 interface Window {
   NDEFRecord: NDEFRecord
 }
-type NDEFRecordTypes = "absolute-url" | "empty" | "mime" | "smart-poster" | "text" | "unknown" | "URL";
+type NDEFRecordTypes = string | "absolute-url" | "empty" | "mime" | "smart-poster" | "text" | "unknown" | "URL";
 
 declare class NDEFRecord {
   constructor(recordInit: NDEFRecordInit)
--- a/web-nfc.d.ts
+++ b/web-nfc.d.ts
@@ -22,7 +22,7 @@ declare type NDEFRecordDataSource = string | BufferSource | NDEFMessageInit
 interface Window {
   NDEFRecord: NDEFRecord
 }
-type NDEFRecordTypes = string | "absolute-url" | "empty" | "mime" | "smart-poster" | "text" | "unknown" | "url";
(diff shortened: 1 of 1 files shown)


beaufortfrancois:
As you can see at https://w3c.github.io/web-nfc/#data-mapping, `recordType` can be any string that follows a certain pattern (See local type and external type).

Is it possible in TypeScript to have something like below?
```ts
  readonly recordType: NDEFRecordTypes | string;
```
or 
```ts
type NDEFRecordTypes = string | "absolute-url" | "empty" | "mime" | "smart-poster" | "text" | "unknown" | "URL";
```

drewmw5:
> As you can see at https://w3c.github.io/web-nfc/#data-mapping, `recordType` can be any string that follows a certain pattern (See local type and external type).
> 
> Is it possible in TypeScript to have something like below?
> 
> ```ts
>   readonly recordType: NDEFRecordTypes | string;
> ```
> 
> or
> 
> ```ts
> type NDEFRecordTypes = string | "absolute-url" | "empty" | "mime" | "smart-poster" | "text" | "unknown" | "URL";
> ```

Ah, good to know, there's still plenty I have left to read. Thanks for that. 
You are correct, the following example you gave is valid TypeScript, the other one was not:

```ts
type NDEFRecordTypes = string | "absolute-url" | "empty" | "mime" | "smart-poster" | "text" | "unknown" | "URL";
```
Is there more to it than that?

drewmw5:
> > As you can see at https://w3c.github.io/web-nfc/#data-mapping, `recordType` can be any string that follows a certain pattern (See local type and external type).
> > Is it possible in TypeScript to have something like below?
> > ```ts
> >   readonly recordType: NDEFRecordTypes | string;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > or
> > ```ts
> > type NDEFRecordTypes = string | "absolute-url" | "empty" | "mime" | "smart-poster" | "text" | "unknown" | "URL";
> > ```
> 
> Ah, good to know, there's still plenty I have left to read. Thanks for that. You are correct, the following example you gave is valid TypeScript, the other one was not:
> 
> ```ts
> type NDEFRecordTypes = string | "absolute-url" | "empty" | "mime" | "smart-poster" | "text" | "unknown" | "URL";
> ```
> 
> I've gone ahead and updated the original commit. Is there more to it than that?

web-nfc.d.ts
```diff
@@ -22,9 +22,11 @@ declare type NDEFRecordDataSource = string | BufferSource | NDEFMessageInit
 interface Window {
   NDEFRecord: NDEFRecord
 }
+type NDEFRecordTypes = string | "absolute-url" | "empty" | "mime" | "smart-poster" | "text" | "unknown" | "URL";
```
beaufortfrancois: Should we have `declare` as well there?

beaufortfrancois: I believe it should be "url", not "URL".

beaufortfrancois:
@takefumi-yoshii Can you have a look as well?

web-nfc.d.ts
takefumi-yoshii: It is incorrect to include `string` here. In this definition, it is synonymous with `string`.
If exclude `string`, we will get the expected compilation error. [What this means](https://www.typescriptlang.org/play?#code/C4TwDgpgBA+gcgEQKIDEBKEDGB7ATgEwBVwIBnKAXilOFwEsA7AcygB8oAiAQwCNTsANgFdgEALRDcAjm04QAtmFAz2HeXXkQVnUvK65gYsNhoRc2jqIAewC0IYBrBtgDuDCwFU0AGQ4BuAFgAKGCAelCoQgALM2g6cmcoM1w8YJwGGmouTQBBUgBlWkYmAC5YRFQMHAJiSHIqAHJ1Gi4HCAbg4NBIKAr0LDwiEnrOXn5hUQkpCwUlEAt1TQtdfUNjU3NZSwgbO0dnN08ff06g8MiY3DjyaqvMYCTcFNw07AyHnch7iHwkJ7wyn0qoNamRKFAmvFgK12sEgA)

Union types of such `string literal type` should be used to avoid including incorrect strings.

takefumi-yoshii: plus one 👍

zolkis: Yes, if we want a language level (early) error. Just noting that it's also possible to defer that decision (to accept or not a given string record type) to implementations: then a general string should be fine, including `"mistake"`. It depends how do we want to do it.
IMHO we (Web NFC) are not in the position to extend the range of valid NFC record types or leave extension possibilities to implementations, therefore I agree on using string literals.

takefumi-yoshii: It is correct that "NDEF Record Type" can specify any character string (local type and external type), So it should be a `string`.Committing as a document is good, but it doesn't contribute to Type Safe.

drewmw5: Good catch. Fixed

drewmw5: I just went ahead and updated it to include the declare keyword and also removed the string type.

zolkis:
As @beaufortfrancois mentioned [earlier](https://github.com/w3c/web-nfc/pull/629#issuecomment-972637525), record type can be a string. How would you write this example after this PR?
https://w3c.github.io/web-nfc/#write-an-external-record-with-an-ndef-message-as-payload 

Then, smart poster local types should also be listed?

zolkis:
Well, to me it seems that if we cannot add to the definitions a generic "string" as a fallback, then we should leave `recordType` as `string`. Static type validation as "string" should be enough, and dynamic validation, including standard record types defined by the NFC Forum, is defined by the algorithms in the spec, and it will depend on context (where the records belong). 
If there are shortcomings there, please file issues.

Issue #558, opened 2020-04-16
Make tag read only


jicihome:
Hi!
I was wondering how can we "lock" the NFC tag after writting or make it read only? I see that this some security specification, but didn't found a way to do it using Write or another method.

Thanks!

pechisworks:
This would also be useful for my use cases.

nico-martin:
I'm currently playing around with this and I think this is a crucial feature. Let's imagine I have NFC Tags in a Museum and anyone would be able to overwrite the content. That would be quite bad.

But looking through the Android specification I've seen NFC as a standard already has such a feature. So if i write an NFC Tag with readOnly from an Android app, I guess WebNFC could not just overwrite that content, right? If thats the case, this would be a workaround for now..

beaufortfrancois:
@kenchris What are your thoughts?

Thanks for the feedback @nico-martin !
I'd love to see more signals from developers requesting this feature before adding it to Web NFC.


```json5
[SecureContext, Exposed=Window]
interface NDEFReader : EventTarget {
  ...
  Promise<undefined> makeReadOnly()
};
```



Note that this addition would be pretty simple indeed on [Android]( https://developer.android.com/reference/android/nfc/tech/Ndef#makeReadOnly()), but not necessarily on [iOS currently](https://github.com/chariotsolutions/phonegap-nfc/issues/409).

kenchris:
I am totally fine with this! It can return a promise and just reject if not supported

nico-martin:
Maybe a general NFC-question: How can I revoke the readOnly flag? Is this possible with NFC Tags?

beaufortfrancois:
This NFC operation is a one-way process and cannot be reverted for NFC tags.

nico-martin:
Ok. Maybe this could lead to some confusion for people that are not familiar with this.
Would it make sense to explicitly require a flag from the user that indicates this one-way usage?
This would be similar to the Web Blutooth API, where `requestDevice` explicitly requires an `acceptAllDevices: true` if no filters are set.

```javascript
const ndef = new NDEFReader();
try {
  await ndef.write({
    records: [{ recordType: "url", data: "https://w3c.github.io/web-nfc/" }]
  });
  await ndef.makeReadOnly({
    allowIrreversibleOperation: true
  });
} catch (e) {
  console.log("Something went wrong.", e);
};
```

[...]

PR #632, opened 2021-11-29
Add experimental NDEFReader.makeReadOnly() method

beaufortfrancois:
Looking at https://github.com/w3c/web-nfc/issues/558 and signals on Twitter, it seems like web developers would love to "lock" NFC tags with Web NFC. This PR is an attempt to address their feedback by adding a new method `makeReadOnly()` to `NDEFReader`. 

```webidl
partial interface NDEFReader {
  Promise<undefined> makeReadOnly(optional NDEFMakeReadOnlyOptions options={});
};

dictionary NDEFMakeReadOnlyOptions {
  AbortSignal? signal;
};
```

Examples:

```js
ndef.makeReadOnly().then(() => {
  console.log("NFC tag has been made permanently read-only.");
}).catch(error => {
  console.log(`Operation failed: ${error}`);
});
```

```js
try {
  await ndef.write("Hello world");
  console.log("Message written.");
  await ndef.makeReadOnly();
  console.log("NFC tag has been made permanently read-only.");
} catch (error) {
  console.log(`Operation failed: ${error}`);
}
```

***
<a href="https://pr-preview.s3.amazonaws.com/beaufortfrancois/web-nfc/pull/632.html" title="Last updated on Dec 7, 2021, 8:34 AM UTC (620b0c7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/632/bfd3769...beaufortfrancois:620b0c7.html" title="Last updated on Dec 7, 2021, 8:34 AM UTC (620b0c7)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -917,25 +917,25 @@
     </p>
   </section>
 
-  <section> <h3>Making an <a>NFC tag</a> read only</h3>
+  <section> <h3>Making an <a>NFC tag</a> read-only</h3>
     <div>
-      The user opens a web page which can make an <a>NFC tag</a> read only.
+      The user opens a web page which can make an <a>NFC tag</a> read-only.
       The operations may be one of the following:
       <ol>
         <li>
-          Making a non-formatted <a>NFC tag</a> read only.
+          Making a non-formatted <a>NFC tag</a> read-only.
         </li>
         <li>
-          Making an empty, but formatted <a>NFC tag</a> read only.
+          Making an empty, but formatted <a>NFC tag</a> read-only.
         </li>
         <li>
           Making an <a>NFC tag</a> which already contains an
-          <a>NDEF message</a> read only.
+          <a>NDEF message</a> read-only.
         </li>
       </ol>
     </div>
     <p class="note">
-      Note that making an <a>NFC tag</a> read only always involves
+      Note that making an <a>NFC tag</a> read-only always involves
       a read operation.
     </p>
   </section>
@@ -999,7 +999,7 @@
     </p>
     <p>
       An NFC reader works by polling, so in order to be able to
-      write to a tag or making it read only, a tag first needs to be found and
+      write to a tag or making it read-only, a tag first needs to be found and
       read, which means that polling needs to be initialized first.
     </p>
     <p>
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -917,6 +917,29 @@ <h4>
...
+        </li>
+        <li>
+          Making an <a>NFC tag</a> which already contains an
+          <a>NDEF message</a> read only.
+        </li>
+      </ol>
+    </div>
+    <p class="note">
+      Note that making an <a>NFC tag</a> read only always involves
+      also a read operation.
```
kenchris: remove also

index.html
```diff
@@ -3386,6 +3484,150 @@ <h3><dfn>Writing content</dfn></h3>
     </section>
   </section>
 
+  <section id="making-content-read-only">
+    <h3><dfn>Making content read only</dfn></h3>
+    <p>
+      This section describes how to make an <a>NFC tag</a> read only when it is
+      in proximity range. At any time there is a maximum of one request for an
+      <a>origin</a> until the <a>NFC tag</a> is made read only or the operation
```
kenchris: I am wondering whether we should say read-only instead with a hyphen

kenchris (approved):

index.html
beaufortfrancois: Fixed.

index.html
beaufortfrancois: Fixed.

index.html
```diff
@@ -2261,6 +2348,19 @@ <h3>The <dfn>NDEFWriteOptions</dfn> dictionary</h3>
     </p>
   </section>
 
+  <section data-dfn-for="NDEFMakeReadOnlyOptions">
+    <h3>The <dfn>NDEFMakeReadOnlyOptions</dfn> dictionary</h3>
+    <pre class="idl">
+      dictionary NDEFMakeReadOnlyOptions {
+        AbortSignal? signal;
+      };
```
zolkis, line 2356: Just a note to self. Later we might want to consider adding a safety-option here, e.g. a serial number, making sure only the given tag is made read-only. It should be taken care by the recommended practice of doing writes and making read-only from a read operation.

zolkis (approved):
Looks good. Great work! Thank you!

zolkis (approved):

kenchris (approved):

kenchris (approved):

reillyeon (changes requested):
I'd like to see an update to the security considerations section which expands on the "tag modification" threat to include permanent destruction of a tag by modifying its contents and then making it read-only.

I'm not convinced that this attack is significantly more destructive than modifying the tag while leaving it writable but it should be mentioned.

beaufortfrancois:
@reillyeon Thank you for the review. I've mentioned "read-only" in the NFC tag modification section as you've suggested.

reillyeon (approved):

PR #636, opened 2022-01-17
add makeReadOnly() to EXPLAINER

zolkis:

### DIFF
--- a/EXPLAINER.md
+++ b/EXPLAINER.md
@@ -53,6 +53,10 @@ post-it notes, anyone can read them, and unless read-only, anyone can write
 them. Web NFC does not handle automatic encryption or signing of NFC content
 (with NDEF signatures): handling PKI infrastructure is left to applications.
 
+The API does not support controlling who can change the content on an NDEF
+tag. However, the API supports making an NDEF tag permanently read-only, see the
+[examples](https://w3c.github.io/web-nfc/#make-an-nfc-tag-permanently-read-only).
+
 ## API shape and examples
 
 The API has gone through multiple iterations and have had feedback from Mozilla,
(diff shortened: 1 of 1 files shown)


EXPLAINER.md
```diff
@@ -53,6 +53,10 @@ post-it notes, anyone can read them, and unless read-only, anyone can write
 them. Web NFC does not handle automatic encryption or signing of NFC content
 (with NDEF signatures): handling PKI infrastructure is left to applications.
 
+The API does not support controlling who can change the content on an NDEF
+tag. However, the API supports making an NDEF tag permanently read-only, see the
+[examples](https://w3c.github.io/web-nfc/#make-an-nfc-tag-permanently-read-only).
+
 ## API shape and examples
```
beaufortfrancois, line 82: Shall we add a simple JS example in this section?

```js
try {
  await ndef.write("Hello world");
  console.log("Message written.");
  await ndef.makeReadOnly();
  console.log("NFC tag has been made permanently read-only after writing to it.");
} catch (error) {
  console.log(`Operation failed: ${error}`);
}
```

EXPLAINER.md
```diff
@@ -284,6 +290,13 @@ Since the merge of `NDEFWriter` and `NDEFReader`, the former `onerror` event on
 relevant on writes, which is not the case. This change clarifies the error is
 only relevant for scans.
 
+### Added the `makeReadOnly()` method
+
+Justified in [issue 558](https://github.com/w3c/web-nfc/issues/558) and by feedback
+received on Twitter. It seems like web developers would like to be able to
```
beaufortfrancois: Can we add one or two examples use cases there as well?

EXPLAINER.md
zolkis: I thought that referring to the examples in the spec was fine, but we can have it inline, too, indeed.

zolkis: Actually in other sections we also only link the examples.
Inline examples are there only for the integrations.
I think it makes sense this way (linking examples).

kenchris: I think you should be able to read an explainer without having to open or link into other documents

EXPLAINER.md
kenchris: I don't like wording like "seems like" - so we don't really know? It's very weak

EXPLAINER.md
zolkis: Should we inline the other examples as well, then?

kenchris: I would say yes

zolkis:
The previous commit amended, with example and use case for readonly. Feel free to further change it.

beaufortfrancois (approved):

PR #641, opened 2022-03-28
Proper document access in page visibility change steps

johannhof:
As I learned from @Domenic, "relevant global object" cannot be used on
"current settings object", only on platform objects
(https://html.spec.whatwg.org/multipage/webappapis.html#concept-relevant-global)

Luckily, the HTML spec already passes visibilityState and document to
other specifications running page visibility change steps. So we can
just use those directly here.

***
<a href="https://pr-preview.s3.amazonaws.com/johannhof/web-nfc/pull/641.html" title="Last updated on Mar 29, 2022, 11:47 AM UTC (a06a7bd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/641/75fc0bf...johannhof:a06a7bd.html" title="Last updated on Mar 29, 2022, 11:47 AM UTC (a06a7bd)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2214,7 +2214,7 @@
 
   <section><h3>Handling visibility change</h3>
     <p>
-      The <a href="https://html.spec.whatwg.org/#page-visibility-change-steps">page visibility change steps</a> for this specification, given |visibilityState| and |document|, are:
+    The [=page visibility change steps=] for this specification, given the string |visibilityState| and the {{Document}} |document|, are:
     </p>
     <ol class="algorithm">
       <li>
--- a/index.html
+++ b/index.html
@@ -2214,17 +2214,11 @@
 
   <section><h3>Handling visibility change</h3>
     <p>
-      When the user agent <a href="https://html.spec.whatwg.org/#update-the-visibility-state">updates the visibility state</a>
-      of the <a>current settings object</a>'s <a>relevant global object</a>'s
-      <a>associated <code>Document</code></a> it must run these steps:
+      The <a href="https://html.spec.whatwg.org/#page-visibility-change-steps">page visibility change steps</a> for this specification, given |visibilityState| and |document|, are:
     </p>
     <ol class="algorithm">
       <li>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2214,17 +2214,11 @@ <h2>The NDEFReader object</h2>
 
   <section><h3>Handling visibility change</h3>
     <p>
-      When the user agent <a href="https://html.spec.whatwg.org/#update-the-visibility-state">updates the visibility state</a>
-      of the <a>current settings object</a>'s <a>relevant global object</a>'s
-      <a>associated <code>Document</code></a> it must run these steps:
+      The <a href="https://html.spec.whatwg.org/#page-visibility-change-steps">page visibility change steps</a> for this specification, given |visibilityState| and |document|, are:
```
johannhof: It might be good practice to link the `visibilityState` and `document` variables to `Document/visibilityState` and `Document` respectively?

johannhof:
@domenic would mind taking a quick look at this as well? :)

index.html
domenic: I don't think so. You could mention their types, e.g. `given a string |visibilityState| and a {{Document}} |document|`? But the variables themselves should not be links.

domenic: This looks like a ReSpec spec, so I'm not sure it will work, but try `[=page visibility change steps=]` instead of a manual `<a>`?

domenic (approved):
Nice!

index.html
kenchris: Yeah or use https://respec.org/xref/ to find the right way to reference it

beaufortfrancois:
I'm happy to rubberstamp merge when @domenic approves.

index.html
johannhof: Very helpful, thanks!

johannhof:
Updated!

domenic (approved):

beaufortfrancois:
Thank you @johannhof for this work! Much appreciated.

PR #643, opened 2022-06-28
Fixed to reading/writing of external records

### LINKED ISSUE #642: Is the external type validate part correct?
image

asciiDomain here is the NSS part of https://datatracker.ietf.org/doc/html/rfc2141.html#section-2.2

For instance there is the rule that if there is a % it must be followed by two hex numbers, that is not checked in the algorithm.

Apart from that  "/" | "?" | "#" are allowed but reserved so we should probably ignore those for now (which we are doing).

kenchris:
Closes #642

***
<a href="https://pr-preview.s3.amazonaws.com/kenchris/web-nfc/pull/643.html" title="Last updated on Jun 29, 2022, 10:58 AM UTC (187e2d5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/web-nfc/643/78f8fb0...kenchris:187e2d5.html" title="Last updated on Jun 29, 2022, 10:58 AM UTC (187e2d5)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3241,6 +3241,13 @@
         To <dfn>map external data to NDEF</dfn> given |record:NDEFRecordInit|,
         |ndef|, and |recordsDepth:unsigned short|, run these steps:
         <ol class=algorithm data-link-for="NDEFRecordInit">
+          <!--
+          <li>
+            If |context| is `"smart-poster"`,
+            If |record|'s <a>mediaType</a> is not `undefined`,
+            [= exception/throw =] a {{TypeError}} and abort these steps.
+          </li>
+          -->
           <li>
             If |record|'s <a>mediaType</a> is not `undefined`,
             [= exception/throw =] a {{TypeError}} and abort these steps.
--- a/index.html
+++ b/index.html
@@ -3244,7 +3244,6 @@
           <!--
           <li>
             If |context| is `"smart-poster"`,
-            If |record|'s <a>mediaType</a> is not `undefined`,
             [= exception/throw =] a {{TypeError}} and abort these steps.
           </li>
           -->
--- a/index.html
+++ b/index.html
@@ -2833,14 +2833,15 @@
       </p>
       <ol class=algorithm id="split-external-type">
         <li>
+          If `U+003A` (`:`) is not found in |input|, return failure.
+        </li>
+        <li>
           Let |domain| be the |input| from the start of |input| up to
-          but excluding the first occurrence of `U+003A` (`:`), or return failure
-          if that is not found.
+          but excluding the first occurrence of `U+003A` (`:`).
         </li>
         <li>
           Let |type| be the |input| after the first occurrence, if any, of
-          `U+003A` (`:`) up to the end of |input|, or return failure if that is
-          not found.
+          `U+003A` (`:`) up to the end of |input|.
         </li>
         <li>
           Return the |domain| and |type| pair.
@@ -2850,13 +2851,13 @@
         To <dfn>validate external type</dfn> given |input:USVString|, run
         these steps:
       </p>
-      <ol class=algorithm id="split-external-type">
+      <ol class=algorithm id="validate-external-type">
         <li>
           Let |domain| and |type| be the result of running [=split external type=], or
           return `false` if failure.
         </li>
         <li>
-          If |domain| is not a [=valid domain string=], return failure.
+          If |domain| is not a [=valid domain string=], return `false`.
         </li>
         <li>
           If |type| contains [=code points=] that are not
@@ -2864,7 +2865,7 @@
           `U+0028` `LEFT PARENTHESIS` (`(`), `U+0029` `RIGHT PARENTHESIS` (`)`),
           `U+002A` (`*`), `U+002B` (`+`), `U+002C` (`,`), `U+002D` (`-`),
           `U+002E` (`.`), `U+003B` (`;`), `U+003D` (`=`), `U+0040` (`@`),
-          `U+005F` (`_`), return failure.
+          `U+005F` (`_`), return `false`.
         </li>
         <li>
           Return `true`.
@@ -3242,7 +3243,7 @@
         <ol class=algorithm data-link-for="NDEFRecordInit">
           <li>
             If |record|'s <a>mediaType</a> is not `undefined`,
-            [= exception/throw =] a {{TypeError}}.
+            [= exception/throw =] a {{TypeError}} and abort these steps.
           </li>
           <li>
             Let |domain| and |type| be the result of running [=split external type=] with
@@ -3253,12 +3254,12 @@
             domain to ASCII</a> with |domain| and true.
           </li>
           <aside class="note">
-            The recordType is user set and in UTF-8 encoding, allowing for international domain names as part of the
-            custom type such as "håndværker.dk:abc". When writing that to the NDEF it has to be converted to ASCII
+            The record's <a>recordType</a> is is user set and in UTF-8 encoding, allowing for international domain names as part of the
+            custom type such as "håndværker.dk:abc". When writing that value to an NDEF record it has to be converted to ASCII
             bytes and after the conversion this would be represented as "xn--hndvrker-9zan.dk:abc".
           </aside>
           <li>
-            Let |customTypeName| be |domain|, ":" and |type| concatenated.
+            Let |customTypeName| be |domain|, "`:`" and |type| concatenated.
           </li>
           <li>
             If |customTypeName| is not a {{USVString}} or its length exceeds 255 bytes, [= exception/throw =] a {{TypeError}}.
@@ -3269,7 +3270,7 @@
           <li>
             If the type of a |record|'s <a>data</a> is not
             {{BufferSource}} or {{NDEFMessageInit}}, [= exception/throw =] a
-            {{TypeError}}.
+            {{TypeError}} and abort these steps.
           </li>
           <li>
             Set |ndef|'s <a>TNF field</a> to `4`
@@ -4455,6 +4456,9 @@
         <li>
           Let |domain| and |type| be the result of running [=split external type=] with the value of |ndefRecord|'s <a>TYPE field</a>.
         </li>
+        <aside class="note">
+          The above step is ensures to work due to earlier validation of the external type.
+        </aside>
         <li>
           Let |domain| be the result of running <a href="https://www.unicode.org/reports/tr46/#ToUnicode">Unicode ToUnicode</a>
           with |domain_name| set to |domain|, |CheckHyphens| set to false, |CheckBidi| set to true, |CheckJoiners| set to true,
@@ -4465,7 +4469,7 @@
           be presented as "håndværker.dk:abc" after the above conversion.
         </aside>
         <li>
-          Set |record|'s <a>recordType</a> to |domain|, ":" and |type| concatenated.
+          Set |record|'s <a>recordType</a> to |domain|, "`:`" and |type| concatenated.
         </li>
         <li>
           Set |record|'s <a>mediaType</a> to `null`.
--- a/index.html
+++ b/index.html
@@ -2808,7 +2808,7 @@
         The [[NFC-RTD]] specifies that external types MUST contain
         the [=domain=] name of the issuing organization, a colon `U+003A` (`:`)
         and a type name that is at least one character long, for instance
-        "`w3.org:member`".
+        "`w3.org:member`", all stored as ASCII characters.
       </div>
       <p>
         The [[NFC-RTD]] specifies the URN prefix “`urn:nfc:ext:`” as well, but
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2828,46 +2828,43 @@ <h3><dfn>Writing content</dfn></h3>
...
-          <a href="https://url.spec.whatwg.org/#concept-domain-to-ascii">
-          domain to ASCII</a> given |domain| and `true` (as |beStrict|).
+          Return the |domain| and |type| pair.
         </li>
+      </ol>
+      <p>
+        To <dfn>validate external type</dfn> given |input:USVString|, run
+        these steps:
+      </p>
+      <ol class=algorithm id="split-external-type">
```
beaufortfrancois: I believe `id` should be `"validate-external-type"`

index.html
```diff
@@ -2828,46 +2828,43 @@ <h3><dfn>Writing content</dfn></h3>
...
-        To <dfn>validate external type</dfn> given |input:USVString|, run
+        To <dfn>split external type</dfn> given |input:USVString|, run
         these steps:
       </p>
-      <ol class=algorithm id="validate-external-type">
-        <li>
-          If |input:string| is not an [=ASCII string=], is empty, or its
-          length exceeds 255 bytes, return `false`.
-        </li>
+      <ol class=algorithm id="split-external-type">
```
beaufortfrancois, line 2834: Shall we fail early when `:` is not found with something like this?
> 1.  If `U+003A` (`:`) is not found in |input|, return failure.
> 2. Let |domain| be the |input| from the start of |input| up to but excluding the first occurrence of `U+003A` (`:`).
> 3. Let |type| be the |input| after the first occurrence of `U+003A` (`:`) up to the end of |input|.
> 4. Return the |domain| and |type| pair.

index.html
```diff
@@ -2828,46 +2828,43 @@ <h3><dfn>Writing content</dfn></h3>
...
         <li>
-          If |asciiDomain| is failure, return `false`.
+          Let |domain| and |type| be the result of running [=split external type=], or
+          return `false` if failure.
         </li>
         <li>
-          If |asciiDomain| contains a [=forbidden host code point=] or
-          `U+005F` `LOW LINE` (`_`), return `false`.
+          If |domain| is not a [=valid domain string=], return failure.
```
beaufortfrancois: Why not returning `false` as before in the case of failure?

index.html
```diff
@@ -2828,46 +2828,43 @@ <h3><dfn>Writing content</dfn></h3>
...
         </li>
         <li>
           If |type| contains [=code points=] that are not
           [=ASCII alphanumeric=], or `U+0024` (`$`), `U+0027` (`'`),
           `U+0028` `LEFT PARENTHESIS` (`(`), `U+0029` `RIGHT PARENTHESIS` (`)`),
           `U+002A` (`*`), `U+002B` (`+`), `U+002C` (`,`), `U+002D` (`-`),
           `U+002E` (`.`), `U+003B` (`;`), `U+003D` (`=`), `U+0040` (`@`),
-          `U+005F` (`_`), return `false`.
+          `U+005F` (`_`), return failure.
```
beaufortfrancois: Why not returning `false` as before in the case of failure?

index.html
```diff
@@ -3243,29 +3240,41 @@ <h3><dfn>Writing content</dfn></h3>
...
+          <li>
+            If |customTypeName| is not a {{USVString}} or its length exceeds 255 bytes, [= exception/throw =] a {{TypeError}}.
+          </li>
+          <li>
+            Set the |ndef|'s <a>TYPE field</a> to |customTypeName|.
           </li>
           <li>
             If the type of a |record|'s <a>data</a> is not
             {{BufferSource}} or {{NDEFMessageInit}}, [= exception/throw =] a
-            {{TypeError}} and abort these steps.
```
beaufortfrancois: I've noticed we're not consistent the spec when it comes to "throw a TypeError" and "throw a TypeError and abort these steps". Please file a separate issue to address those and not pollute this PR.

index.html
```diff
@@ -3243,29 +3240,41 @@ <h3><dfn>Writing content</dfn></h3>
         To <dfn>map external data to NDEF</dfn> given |record:NDEFRecordInit|,
         |ndef|, and |recordsDepth:unsigned short|, run these steps:
         <ol class=algorithm data-link-for="NDEFRecordInit">
-          <!--
```
beaufortfrancois: Why did you remove this comment? Is this intentional?

index.html
```diff
@@ -3243,29 +3240,41 @@ <h3><dfn>Writing content</dfn></h3>
...
           <li>
-            If |context| is `"smart-poster"`,
-            [= exception/throw =] a {{TypeError}} and abort these steps.
+            If |record|'s <a>mediaType</a> is not `undefined`,
+            [= exception/throw =] a {{TypeError}}.
```
beaufortfrancois: Same here. Let's not remove the "and abort these steps" in this PR.

index.html
```diff
@@ -3243,29 +3240,41 @@ <h3><dfn>Writing content</dfn></h3>
...
-            [= exception/throw =] a {{TypeError}} and abort these steps.
+            Let |domain| and |type| be the result of running [=split external type=] with
+            |record|'s <a>recordType</a>.
+          </li>
+          <li>
+            Let |domain| be the result of running <a href="https://url.spec.whatwg.org/#concept-domain-to-ascii">
+            domain to ASCII</a> with |domain| and true.
+          </li>
+          <aside class="note">
+            The recordType is user set and in UTF-8 encoding, allowing for international domain names as part of the
```
beaufortfrancois: Nit: `"The record's <a>recordType</a> is..."`

index.html
```diff
@@ -3243,29 +3240,41 @@ <h3><dfn>Writing content</dfn></h3>
...
+            custom type such as "håndværker.dk:abc". When writing that to the NDEF it has to be converted to ASCII
```
beaufortfrancois: Nit: `"When writing that value to an NDEF record"`

index.html
```diff
@@ -3243,29 +3240,41 @@ <h3><dfn>Writing content</dfn></h3>
...
+            bytes and after the conversion this would be represented as "xn--hndvrker-9zan.dk:abc".
+          </aside>
+          <li>
+            Let |customTypeName| be |domain|, ":" and |type| concatenated.
```
beaufortfrancois: Nit: Use backticks for colon.

index.html
```diff
@@ -4444,8 +4453,19 @@ <h3>Parsing content</h3>
...
+          Let |domain| be the result of running <a href="https://www.unicode.org/reports/tr46/#ToUnicode">Unicode ToUnicode</a>
+          with |domain_name| set to |domain|, |CheckHyphens| set to false, |CheckBidi| set to true, |CheckJoiners| set to true,
+          |UseSTD3ASCIIRules| set to true, and |Transitional_Processing| set to false.
+        </li>
+        <aside class="note">
+          The <a>TYPE field</a> contains a domain part stored as ASCII, so for instance "xn--hndvrker-9zan.dk:abc" would
+          be presented as "håndværker.dk:abc" after the above conversion.
+        </aside>
+        <li>
+          Set |record|'s <a>recordType</a> to |domain|, ":" and |type| concatenated.
```
beaufortfrancois: Nit: Use backticks for colon.

index.html
```diff
@@ -4444,8 +4453,19 @@ <h3>Parsing content</h3>
       a |record:NDEFRecord|, run these steps:
       <ol class=algorithm data-link-for="NDEFRecord">
         <li>
-          Set |record|'s <a>recordType</a> to the
-          value of |ndefRecord|'s <a>TYPE field</a>.
+          Let |domain| and |type| be the result of running [=split external type=] with the value of |ndefRecord|'s <a>TYPE field</a>.
```
beaufortfrancois, line 4469: We should return null in cases of failure so that the record is not added to message's records.

beaufortfrancois (changes requested):
Thanks for the PR @kenchris!

index.html
kenchris: Yeah makes sense

index.html
kenchris: My mistake :-) I guess I should have paid more attention

index.html
kenchris: Yes, I really don't see what smart-poster has to do with this, I guess it was copied earlier to use the same language

index.html
kenchris: OK, you don't need to say that when throwing though, but I guess we could fix separately... same as `false` should just be false etc

index.html
kenchris: In this case, they have been validated earlier so they will always be valid

index.html
zolkis: I don't know. To me it sounds more intuitive to have step 2 first since that is the real operation, and then step 1 as an error step of the former, otherwise someone might think they have to scan the string separately to check for `:`? 
No strong opinion here.

index.html
```diff
@@ -3243,15 +3241,31 @@ <h3><dfn>Writing content</dfn></h3>
...
-            [= exception/throw =] a {{TypeError}} and abort these steps.
+            Let |domain| and |type| be the result of running [=split external type=] with
+            |record|'s <a>recordType</a>.
+          </li>
+          <li>
+            Let |domain| be the result of running <a href="https://url.spec.whatwg.org/#concept-domain-to-ascii">
+            domain to ASCII</a> with |domain| and true.
+          </li>
+          <aside class="note">
+            The record's <a>recordType</a> is is user set and in UTF-8 encoding, allowing for international domain names as part of the
```
zolkis: typo: s/is is/is

index.html
```diff
@@ -4444,8 +4454,22 @@ <h3>Parsing content</h3>
       a |record:NDEFRecord|, run these steps:
       <ol class=algorithm data-link-for="NDEFRecord">
         <li>
-          Set |record|'s <a>recordType</a> to the
-          value of |ndefRecord|'s <a>TYPE field</a>.
+          Let |domain| and |type| be the result of running [=split external type=] with the value of |ndefRecord|'s <a>TYPE field</a>.
+        </li>
+        <aside class="note">
+          The above step is ensures to work due to earlier validation of the external type.
```
zolkis: typo: "is ensures"?

index.html
kenchris:
```suggestion
            The record's <a>recordType</a> is user set and in UTF-8 encoding, allowing for international domain names as part of the
```

zolkis (approved):
Nice, thanks.

index.html
kenchris: People can implement it that way as the result is the same, this is about clear spec text :-)

index.html
beaufortfrancois: @zolkis added that comment as part of https://github.com/w3c/web-nfc/commit/24f5f88a074ecde6c6eccada12dff434de031ce8

index.html
beaufortfrancois: Where have they been validated earlier?

beaufortfrancois (changes requested):

index.html
zolkis: IIRC the algorithm was reused in multiple contexts. When the context was "smart-poster", the algorithm was in the course of building a smart poster record. So it's the mode of the algorithm in the context it was called.

If that is confusing, we might want to revisit that and normalize by removing context and repeating the algorithms in the spec. I think I attempted that and didn't like the result, but we could give it a go.

kenchris:
> Where have they been validated earlier?

Actually it hasn't that was the other way around :/ So le me move the comment there and fix

index.html
beaufortfrancois: Let's not remove this as part of this PR and file an issue to clean this if necessary.

index.html
```diff
@@ -3243,15 +3241,34 @@ <h3><dfn>Writing content</dfn></h3>
...
           </li>
-          -->
           <li>
-            If |record|'s <a>mediaType</a> is not `undefined`,
-            [= exception/throw =] a {{TypeError}} and abort these steps.
+            Let |domain| and |type| be the result of running [=split external type=] with
+            |record|'s <a>recordType</a>.
+          </li>
+          <aside class="note">
+            The above step is ensures to work due to earlier validation of the external type.
```
beaufortfrancois: Nit: "The above step is ensured to..." or "The above step always succeeds due to..."

beaufortfrancois (changes requested):

index.html
```diff
@@ -3243,15 +3241,34 @@ <h3><dfn>Writing content</dfn></h3>
         To <dfn>map external data to NDEF</dfn> given |record:NDEFRecordInit|,
         |ndef|, and |recordsDepth:unsigned short|, run these steps:
         <ol class=algorithm data-link-for="NDEFRecordInit">
```
kenchris:
```suggestion
        <ol class=algorithm data-link-for="NDEFRecordInit">
          <!--
          <li>
            If |context| is `"smart-poster"`,
            If |record|'s <a>mediaType</a> is not `undefined`,
            [= exception/throw =] a {{TypeError}} and abort these steps.
          </li>
          -->
```

index.html
```diff
@@ -3246,13 +3244,39 @@ <h3><dfn>Writing content</dfn></h3>
           <!--
           <li>
             If |context| is `"smart-poster"`,
+            If |record|'s <a>mediaType</a> is not `undefined`,
```
kenchris:
```suggestion
```

index.html
```diff
@@ -3253,6 +3251,31 @@ <h3><dfn>Writing content</dfn></h3>
             If |record|'s <a>mediaType</a> is not `undefined`,
             [= exception/throw =] a {{TypeError}} and abort these steps.
           </li>
+          <li>
+            Let |domain| and |type| be the result of running [=split external type=] with
+            |record|'s <a>recordType</a>.
+          </li>
+          <aside class="note">
+            The above step is ensures to work due to earlier validation of the external type.
```
kenchris:
```suggestion
            The above step always succeeds due to earlier validation of the external type.
```

beaufortfrancois (approved):
Thank you @kenchris!

PR #679, opened 2026-04-16
Fix latest published version header

anssiko:
ReSpec docs:
https://respec.org/docs/#notes-47


anssiko:
Deploying as a hotfix.

